PR #5, opened 2020-05-18
customizations-base: Allow local extensions

jan-kiszka:
Taken from jailhouse-images: Allow to enhance the customizations-base
package via a local.inc that is outside of versioning control, e.g. to
inject wifi credentials or authorized ssh keys.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 build/
 isar/
+recipes-core/customizations-base/local.inc
--- a/recipes-core/customizations-base/customizations-base_0.1-iot2050-debian.bb
+++ b/recipes-core/customizations-base/customizations-base_0.1-iot2050-debian.bb
@@ -10,6 +10,9 @@
 
 inherit dpkg-raw
 
+# optional local customizations, not part of the repository
+include local.inc
+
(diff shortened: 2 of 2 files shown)


BaochengSu:
Any example of how to use this .inc file to do injection? Do we need document these examples just in case users need it?

jan-kiszka:
It's more a developer tool, poor-man's customization without versioning, very close to the normally produced image. I've something like this in `recipes-core/customizations-base/local.inc`:

```
WLAN_FILE = "${D}/etc/NetworkManager/system-connections/Private-Network.nmconnection"
PRESERVE_PERMS = " \
    /etc/NetworkManager/system-connections/Private-Network.nmconnection \
    /root"

do_install_append() {
    mkdir -p ${D}/etc/NetworkManager/system-connections
    touch ${WLAN_FILE}
    chmod 0600 ${WLAN_FILE}
    cat <<EOF >> ${WLAN_FILE}
...
EOF

    mkdir -p -m 0700 ${D}/root/.ssh
    echo "ssh-rsa ..." > ${D}/root/.ssh/authorized_keys
}
```

If you build images for production, an example for a custom layer on top would likely be better.

PR #6, opened 2020-05-18
Clean up boot image

jan-kiszka:
Baseline, support for building upstream U-Boot will follow on top.


jan-kiszka:
Will send a v2, there are some more details to be cleaned up (obsolete armhf tools).

PR #7, opened 2020-05-22
linux-iot2050: Disable 64K pages

jan-kiszka:
Breaks at least chromium in various ways, partly because of missing
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=b212921b13bda088a004328457c5c21458262fe2
in 4.19, but it also crashes later when applying that. There are also
other problems to expect for this rather exotic feature, so better drop
it.

@bergmanu


jan-kiszka:
That said, I could imagine TI having enabled it in their BSP to improve slightly the performance (it reduces the TLB pressure and the number of page table levels). We can check back with them how much impact to expect, but I don't think it's a lot.

PR #10, opened 2020-05-26
npm: fix fetch package.json bundled dependencies

davimaf:
fixed npm.bblclass file - when installing a node package that has no bundled dependencies except itself there is a missing key error in function fetch_npm( ).


jan-kiszka:
I would need only the fix up patch for master at this stage, the rest is not yet fully ready. The patch needs to carry a signed-off-by (see other commits) to confirm https://developercertificate.org.

Sorry, some formal effort, but I'd like you to be credited properly for the fix.

classes/npm.bbclass
```diff
@@ -159,8 +159,10 @@ python fetch_npm() {
     with open("package.json") as infile:
         json_objs = json.load(infile)
 
-    deps = [d for d in json_objs['dependencies']]
-    json_objs.update({'bundledDependencies': deps})
+    deps = {}
```
jan-kiszka, line 162: Not needed.

classes/npm.bbclass
```diff
@@ -159,8 +159,10 @@ python fetch_npm() {
...
+    if 'dependencies' in json_objs.keys():
+        deps = [d for d in json_objs['dependencies']]
+        json_objs.update({'bundledDependencies': deps})
```
jan-kiszka, line 165: How about `dependencies = json_objs.get('dependencies')`? Then you could check that for None and also reuse it later.

classes/npm.bbclass
davimaf: right. delete deps init

classes/npm.bbclass
davimaf: right. modified with suggested method 
```
deps = json_objs.get('dependencies')
if deps:
  json_objs.update({'bundledDependencies': deps})
```

PR #15, opened 2020-06-09
SDK improvements

jan-kiszka:


conf/distro/isar.conf
```diff
@@ -37,3 +37,5 @@ HOST_DISTRO_APT_SOURCES_remove_packages-snapshot = "${DISTRO_APT_SOURCES_MAINLIN
 
 # Required as long as openssl is not rebuilt for the host arch
 HOST_DISTRO_APT_PREFERENCES = "conf/distro/preferences.upstream-libssl"
+
+SDK_PREINSTALL += "zlib1g-dev"
```
johnson325, line 42: This also implies that any customized package should be here right?

jan-kiszka: That list is suspiciously short, indeed. We may need to add further -dev packages here that the user may need in order to build against pre-installed lib packages etc.

PR #25, opened 2020-06-17
ethernet: Swap eth0 and eth1 port

AsuraZeng:
as the ethernet labels on the box do not correspond to the
    ethernet port on the board.so needed swap the ethernet port.

    1. add the alias support for the ethernet driver

    2. modify the alias configuration to swap the ethernet port

### DIFF
--- a/recipes-kernel/linux/files/0023-ethernet-add-the-alias-support-for-the-ethernet-driv.patch
+++ b/recipes-kernel/linux/files/0023-ethernet-add-the-alias-support-for-the-ethernet-driv.patch
@@ -1,0 +1,52 @@
+From f75956d44496092c325955dc93f4563656f58dbd Mon Sep 17 00:00:00 2001
+From: Chao Zeng <chao.zeng@siemens.com>
+Date: Thu, 18 Jun 2020 09:13:39 +0800
+Subject: [PATCH 23/23] ethernet:add the  alias support for the ethernet driver
+
+Signed-off-by: Chao Zeng <chao.zeng@siemens.com>
+---
+ drivers/net/ethernet/ti/icssg_prueth.c | 14 +++++++++++++-
+ 1 file changed, 13 insertions(+), 1 deletion(-)
+
+diff --git a/drivers/net/ethernet/ti/icssg_prueth.c b/drivers/net/ethernet/ti/icssg_prueth.c
+index 0984daafa585..3517be959f4d 100644
+--- a/drivers/net/ethernet/ti/icssg_prueth.c
++++ b/drivers/net/ethernet/ti/icssg_prueth.c
+@@ -1808,7 +1808,7 @@ static int prueth_probe(struct platform_device *pdev)
+ 	struct device_node *eth0_node, *eth1_node;
+ 	const struct of_device_id *match;
+ 	struct pruss *pruss[NUM_ICSSG];
+-	int i, ret, msmc_ram_size;
++	int i, ret, msmc_ram_size, id;
+ 
+ 	if (!np)
+ 		return -ENODEV;	/* we don't support non DT */
+@@ -2017,6 +2017,12 @@ static int prueth_probe(struct platform_device *pdev)
+ 
+ 	/* register the network devices */
+ 	if (eth0_node) {
++		id = of_alias_get_id(eth0_node,"ethernet");
++		if (id < 0){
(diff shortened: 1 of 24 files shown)


recipes-kernel/linux/files/0023-ethernet-Swap-eth0-and-eth1-port.patch
```diff
@@ -0,0 +1,69 @@
+From 8dc174b64873a9be4a30862c1c952f460f2ef38d Mon Sep 17 00:00:00 2001
+From: Chao Zeng <chao.zeng@siemens.com>
+Date: Wed, 17 Jun 2020 13:20:16 +0800
+Subject: [PATCH 23/23] ethernet: Swap eth0 and eth1 port 	1. add the
+ alias support for the ethernet driver         2. modify the alias
+ configuration to swap the ethernet port
```
jan-kiszka, line 6: Should be separated into two patches. The alias support needs to go to TI so that they integrate it into the driver (until that is upstream).

recipes-kernel/linux/files/0023-ethernet-add-the-alias-support-for-the-ethernet-driv.patch
```diff
@@ -0,0 +1,52 @@
...
+ 	if (!np)
+ 		return -ENODEV;	/* we don't support non DT */
+@@ -2017,6 +2017,12 @@ static int prueth_probe(struct platform_device *pdev)
+ 
+ 	/* register the network devices */
+ 	if (eth0_node) {
++		id = of_alias_get_id(eth0_node,"ethernet");
++		if (id < 0){
++			dev_err(dev, "device does not have an alias id\n");
++			return id;
```
BaochengSu, line 31: So the aliase is a must for prueth node? is it too stiff?

BTW, please use `goto netdev_exit;` instead of `return id;`

recipes-kernel/linux/files/0023-ethernet-add-the-alias-support-for-the-ethernet-driv.patch
```diff
@@ -0,0 +1,52 @@
...
+@@ -2027,6 +2033,12 @@ static int prueth_probe(struct platform_device *pdev)
+ 	}
+ 
+ 	if (eth1_node) {
++		id = of_alias_get_id(eth1_node,"ethernet");
++		if (id < 0){
++			dev_err(dev, "device does not have an alias id\n");
++			return id;
++		}
++		sprintf(prueth->emac[PRUETH_MAC1]->ndev->name,"eth%d",id);
```
BaochengSu, line 46: What if the ethX has already be occupied by other devices? do we need some fallback mechanism?

recipes-kernel/linux/files/0023-ethernet-Swap-eth0-and-eth1-port.patch
BaochengSu: @jan-kiszka  do you think if it's better to change the network interface name from user space, such as udev?

jan-kiszka: It would be an alternative, indeed.

Looking at the kernel, there are a few drivers doing aliases:
```
drivers/net/ethernet/broadcom/genet/bcmmii.c:           id = of_alias_get_id(dn, "eth");
drivers/net/ethernet/samsung/sxgbe/sxgbe_platform.c:    plat->bus_id = of_alias_get_id(np, "ethernet");
drivers/net/ethernet/stmicro/stmmac/stmmac_platform.c:  plat->bus_id = of_alias_get_id(np, "ethernet");
```

jan-kiszka: Thinking about this further: an udev rule would allow to call the interfaces exactly like the ports (x1p1, x1p2).

AsuraZeng: The concern is that if we add the alias support for the driver. In the future, we submit the patch to the upstream.  Anyone else uses this driver would cause some error. 
Maybe Ethernet  port conflict  if we assign the port with duplicate port number.  

Maybe use the Udev is a good alternative solution for this feature.

PR #26, opened 2020-06-19
ethernet: swap the ethernet port name

AsuraZeng:
add the udev rules to swap the ethernet port name

### DIFF
--- a/recipes-core/customizations-example/files/20-swap-ethernet-port.rules
+++ b/recipes-core/customizations-example/files/20-swap-ethernet-port.rules
@@ -1,0 +1,2 @@
+KERNEL=="eth0",SUBSYSTEM=="net",KERNELS=="pruss0_eth",SUBSYSTEMS=="platform",NAME="eth1"
+KERNEL=="eth1",SUBSYSTEM=="net",KERNELS=="pruss0_eth",SUBSYSTEMS=="platform",NAME="eth0"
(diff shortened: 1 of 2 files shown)


recipes-core/customizations-example/files/20-swap-ethernet-port.rules
```diff
@@ -0,0 +1,2 @@
+KERNEL=="eth0",SUBSYSTEM=="net",KERNELS=="pruss0_eth",SUBSYSTEMS=="platform",NAME="eth1"
+KERNEL=="eth1",SUBSYSTEM=="net",KERNELS=="pruss0_eth",SUBSYSTEMS=="platform",NAME="eth0"
```
jan-kiszka, line 2: eth<n> or x1p<n>? Did you discuss this? @BaochengSu, @johnson325?

Looks good to me otherwise.

AsuraZeng: I  have discussed with @BaochengSu , We think eth  is compatible with IOT2000.
And the x1p is a good sign to identify Ethernet port. Both of them are ok. 
We would discuss with  @johnson325  at the next Monday to make a decision

johnson325: I'd like to have the same name as v1.   
No problem from my side.

PR #31, opened 2020-07-09
Provide uboot upgrade script

AsuraZeng:
as needed update uboot from user space,provide update script

### DIFF
--- a/recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.py
+++ b/recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.py
@@ -1,0 +1,200 @@
+#!/usr/bin/env python3
+import sys
+import os
+import fcntl
+import struct
+import mmap
+import hashlib
+from optparse import OptionParser
+
+force_update = False
+
+
+def FirmwareUpdateException(str):
+    """Report the error info and exit"""
+    if not force_update:
+        print(str)
+        sys.exit(1)
+
+
+class FirmwareUpdate(object):
+    def __init__(self, force):
+        self.force_update = force
+
+    def sha256_check(self, file):
+        """Firmware Sha256 Check"""
+        check_sha256_file = file+'.sha256'
+        file_sha256 = None
+        try:
+            with open(check_sha256_file, 'r') as f:
+                file_sha256 = str(f.read()).split()[0]
+        except FileNotFoundError:
+            FirmwareUpdateException("sha256 file not exists")
+
+        try:
+            with open(file, 'rb+') as f:
+                method = hashlib.sha256()
+                method.update(f.read())
+                hash_code = method.hexdigest()
+        except FileNotFoundError:
+            print("%s not Exists" % file)
+            sys.exit(1)
+
+        if file_sha256 != hash_code:
+            if(self.force_update is not True):
+                FirmwareUpdateException("Image Checksum is not Correct")
+
+    def get_path_type_value(self, path):
+        """get the path value"""
+        try:
+            with open(path, "r") as f:
+                return f.read()
+        except FileNotFoundError:
+            print("%s is not exists" % path)
+            sys.exit(1)
+
+    @staticmethod
+    def flash_erase(dev, start, nbytes):
+        """This function erase the flash
+
+        @dev:the flash device
+
+        @start:start address
+
+        @nbytes:the number of erase bytes
+        """
+        MEMERASE = 0x40084d02
+
+        try:
+            fd = os.open(dev, os.O_SYNC | os.O_RDWR)
+        except FileNotFoundError:
+            print("failed to open %s" % dev)
+            sys.exit(1)
+
+        ioctl_data = struct.pack('II', start, nbytes)
+
+        try:
+            fcntl.ioctl(fd, MEMERASE, ioctl_data)
+        except IOError:
+            print("ioctl failed")
+            sys.exit(1)
+
+        os.close(fd)
+
+    def cpu_id_check(self):
+        """cpu id check"""
+        cpu_device_id = [0x142ba, 0x142fa, 0x140ff]
+        cpuid_register_addr = 0x43000018
+        base_addr = cpuid_register_addr & ~(mmap.PAGESIZE - 1)
+        base_addr_offset = cpuid_register_addr - base_addr
+        try:
+            f = os.open('/dev/mem', os.O_RDWR | os.O_SYNC)
+        except FileNotFoundError:
+            print("Open /dev/mem Failed")
+            sys.exit(1)
+
+        mem = mmap.mmap(f, mmap.PAGESIZE, mmap.MAP_SHARED, mmap.PROT_READ, offset=base_addr)
+        mem.seek(base_addr_offset)
+        data = []
+        data.append(struct.unpack('I', mem.read(4))[0])
+        device_id = hex(data[0])
+        current_id = int(device_id, 16) >> 11
+        if current_id not in cpu_device_id:
+            FirmwareUpdateException("Upgrade is not supported for the board")
+
+    def update_firmware(self, filename):
+        """Update Firmware"""
+        mtd_num = 0
+        img_pos = 0
+
+        print("===================================================")
+        print("IOT2050 firmware update started - DO NOT INTERRUPT!")
+        print("===================================================")
+
+        while True:
+            mtd_sys_path = "/sys/bus/platform/devices/47040000.spi/mtd/mtd{}".format(mtd_num)
+            mtd_name_path = "{}/name".format(mtd_sys_path)
+            mtd_size_path = "{}/size".format(mtd_sys_path)
+            mtd_erasesize_path = "{}/erasesize".format(mtd_sys_path)
+            mtd_dev_path = "/dev/mtd{}".format(mtd_num)
+
+            mtd_size = self.get_path_type_value(mtd_size_path)
+            mtd_size = int(mtd_size)
+
+            mtd_erasesize = self.get_path_type_value(mtd_erasesize_path)
+            mtd_erasesize = int(mtd_erasesize)
+
+            mtd_name = self.get_path_type_value(mtd_name_path)
+
+            print("Updating %-20s" % mtd_name.strip(),end="")
+            if mtd_name.strip() == "ospi.rootfs":
+                print("\n\nCompleted. Please reboot the device\n")
+                sys.exit(0)
+
+            mtd_pos = 0
+            while mtd_pos < mtd_size:
+                try:
+                    with open(mtd_dev_path, 'rb') as mtd:
+                        mtd.seek(mtd_pos)
+                        mtd_contents = mtd.read(mtd_erasesize)
+                except FileNotFoundError:
+                    print("Open %s failed" % mtd_dev_path)
+                    sys.exit(1)
+
+                with open(filename, 'rb') as firmware:
+                    firmware.seek(img_pos)
+                    firmware_contents = firmware.read(mtd_erasesize)
+
+                if not mtd_contents == firmware_contents:
+                    print("U", end="")
+                    sys.stdout.flush()
+                    self.flash_erase(mtd_dev_path, mtd_pos, mtd_erasesize)
+                    with open(mtd_dev_path, 'wb') as mtd_unity:
+                        mtd_unity.seek(mtd_pos)
+                        mtd_unity.write(firmware_contents)
+                else:
+                    print(".", end="")
+                    sys.stdout.flush()
+                mtd_pos = mtd_pos+mtd_erasesize
+                img_pos = img_pos+mtd_erasesize
+            print()
+            mtd_num = (mtd_num+1)
+
+
+def main(argv):
+    usage = "%prog [Options] Firmware"
+    parser = OptionParser(usage)
+    parser.add_option('-f', '--force', action='store_true', dest='force', default=False, help="Force update, ignoring image checksum or device ID mismatches")
+
+    (options, args) = parser.parse_args()
+
+    if not args:
+        print(parser.parse_args(['-h']))
+        sys.exit(1)
+
+    erase_env_input = input("\nWarning: All U-Boot environment variables will be reset to factory settings. Continue (y/N)? ")
+
+    if not erase_env_input == "y":
+        sys.exit(1)
+
+    if options.force:
+        force_update_input = input("\nWarning: Enforced update may render device unbootable. Continue (y/N)? ")
+        if not force_update_input == "y":
+            sys.exit(1)
+        else:
+            global force_update
+            force_update = options.force
+
+    filename = argv[-1]
+
+    firmupdate = FirmwareUpdate(options.force)
+
+    firmupdate.sha256_check(filename)
+
+    firmupdate.cpu_id_check()
+
+    firmupdate.update_firmware(filename)
+
+
+if __name__ == '__main__':
+    main(sys.argv)
--- a/recipes-app/iot2050-firmware-update/iot2050-firmware-update_0.1.bb
+++ b/recipes-app/iot2050-firmware-update/iot2050-firmware-update_0.1.bb
@@ -1,0 +1,11 @@
+DESCRIPTION = "Uboot Firmware Update Scripts"
+MAINTAINER = "chao.zeng@siemens.com"
+
+SRC_URI = "file://iot2050-firmware-update.py"
+
+inherit dpkg-raw
+
+do_install() {
+    install -v -d ${D}/usr/sbin/
+    install -v -m 755 ${WORKDIR}/iot2050-firmware-update.py ${D}/usr/sbin/iot2050update
+}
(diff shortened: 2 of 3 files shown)


recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.py
```diff
@@ -0,0 +1,183 @@
...
+        excute_result=subprocess.run(cmd,stdout=subprocess.PIPE)
+        return excute_result
+
+    def dd_execute(self,input,skip,count,output,seek,bs=512):
+        """
+        wrap the shell dd cmd
+        """
+        DD_DEFINE_BASE=["dd","iflag=skip_bytes,count_bytes","oflag=seek_bytes","conv=notrunc","status=none"]
+        DD_EXECUTE_CMD=["if={}".format(input),"skip={}".format(skip),"count={}".format(count),"of={}".format(output),"seek={}".format(seek),"bs={}".format(bs)]
+        self.shell_cmd_execute(DD_DEFINE_BASE+DD_EXECUTE_CMD)
```
jan-kiszka, line 64: Why not writing out to /dev/mtdX with Python means?

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.py
```diff
@@ -0,0 +1,183 @@
...
+    def cpu_id_check(self):
+        """
+        CPU ID CHECK
+        """
+        cpu_device_id=[0x142fa,]
+        READ_DEVICE_ID_CMD=["busybox","devmem","0x43000018"]
+        DEVICE_ID_RE=self.shell_cmd_execute(READ_DEVICE_ID_CMD).stdout.decode()
```
jan-kiszka, line 71: Why not using the pattern I provided? You can use native Python for this task.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.py
```diff
@@ -0,0 +1,183 @@
...
+                    print("U",end="")
+                    ERASE_CMD=["flash_erase","-q","/dev/mtd{}".format(MTDNUM),"{}".format(MTDPOS),"1"]
+                    self.shell_cmd_execute(ERASE_CMD)
+                    self.dd_execute(TMP_IMG,0,MTD_ERASESIZE,"/dev/mtd{}".format(MTDNUM),MTDPOS,MTD_ERASESIZE)
+                else:
+                    print(".",end="")
+                IMGPOS=IMGPOS+MTD_ERASESIZE
+                MTDPOS=MTDPOS+MTD_ERASESIZE
+            MTDNUM=(MTDNUM+1)
+            print("\n")
```
jan-kiszka, line 146: This should also be rewritten to use Python were possible. The current code is basically wrapping the a shell script into Python, rather than expressing the task in Python directly.

There is even a way to request the flash erasing natively (MEMERASE ioctl for mtd), which would have the benefit of overcoming the tools dependency.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.py
```diff
@@ -0,0 +1,183 @@
...
+class FirmwareUpdateException(Exception):
+    def __init__(self,ErrorInfo):
+        self.errorinfo=ErrorInfo
+    def __str__(self):
+        if not force_update:
+            return self.errorinfo
+
+def main():
+    parser = OptionParser()
+    parser.add_option('-f','--force',action='store_true',dest='force',default=False,help="Force update the firmware,ignore the error")
```
jan-kiszka, line 157: Which errors does this mode ignore? What is the use case behind it?

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.py
```diff
@@ -0,0 +1,183 @@
...
+    parser.add_option('-e','--erase',action='store_true',dest='erase',default=False,help="Erase the Environment")
+    parser.add_option('-F','--file',type='string',dest='filename',help="firmware file name")
```
jan-kiszka, line 159: I would model the file name as positional parameter.

recipes-app/iot2050-firmware-update/iot2050-firmware-update_0.1.bb
```diff
@@ -0,0 +1,11 @@
...
+MAINTAINER = "chao.zeng@siemens.com"
+
+SRC_URI = "file://iot2050-firmware-update.py"
+
+inherit dpkg-raw
+
+do_install() {
+    install -v -d ${D}/usr/bin/
+    install -v -m 755 ${WORKDIR}/iot2050-firmware-update.py ${D}/usr/bin/iot2050-firmware-update
+}
```
jan-kiszka, line 11: As long as you need mtd-utils, this package should runtime depend on it.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.py
```diff
@@ -0,0 +1,183 @@
...
+        wrap the shell dd cmd
+        """
+        DD_DEFINE_BASE=["dd","iflag=skip_bytes,count_bytes","oflag=seek_bytes","conv=notrunc","status=none"]
+        DD_EXECUTE_CMD=["if={}".format(input),"skip={}".format(skip),"count={}".format(count),"of={}".format(output),"seek={}".format(seek),"bs={}".format(bs)]
+        self.shell_cmd_execute(DD_DEFINE_BASE+DD_EXECUTE_CMD)
+    def cpu_id_check(self):
+        """
+        CPU ID CHECK
+        """
+        cpu_device_id=[0x142fa,]
```
jan-kiszka, line 69: We likely want to list all supported types here, ie. basic and advanced, possibly also variants like mine (0x140ff).

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.py
```diff
@@ -0,0 +1,183 @@
...
+
+        try:
+            with open(file, 'rb+') as f:
+                hash_code=sha256_file(f,0)
+        except:
+            raise FirmwareUpdateException("Firmware not Exists")
+        
+        if check_sha256 != hash_code:
+            if( self.force_update != True):
+                raise FirmwareUpdateException("File Checksum is not Correct")
```
jan-kiszka, line 47: I suspect those exceptions will give rather verbose error reportings when they occur. I would recommend using plain error messages and sys.exit(1) instead.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.py
AsuraZeng: add the "-f" ,it will ignore  the "checksum"   "cpuid" not match error and so on.
or it would show the error.
and add the prompt to reminder user to do the operation

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.py
AsuraZeng: this has  marked TBD,it would be filled when I get the correct CPUID of the basic and advanced board

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.py
```diff
@@ -0,0 +1,242 @@
...
+from optparse import OptionParser
+from tempfile import NamedTemporaryFile
+
+force_update=False
+
+def FirmwareUpdateException(str):
+    """Report the error info and exit"""
+    if not force_update:
+        print(str)
+        sys.exit(1)
```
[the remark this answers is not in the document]
AsuraZeng: there are server possible use cases: 
1. no .sha file,but just want to update the firmware
2. the cpuid not list in the code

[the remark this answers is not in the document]
jan-kiszka: OK, but then only address those two, not all the other errors.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.py
```diff
@@ -0,0 +1,242 @@
...
+
+def sha256_file(file,length):
+    """Return sha256 of the file"""
+    method=hashlib.sha256()
+    try:
+        with mmap.mmap(file.fileno(),length,access=mmap.ACCESS_READ) as mm:
+            chunk = mm.read(length)
```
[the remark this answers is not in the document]
AsuraZeng: using mmap and read both are ok for this case 
I dont think there is any disadvantage using mmap?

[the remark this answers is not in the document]
jan-kiszka: Both are fine, mmap just looked to me like over-optimization (at the price of an extra line). We are not handling gigabytes of data.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.py
```diff
@@ -0,0 +1,193 @@
...
+                print(".")
+            img_pos = img_pos+mtd_size
+            mtd_num = (mtd_num+1)
+            print("\n")
+
+
+def main(argv):
+    parser = OptionParser()
+    parser.add_option('-f', '--force', action='store_true', dest='force', default=False, help="Force update the firmware,ignore the error")
+    parser.add_option('-e', '--erase', action='store_true', dest='erase', default=False, help="Erase the Environment")
```
jan-kiszka, line 165: Just realized: --erase must be default on unless we add a method to ensure that the vars u-boot normally writes are updated.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.py
```diff
@@ -0,0 +1,193 @@
...
+        result = input("\nDo force update.Maybe Destory your System:Y/n ")
+        if result == "n":
+            options.force = False
+        else:
+            global force_update
+            force_update = options.force
+
+    if not args:
+        print("Require Firmware")
+        sys.exit(1)
```
jan-kiszka, line 179: Flash file must be modeled as positional parameter so that --help also shows it. I'm using ArgumentParser elsewhere, but I suspect the OptionParser can do that as well.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.py
```diff
@@ -0,0 +1,193 @@
...
+            else:
+                print(".")
+            img_pos = img_pos+mtd_size
+            mtd_num = (mtd_num+1)
+            print("\n")
+
+
+def main(argv):
+    parser = OptionParser()
+    parser.add_option('-f', '--force', action='store_true', dest='force', default=False, help="Force update the firmware,ignore the error")
```
jan-kiszka, line 164: I would rephrase that to "Force update, ignoring image checksum or device ID mismatches".

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.py
```diff
@@ -0,0 +1,193 @@
...
+    parser.add_option('-e', '--erase', action='store_true', dest='erase', default=False, help="Erase the Environment")
+
+    (options, args) = parser.parse_args()
+
+    if options.force:
+        result = input("\nDo force update.Maybe Destory your System:Y/n ")
```
jan-kiszka, line 170: "Warning: Enforced update may render device unbootable. Continue (y/N)?"

Means, also default to "no".

jan-kiszka:
Just used the script. It worked fine, but I think we can improve progress reporting a bit. There are unneeded line-breaks, and I'm missing "." to show that sectors are processed, like there was in my shell-based script.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.py
AsuraZeng: If do that, we can not retain the environment that user sets and changes.

I found it can not parse the VARS "default_environment" from .bin file.
It could combine the ELF file(uboot) to get some symbol info and get the VARS from the decompile .bin to do the upgrade.  Or just provide the symbol info of the "default_environment" along with upgrade bin file.

Do you some suggestions how to retain the  environment  for both of the user and the upgrade Firmware Simultaneously

jan-kiszka: As with many low-level updates, keeping user settings is an add-on, not a must-have. The higher value is maintaining consistency, and that implies in its basic form overwriting the env.

The problem is that even the default env, eventually available via a new package (feature under development elsewhere), is not enough because we transfer dynamic information into the env, see https://github.com/siemens/u-boot/blob/dfdc967909de39a8abfaf02551f2a3fd8d24ac8a/board/siemens/iot2050/board.c#L56. By default, this must be in sync after an update, thus requires to reset the env. Well, at least to clear `board_uuid` in order to trigger rewriting of the other vars on first boot. But clearing `board_uuid` would also overwrite potential customizations of MAC addresses or fdtfile or who knows.

And, finally, the risk that keeping the old env will break booting with the new firmware version is simply too high, e.g. when switching from the current downstream U-Boot to upstream. Or when switching between current open boot and locked-down secure boot. So we shouldn't try this by default, maybe not even at all. Experts editing the env can also easily reapply changes to the new version.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.py
```diff
@@ -0,0 +1,200 @@
...
+    def __init__(self, force):
+        self.force_update = force
+
+    def sha256_check(self, file):
+        """Firmware Sha256 Check"""
+        check_sha256_file = file+'.sha256'
+        file_sha256 = None
+        try:
+            with open(check_sha256_file, 'r') as f:
+                file_sha256 = str(f.read()).strip()
```
jan-kiszka, line 30: Just tested this: Needs to be `str(f.read()).split()[0]` so that it works with regular .sha256 files. Those have the format that sha256sum generates (`<sum> <filename>`).

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.py
```diff
@@ -0,0 +1,200 @@
...
+                    self.flash_erase(mtd_dev_path, mtd_pos, mtd_erasesize)
+                    with open(mtd_dev_path, 'wb') as mtd_unity:
+                        mtd_unity.seek(mtd_pos)
+                        mtd_unity.write(firmware_contents)
+                else:
+                    print(".", end="")
+                    sys.stdout.flush()
+                mtd_pos = mtd_pos+mtd_erasesize
+                img_pos = img_pos+mtd_erasesize
+            print("\n")
```
jan-kiszka, line 160: `print("\n")` should just be `print()`, otherwise you have two linebreaks.

recipes-app/iot2050-firmware-update/iot2050-firmware-update_0.1.bb
```diff
@@ -0,0 +1,11 @@
+DESCRIPTION = "Uboot Firmware Update Scripts"
+MAINTAINER = "chao.zeng@siemens.com"
+
+SRC_URI = "file://iot2050-firmware-update.py"
+
+inherit dpkg-raw
+
+do_install() {
+    install -v -d ${D}/usr/bin/
```
jan-kiszka, line 9: Better would be `/usr/sbin` - it's an privileged tool.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.py
```diff
@@ -0,0 +1,200 @@
...
+    parser = OptionParser(usage)
+    parser.add_option('-f', '--force', action='store_true', dest='force', default=False, help="Force update, ignoring image checksum or device ID mismatches")
+
+    (options, args) = parser.parse_args()
+
+    if not args:
+        print(parser.parse_args(['-h']))
+        sys.exit(1)
+
+    erase_env_input = input("\nWarning: Environment variables set by the user may be destroyed. Continue (y/N)?")
```
jan-kiszka, line 175: I would propose a little bit tuning of the working:
`"\nWarning: All U-Boot environment variables will be reset to factory settings. Continue (y/N)? "`

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.py
```diff
@@ -0,0 +1,200 @@
...
+
+    if not erase_env_input == "y":
+        sys.exit(1)
+
+    if options.force:
+        force_update_input = input("\nWarning: Enforced update may render device unbootable. Continue (y/N)?")
```
jan-kiszka, line 181: Wording it good, just add a space at the end of the string so that the user input will not appear as "glued" to it.

AsuraZeng: @jan-kiszka   Have tested on the basic and advanced board. both are ok.

jan-kiszka:
Looks good to me now!

PR #35, opened 2020-07-22
upstream u boot kernel

jan-kiszka:
This allows to test upstream queues for ATF, OP-TEE, U-Boot and kernel, just by appending :kas/opt/upstream.yml to a build. Not for production yet, but getting much closer.

Note that someone should check of the generated U-Boot and image work fine on a vanilla IOT2050 advanced.


jan-kiszka:
Also note that we still have two U-Boot kas files, also for upstream. But both will generate the some image. To be resolved when we switch completely to upstream.

PR #36, opened 2020-08-03
u-boot-iot2050: Clean Up U-BOOT

AsuraZeng:
1.as the patches are already merge into the upstrem u-boot
branch,delete the ununsed patches
        2.Set the upstrem u-boot as default


jan-kiszka:
The originating branch for U-Boot is not stable yet. We should only switch once that has happened, using a fixed upstream version + ours and TI's pending patches.

AsuraZeng:
I think there should be only one upstream branch, otherwise it may cause confusion for users.
which one should the user use?

and If there is a problem, is it possible to roll back the version to solve this problem

PR #38, opened 2020-08-21
WIP: Add mtdparts definition for upstream 5.9

jan-kiszka:
The device name of the OSPI changed in 5.9, and that requires us to
provide mtdparts both for the old 47040000.spi.0 as well as the new
spi7.0. Move the common part into a variable and convert the wks file
into a template for that purpose.


jan-kiszka:
I'm clarifying if the renaming was actually intended: https://lkml.org/lkml/2020/8/24/18

PR #43, opened 2020-08-31
enable the traffic control function for kernel

AsuraZeng:


applea9:
I suggest to support NETMAP for iptables/netfilter, need to add CONFIG_IP_NF_TARGET_NETMAP=m to kernel config.

jan-kiszka:
Patch looks good to me.

@applea9 NETMAP is a separate topic - PR welcome!

PR #44, opened 2020-08-31
enable iptables NETMAP function for kernel

applea9:


jan-kiszka:
LGTM.

jan-kiszka:
...except: We need a Signed-off-by according to https://developercertificate.org/.

PR #45, opened 2020-09-09
enable iptables NETMAP function for kernel

applea9:
Signed-off-by: applea9 <applea9@outlook.com>


applea9:
add Signed-off-by according to jan's comments.

BaochengSu (approved):

Issue #40, opened 2020-08-24
add libgpiod with tools to example image


gramss:
Since Linux kernel version 4.8 gpios can be controlled via an chardev interface, see here: https://embeddedbits.org/new-linux-kernel-gpio-user-space-interface/

I was able to install libgpiod from apt, but this seems to be build without the tools.

So I was able to install libgpiod from source https://github.com/brgl/libgpiod/tree/v0.1.x with version v0.1.x (as the version installed as binary from apt with standard sources).

Needed to install `autoconf autoconf-archive libtool` but then it worked.

I see a lot of pins, but only some of them are named. Would it be possible to name them in future releases?

Here is the output I get:

```
root@iot2050-debian:~# gpioinfo 
gpiochip6 - 4 lines:
	line   0:      unnamed       unused   input  active-high 
	line   1:      unnamed       unused   input  active-high 
	line   2:      unnamed       unused   input  active-high 
	line   3:      unnamed       unused   input  active-high 
gpiochip5 - 90 lines:
	line   0:      unnamed       unused   input  active-high 
	line   1:      unnamed       unused   input  active-high 
	line   2:      unnamed       unused   input  active-high 
	line   3:      unnamed       unused   input  active-high 
	line   4:      unnamed       unused   input  active-high 
	line   5:      unnamed       unused   input  active-high 
	line   6:      unnamed       unused   input  active-high 
	line   7:      unnamed       unused   input  active-high 
	line   8:      unnamed       unused   input  active-high 
	line   9:      unnamed       unused   input  active-high 
	line  10:      unnamed       unused   input  active-high 
	line  11:      unnamed       unused   input  active-high 
	line  12:      unnamed       unused   input  active-high 
	line  13:      unnamed       unused   input  active-high 
	line  14:      unnamed       unused   input  active-high 
	line  15:      unnamed       unused   input  active-high 
	line  16:      unnamed       unused   input  active-high 
	line  17:      unnamed       unused   input  active-high 
	line  18:      unnamed       unused   input  active-high 
	line  19:      unnamed       unused   input  active-high 
	line  20:      unnamed       unused   input  active-high 
	line  21:      unnamed       unused   input  active-high 
	line  22:      unnamed       unused   input  active-high 
	line  23:      unnamed       unused   input  active-high 
	line  24:      unnamed       unused   input  active-high 
	line  25:      unnamed       unused   input  active-high 
	line  26:      unnamed       unused   input  active-high 
[...]

PR #48, opened 2020-09-29
Give up the use of fixed gpio numbers

AsuraZeng:
insead of using dynamic gpio numbers according to the probe sequence

### DIFF
--- a/recipes-app/switchserialmode/files/src/switchserialmode.c
+++ b/recipes-app/switchserialmode/files/src/switchserialmode.c
@@ -3,6 +3,7 @@
  *
  * Authors:
  *  Gao Nian <nian.gao@siemens.com>
+ *  Chao Zeng <chao.zeng@siemens.com>
  *
  * This file is subject to the terms and conditions of the MIT License.  See
  * COPYING.MIT file in the top-level directory.
@@ -27,6 +28,7 @@
 #include <malloc.h>
 #include <libusb-1.0/libusb.h>
 #include <stdarg.h>
+#include <glob.h>
 
 typedef unsigned char   BYTE;
 typedef unsigned int    UINT;
@@ -47,6 +49,27 @@ typedef enum{
 }E_CASE_SENSITIVE;
 
 #define MUTIL_THREAD   100
+#define SYSFS_CLASS_GPIO "/sys/class/gpio"
+#define MAX_SIZE 64
+#define UART_CTRL_NUM   4
+
+enum uart_gpio_ctrl {
+    UART_MODE0_CTRL,
+    UART_MODE1_CTRL,
+    UART_EN_CTRL,
+    UART_TERMINATE_CTRL,
+};
+
+typedef struct uart_ctrl_info
+{
+    char (*uart_ctrl_path)[MAX_SIZE];
+    char (*uart_ctrl_base)[MAX_SIZE];
+}uart_ctrl_info_t;
+
+char uart_sysfs_base[UART_CTRL_NUM][MAX_SIZE];
+char uart_sysfs_path[UART_CTRL_NUM][MAX_SIZE];
+
+uart_ctrl_info_t ctr_info;
 
 static char *format(const char *fmt, ...)
 {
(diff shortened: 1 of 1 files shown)


AsuraZeng:
these changes are for upstream kernel.
should be used when migrating to the upstream kernel.

jan-kiszka:
> these changes are for upstream kernel.
> should be used when migrating to the upstream kernel.

But they do not break the 4.19-ti kernel, do they?

AsuraZeng:
This means that it cannot be used with the current ti kernel. To be used with the upstream kernel

jan-kiszka:
Because wkup_gpio0_base does not match the DT used there? Maybe you can add the corresponding patch for the TI kernel as well and probe for what is available? Then we can easily allow switching kernels without patching userland.

AsuraZeng:
I think just like MRAA, there should be some modifications for the switchserial on the userland instead of patching for the ti kernel.

jan-kiszka:
I'm not suggesting to patch the kernel. I'm suggesting to account for any potential differences between upstream DT and TI DT in the switch-serial tool to make it work on both.

BaochengSu:
Should we use the "gpiolib" interface introduced since v4.8 kernel?

jan-kiszka:
If it simplifies the code, let's do that. But speed is surely not a driver here.

gramss:
> 
> 
> Should we use the "gpiolib" interface introduced since v4.8 kernel?

As suggested here: #40 , I can only say +1 on that.. 👍

recipes-app/switchserialmode/files/src/switchserialmode.c
```diff
@@ -237,20 +297,56 @@ void gpio_set(const char *gpiodev, const char *offset, const char *value)
...
+    int offset = 4;//4 5 6 7
+    char uart_path[MAX_SIZE] = {0};
+    char uart_base[MAX_SIZE] = {0};
+    int num;
+    int wkup_gpio0_base = get_gpiochip_base("/sys/devices/platform/interconnect@100000/interconnect@100000:interconnect@28380000/interconnect@100000:interconnect@28380000:interconnect@42040000/42110000.wkup_gpio0");
+
+    if(wkup_gpio0_base < 0)
+    {
+        return -1;
+    }
```
jan-kiszka, line 312: Inconsistent coding style.

recipes-app/switchserialmode/files/src/switchserialmode.c
```diff
@@ -237,20 +297,56 @@ void gpio_set(const char *gpiodev, const char *offset, const char *value)
...
+    gpio_set(ctr_info.uart_ctrl_path[UART_EN_CTRL],ctr_info.uart_ctrl_base[UART_EN_CTRL] , "1");            /* uart_en gpio */
+    gpio_set(ctr_info.uart_ctrl_path[UART_MODE0_CTRL], ctr_info.uart_ctrl_base[UART_MODE0_CTRL], va ? "1" : "0"); /* uart_mode0 gpio */
+    gpio_set(ctr_info.uart_ctrl_path[UART_MODE1_CTRL], ctr_info.uart_ctrl_base[UART_MODE1_CTRL], vb ? "1" : "0"); /* uart_mode1 gpio */
 }
 
 static void gpio_set_terminate(int onoff)
 {
-    gpio_set("/sys/class/gpio/gpio193", "193", onoff ? "1" : "0"); /* uart_terminate gpio */
+    gpio_set(ctr_info.uart_ctrl_path[UART_TERMINATE_CTRL], ctr_info.uart_ctrl_base[UART_TERMINATE_CTRL], onoff ? "1" : "0"); /* uart_terminate gpio */
+
```
jan-kiszka, line 339: stray newline

recipes-app/switchserialmode/files/src/switchserialmode.c
```diff
@@ -237,20 +297,56 @@ void gpio_set(const char *gpiodev, const char *offset, const char *value)
     exe_shell(cmd_set_gpio_value, LEN(cmd_set_gpio_value), NULL, 0);
 }
 
+static int uart_ctrl_info_init(void)
+{
+    int i;
+    int offset = 4;//4 5 6 7
```
jan-kiszka, line 303: Forgotten debug values?

recipes-app/switchserialmode/files/src/switchserialmode.c
```diff
@@ -221,6 +244,43 @@ int exe_shell(const char **lines, int lines_num, char *res, int res_size)
...
+    snprintf(buf, sizeof(buf) - 1, "%s/gpio/gpiochip*/base", device_path);
+    base_file = file_unglob(buf);
+    if (!base_file)
+        return -1;
+
+    fh = fopen(base_file, "r");
+    if (!fh)
+        return -1;
+
+    ret = fread(buf, 1, sizeof(buf) - 1, fh);
```
jan-kiszka, line 274: you could also use fscanf.

jan-kiszka:
This became obsolete via #95.

PR #53, opened 2020-10-10
using linux already defined interface,delete some unused codes

AsuraZeng:


jan-kiszka:
Is this now aligned with what upstream is accepting? I makes sense to update the patch once it is merged, by simply cherry-picking the final version. Before that, an update makes little sense.

AsuraZeng:
> Is this now aligned with what upstream is accepting? I makes sense to update the patch once it is merged, by simply cherry-picking the final version. Before that, an update makes little sense.

I just found there were some redundant self-defined functions in this patch.
Just replace the function with the one has already defined in  the cp210x. I think this is more in line with programming principles.

jan-kiszka:
The patches in usb-next (https://git.kernel.org/pub/scm/linux/kernel/git/johan/usb-serial.git/log/?h=usb-serial-5.12-rc1) should rather be used.

PR #61, opened 2020-12-04
Firmware, kernel, npm, Isar, kas updates

jan-kiszka:
Specifically allows us to move forward with upstreaming U-Boot as it acts as in integration reference and source for binary-only artifacts.


conf/distro/debian-buster-backports.list
```diff
@@ -0,0 +1,2 @@
+deb	http://deb.debian.org/debian buster-backports main
+deb-src	http://deb.debian.org/debian buster-backports main
```
BaochengSu, line 2: should the backports be added into snapshot list?

jan-kiszka: I can add it there as placeholder - reproducing the new npm version from the current snapshot date is not working, obviously. But it should be updated on next freeze, yet.

BaochengSu: It's OK, we can schedule the modification by next release, currently just a placeholder or even a todo comment is sufficient.

jan-kiszka: Done in the updated version.

PR #66, opened 2020-12-15
add the sysfw ABI3.x support

AsuraZeng:


recipes-kernel/linux/files/0018-add-the-sysfw-ABI3.X-support.patch
```diff
@@ -0,0 +1,74 @@
+From eee32b0cfc5eaa8419b5f73770a34bd2aadd9318 Mon Sep 17 00:00:00 2001
+From: Chao Zeng <chao.zeng@siemens.com>
+Date: Tue, 15 Dec 2020 14:33:34 +0800
+Subject: [PATCH 18/18] add the sysfw ABI3.X support
+
+Signed-off-by: Chao Zeng <chao.zeng@siemens.com>
+---
+ arch/arm64/boot/dts/siemens/Makefile             |  2 +-
+ .../boot/dts/siemens/iot2050-advanced-abi3.x.dts | 16 ++++++++++++++++
+ .../boot/dts/siemens/iot2050-basic-abi3.x.dts    | 16 ++++++++++++++++
```
jan-kiszka, line 10: iot2050-{basic,advanced}.dts should imply abi3.x. Use special names only for the legacy dtbs. Just like we do for the basic dtb (siemens/iot2050-basic-oldfw.dtb).

BaochengSu: Question is that when using transition kernel with the released old u-boot, the dtb file is recgnized to "iot2050-basic.dtb" or "iot2050-advanced.dtb", which is using the new 3.x abi.

jan-kiszka: We need to following logic in the wks file:

if env exists fdtfile -> new firmware, new sysfw
else if board_name = advanced -> setenv fdtfile advanced-oldfw.dtb
​else setenv fdtfile basic-oldfw.dtb

That's the existing one plus detection of legacy fw on advanced devices.

jan-kiszka: ...and all this legacy stuff will then be removed without traces in the next-but-one image release. Think this through from that perspective.

AsuraZeng: could the environment `fdtfile`  act as a flag to distinguish new firmware , new sysfw ? 
I have seen the the fdtfile has been set in the upstream u-boot according to the board type. am I right?

jan-kiszka: Exactly, see also f5c53afb9b82f748a793a8620bee475c1e5c314b.

AsuraZeng: Ok I get it. thanks

wic/iot2050.wks
```diff
@@ -8,6 +8,6 @@
 # COPYING.MIT file in the top-level directory.
 #
 
-part / --source rootfs-u-boot --sourceparams="no_initrd=yes,script_prepend=env exists fdtfile || if test \\\"\${board_name}\\\" = \\\"IOT2050-ADVANCED\\\"; then set fdtfile siemens/iot2050-advanced.dtb; else setenv fdtfile siemens/iot2050-basic-oldfw.dtb; fi" --fstype ext4 --label rootfs --align 1024 --use-uuid
+part / --source rootfs-u-boot --sourceparams="no_initrd=yes,script_prepend=env exists fdtfile || if test \\\"\${board_name}\\\" = \\\"IOT2050-ADVANCED\\\"; then set fdtfile siemens/iot2050-advanced-oldfw.dtb; else setenv fdtfile siemens/iot2050-basic-oldfw.dtb; fi" --fstype ext4 --label rootfs --align 1024 --use-uuid
```
jan-kiszka, line 11: That's not yet the logic I sketched. This will select "oldfw" as dtb for the advanced board also over the new firmware. Note the `||`-

AsuraZeng: the `||` , new firmware exists the env fdtfile , so It would not execute the following code .
I think it would set the fdtfile = iot2050-advanced.dtb not iot2050-advanced-oldfw.dtb

recipes-kernel/linux/files/0018-add-the-sysfw-ABI3.X-support.patch
AsuraZeng: But I have another question. I see there is a iot2050-basic-oldfw.dts, it is compat the V01.0.0.1 . and could it able to Isolate this file?

wic/iot2050.wks
jan-kiszka: You are right, the whole if-statement will not be run. All should be fine then.

recipes-kernel/linux/files/0018-add-the-sysfw-ABI3.X-support.patch
jan-kiszka: Sorry, I do not understand the question yet.

jan-kiszka:
Just tried on Basic with the upstream U-Boot, but networking does not work and leads to the usual kernel panic. From the file name POV, the right DTB seems to be loaded.

BaochengSu:
> Just tried on Basic with the upstream U-Boot, but networking does not work and leads to the usual kernel panic. From the file name POV, the right DTB seems to be loaded.

same issue here, investigating...

jan-kiszka:
FWIW, I'm now using https://groups.google.com/d/msgid/isar-users/a7225f82-a176-adfd-9e07-98dc745ca0cd%40siemens.com to check which artifacts are loaded by u-boot.

PR #68, opened 2020-12-17
kernel: load pru firmware from filesystem

BaochengSu:
Loading pru fw from mtd is insecure, drop the 8th patch to restore the
default behavior that load the firmware from file system.


jan-kiszka:
Did anything happen to 0015-fix-PLL4_DCO-freq-over-range-cause-DP-not-display.patch so that git does not detect it as renamed? Looks fine otherwise.

BaochengSu:
> Did anything happen to 0015-fix-PLL4_DCO-freq-over-range-cause-DP-not-display.patch so that git does not detect it as renamed? Looks fine otherwise.

It only fixes some trailing and leading whitespace error, which is really anonying during git am.

PR #70, opened 2020-12-21
refine basic board dts

AsuraZeng:
as released basic board has ospi compat settings.
Refined the basic dts to compat the released basic board and currently sysfw abi3.x
about opsi


BaochengSu:
@jan-kiszka, Do you think if it's a good idea to use device tree overlay here?

jan-kiszka:
Well, we could do that as well, model the deviation of the original firmware as overlays to the upstream firmware dtbs. We have support for loading overlays via u-boot in u-boot-script. OTOH, it does no simplify anything, just marginally reduces the size of distributed dtb artifacts.

PR #73, opened 2020-12-22
fix: using tab instead of spaces in dts

BaochengSu:


BaochengSu:
An overall refact will follow later, close this one.

PR #74, opened 2020-12-22
Refine kernel patch

BaochengSu:
1. Put IOT2050 dts together with ti's. Rename them according to the upstreaming patch
2. Put all device tree change to the first patch
2. Adjust some formats, using tab instead of spaces.
3. Use https to download code from TI


AsuraZeng:
naming follow upstream is ok but  I don't think it's a good way to integrate  all of dts to the ti folder. 
My reason is that in the ti folder are the abstract of ti related　and we should keep the siemens folder to indicate that
this is siemens  board not TI's board

jan-kiszka:
Not our invention: Upstream sorts DTs according to chip vendors - that's why you find the Raspberry boards under broadcom.

PR #76, opened 2020-12-29
iot2050: organizing the recipes of u-boot in the form of patch

AsuraZeng:
take the u-boot commit:51f65b506f37252acb3cd4184ef5e1fc20da13a2 as the
base.patch the modify for our board

### DIFF
--- a/recipes-bsp/u-boot/u-boot-iot2050-2021.x-upstream.inc
+++ b/recipes-bsp/u-boot/u-boot-iot2050-2021.x-upstream.inc
@@ -11,12 +11,38 @@
 require u-boot-iot2050.inc
 
 SRC_URI += " \
-    git://github.com/siemens/u-boot.git;protocol=https;branch=${U_BOOT_BRANCH};rev=${U_BOOT_BRANCH} \
+    https://gitlab.denx.de/u-boot/u-boot/-/archive/${U_BOOT_REV}/u-boot-${U_BOOT_REV}.tar.gz; \
+    file://upstream/0001-arm-dts-Add-IOT2050-device-tree-files.patch \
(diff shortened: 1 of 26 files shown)


BaochengSu:
Change the header of the commit message, remove the leading "iot2050:" and capitalize the first word.

BaochengSu:
Also note in commit message this is only for upstream u-boot and it's only temporary, all the patches will be deleted after our upstreaming work is done.

recipes-bsp/u-boot/u-boot-iot2050-2021.x-upstream.inc
```diff
@@ -11,10 +11,36 @@
 require u-boot-iot2050.inc
 
 SRC_URI += " \
-    git://github.com/siemens/u-boot.git;protocol=https;branch=${U_BOOT_BRANCH};rev=${U_BOOT_BRANCH} \
+    git://github.com/u-boot/u-boot.git;protocol=https;branch=${U_BOOT_BRANCH};rev=${U_BOOT_REV} \
```
BaochengSu, line 14: I don't think it's a good idea to use the github fork. Instead we should use our siemens fork or use the gitlab.denx.de directly.

Also considering the git clone speed, better to use the https://gitlab.denx.de/u-boot/u-boot/-/archive/51f65b506f37252acb3cd4184ef5e1fc20da13a2/u-boot-51f65b506f37252acb3cd4184ef5e1fc20da13a2.tar.gz, it's only 20MB in size.

BaochengSu: Or use https://github.com/siemens/u-boot/archive/51f65b506f37252acb3cd4184ef5e1fc20da13a2.zip, speed is almost the same for me between github and denx.de. But not sure which one is better for somebody else.

PR #78, opened 2020-12-29
refine kernel device tree

BaochengSu:
Extract oldfw related changes to k3-am65-iot2050-oldfw.dtsi, includes
uart dma workaround and the clock swapping of sdhci. Also change the
includings correspondingly.

Remove the word 'common' from iot2050.dtsi, sarcasticly our boards have
a lot that not in common.

Change the compatible and model string according to Jan's upstreaming
patch.

Change the dr_mode of usb1 to `host`.

Change board name in iot2050setup.py and board-configuration

### DIFF
--- a/recipes-kernel/linux/files/0001-iot2050-add-iot2050-platform-support.patch
+++ b/recipes-kernel/linux/files/0001-iot2050-add-iot2050-platform-support.patch
@@ -1,4 +1,4 @@
-From f3cff2d5e56dcc2287d1689894b576e87c98562c Mon Sep 17 00:00:00 2001
+From ed5b2c3e07b04ac89344af3b118ece6650666cfd Mon Sep 17 00:00:00 2001
 From: Le Jin <le.jin@siemens.com>
 Date: Mon, 18 Nov 2019 17:58:08 +0800
 Subject: [PATCH 01/14] iot2050: add iot2050 platform support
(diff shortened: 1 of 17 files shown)


recipes-kernel/linux/files/0001-iot2050-add-iot2050-platform-support.patch
```diff
@@ -41,24 +45,70 @@ index b952d3e730d6..3f6fc2f8445c 100644
...
++
++&sdhci0 {
++	clocks = <&k3_clks 48 0>, <&k3_clks 48 1>;
++	assigned-clocks = <&k3_clks 48 1>;
++	assigned-clock-rates = <142860000>;
++};
++
++&sdhci1 {
++	clocks = <&k3_clks 47 0>, <&k3_clks 47 1>;
++};
```
AsuraZeng, line 92: the clock swapping of sdhci is only for release basic board..
not for all released board

recipes-kernel/linux/files/0001-iot2050-add-iot2050-platform-support.patch
```diff
@@ -797,24 +832,56 @@ index 000000000000..c585fdfcc5cc
...
+@@ -0,0 +1,12 @@
++// SPDX-License-Identifier: GPL-2.0
++/*
++ * (C) Copyright 2018-2020 Siemens AG
++ */
++
++/dts-v1/;
++
++#include "k3-am65-iot2050.dtsi"
++#include "k3-am6548-iot2050-basic.dtsi"
```
AsuraZeng, line 865: the dts name should am6528 for basic

recipes-kernel/linux/files/0001-iot2050-add-iot2050-platform-support.patch
```diff
@@ -797,24 +832,56 @@ index 000000000000..c585fdfcc5cc
...
+@@ -0,0 +1,10 @@
 +// SPDX-License-Identifier: GPL-2.0
 +/*
-+ * (C) Copyright 2020 Siemens AG
++ * (C) Copyright 2018-2020 Siemens AG
++ */
++
++/dts-v1/;
++
++#include "k3-am65-iot2050.dtsi"
```
AsuraZeng, line 848: should k3-am65-iot2050.dtsi be included into basic.dtsi or advanced.dtsi ?

recipes-kernel/linux/files/0001-iot2050-add-iot2050-platform-support.patch
BaochengSu: Checked on V01.00.02, this workaround is both for advanced and basic, see

https://github.com/siemens/meta-iot2050/blob/V01.00.02/recipes-kernel/linux/files/0019-feat-set-sdhci0-clock-frequency-to-142.86MHz.patch

```diff
diff --git a/arch/arm64/boot/dts/siemens/iot2050-advanced.dts b/arch/arm64/boot/dts/siemens/iot2050-advanced.dts
index 88b04c974294..9a83edbbeae1 100644
--- a/arch/arm64/boot/dts/siemens/iot2050-advanced.dts
+++ b/arch/arm64/boot/dts/siemens/iot2050-advanced.dts
@@ -43,9 +43,18 @@
 };
 
 &sdhci0 {
+	/*
+	 * Swap clock TISCI clock IDs between sdhci0 and sdhci1 to work
+	 * around an issue in System Firmware 2019.12a (and earlier) known
+	 * as SYSFW-3179.
+	 */
+	clocks = <&k3_clks 48 0>, <&k3_clks 48 1>;
+	assigned-clocks = <&k3_clks 48 1>;
+	assigned-clock-rates = <142860000>;
 	pinctrl-names = "default";
 	pinctrl-0 = <&main_mmc0_pins_default>;
 	bus-width = <8>;
 	non-removable;
 	ti,driver-strength-ohm = <50>;
+	disable-wp;
 };
\ No newline at end of file
diff --git a/arch/arm64/boot/dts/siemens/iot2050-common.dtsi b/arch/arm64/boot/dts/siemens/iot2050-common.dtsi
index 89f354528830..c58ae32547a2 100644
--- a/arch/arm64/boot/dts/siemens/iot2050-common.dtsi
+++ b/arch/arm64/boot/dts/siemens/iot2050-common.dtsi
@@ -585,9 +585,16 @@
 };
 
 &sdhci1 {
+	/*
+	 * Swap clock TISCI clock IDs between sdhci0 and sdhci1 to work
+	 * around an issue in System Firmware 2019.12a (and earlier) known
+	 * as SYSFW-3179.
+	 */
+	clocks = <&k3_clks 47 0>, <&k3_clks 47 1>;
 	pinctrl-names = "default";
 	pinctrl-0 = <&main_mmc1_pins_default>;
 	ti,driver-strength-ohm = <50>;
+	disable-wp;
 };
```

recipes-kernel/linux/files/0001-iot2050-add-iot2050-platform-support.patch
BaochengSu: Modified with new force push.

recipes-kernel/linux/files/0001-iot2050-add-iot2050-platform-support.patch
BaochengSu: seems to me not a good idea to include dtsi in dtsi, better follow my change, one benefit is that the including relationship is clear and straightforward.

jan-kiszka:
Only briefly scanned through it but it looks good so far.

Do we have any differences to https://github.com/siemens/linux/commits/jan/iot2050 with this? Then those should be resolves soon before we lose that with the next rebase.

BaochengSu:
> Do we have any differences to https://github.com/siemens/linux/commits/jan/iot2050 with this? Then those should be resolved soon before we lose that with the next rebase.
Hi Jan,

I change the filename from `am6-iot2050-common.dtsi` to `am6-iot2050.dtsi`, due to for transition kernel, we have oldfw and newfw both supported, and it may looks too much if adding word `common` for oldfw-wise or newfw-wise things, so I just remove that word. However, I think it's ok for the upstreaming using `common` as only sysfw 3.x is supported in upstream. Anyway, this is only a trivial change.

By the way I noticed that the `panic-ind-off` is removed from led-status-green in upstreaming patch. Per the original design, the panic led should blink red if panic happens, and this led is composed by a red led and a green led, and the green led blinks if everything goes right. So if we don't turn off the green led for panic, it would blink amber or in a red-green-red manner. That why the `panic-ind-off` quirk is added.

Due to releasing to our internal system test guys, I will merge this PR, we can continue discussing here.

Oh, and a early happy new year to you.

PR #80, opened 2021-01-04
linux: Fix dtb path error in boot.scr

BaochengSu:
The dtb paths in boot.scr are wrongly swapped between am6528 and am6548.


AsuraZeng (commented):
looks ok from my side.

PR #82, opened 2021-01-08
sr2 soc support

AsuraZeng:


jan-kiszka:
Can't we do the kernel development directly based on our upstream kernel queue?

AsuraZeng:
Indeed,we can uniform kernel for SR1 and SR2 .as sr2 needs to work quickly for thermal testing . so currently we decide to use TI kernel not upstream to do the development. 
when upstream kernel is stable. we would switch to upstream kernel.

the kas option is mainly for u-boot. as SR1 and SR2 have different sysfw, we'd like to generate two type of flash.bin for sr1 and sr2.

jan-kiszka:
Then maybe keep those changes in staging for now, to avoid complications on the ongoing SR1 work? We may also find a better solution for U-Boot before the final release of the new hardware revision.

BaochengSu:
> Then maybe keep those changes in staging for now, to avoid complications on the ongoing SR1 work? We may also find a better solution for U-Boot before the final release of the new hardware revision.

Sure, that's why this pr is targeting to a sr-develop branch, other than the master branch.

BaochengSu (commented):
Why  recipes-bsp/u-boot/files/upstream/0027-mmc-Add-a-saved_clock-member.patch  and recipes-bsp/u-boot/files/upstream/0028-spl-mmc-Fix-spl_mmc_get_uboot_raw_sector-implementat.patch are empty?

recipes-bsp/u-boot/u-boot-iot2050-2021.x-upstream.inc
```diff
@@ -2,7 +2,7 @@
 # Copyright (c) Siemens AG, 2020
 #
 # Authors:
-#  Jan Kiszka <jan.kiszka@siemens.com>
```
BaochengSu, line 5: Instead changing the author, please add new author.

recipes-kernel/linux/linux-iot2050_5.x-ti-kernel.bb
```diff
@@ -0,0 +1,34 @@
...
+    file://ti-kernel/0007-feat-extend-led-panic-indicator-on-and-off.patch \
+    file://ti-kernel/0008-feat-change-mmc-order-using-alias-in-dts.patch \
+    file://ti-kernel/0009-Workaround-to-correct-DP-clock-to-154MHZ.patch"
+
+SRC_URI += " \
+    file://${KERNEL_DEFCONFIG} \
+    file://iot2050-upstream.cfg \
+    file://iot2050_defconfig_extra.cfg"
+
+KERNEL_REV = "07.01.00.006"
```
BaochengSu, line 30: Personally I prefer to the commit-ish rather than the tag. The assumption that git tag never change is not always true.

recipes-kernel/linux/linux-iot2050_5.x-ti-kernel.bb
```diff
@@ -0,0 +1,34 @@
+#
```
BaochengSu, line 1: Do we know the exact kernel version? Better to use that exact version instead of the blurry 5.x.

BaochengSu: Also better to keep the same naming schema with the current `linux-iot2050_4.19.94+.bb`

recipes-kernel/linux/linux-iot2050_5.x-ti-kernel.bb
```diff
@@ -0,0 +1,34 @@
...
+# This file is subject to the terms and conditions of the MIT License.  See
+# COPYING.MIT file in the top-level directory.
+#
+
+require recipes-kernel/linux/linux-custom.inc
+
+SRC_URI += " \
+    https://git.ti.com/cgit/ti-linux-kernel/ti-linux-kernel/snapshot/ti-linux-kernel-${KERNEL_REV}.tar.xz;protocol=https; \
+    file://ti-kernel/0001-iot2050-add-iot2050-platform-support.patch \
+    file://ti-kernel/0002-Add-support-for-U9300C-TD-LTE-module.patch \
```
BaochengSu, line 16: In my opinion, the word `ti-kernel` does not express the our exact intent very clearly.

recipes-bsp/u-boot/u-boot-iot2050-2021.x-upstream.inc
```diff
@@ -54,7 +80,13 @@ DEBIAN_BUILD_DEPENDS =. "trusted-firmware-a-iot2050, optee-os-iot2050, \
 
 do_prepare_build_append() {
     cd ${S}
-    ln -sf ../prebuild/* ${S}
+    if [ ${SOC_VERSION} = "sr1" ]; then
+        ln -sf ../prebuild/prufw_sr1/* ${S}
+        ln -sf ../prebuild/sysfw_sr1/* ${S}
```
BaochengSu, line 85: Just curious, we do have another am65x-sr2-txpru0-prueth-fw.elf and am65x-sr2-txpru0-prueth-fw.elf for kernel, but not for u-boot. Doesn't u-boot need it?

recipes-kernel/linux/files/ti-kernel/0001-iot2050-add-iot2050-platform-support.patch
```diff
@@ -0,0 +1,934 @@
...
++/*
++ * (C) Copyright 2019 Siemens AG
++ */
++#include <dt-bindings/input/input.h>
++#include <dt-bindings/net/ti-dp83867.h>
++#include <dt-bindings/phy/phy.h>
++
++/ {
++	compatible =  "siemens,iot2050-common", "ti,am654";
++	model = "SIMATIC IOT2050 Common";
```
BaochengSu, line 63: This is meaningless.

kas/opt/sr2-select.yml
```diff
@@ -0,0 +1,16 @@
+#
```
BaochengSu, line 1: Please document how to use these kas options.

recipes-kernel/linux/linux-iot2050_5.x-ti-kernel.bb
AsuraZeng: This is from snapshot of the TI archive. I don't think it would change or deleted by TI

recipes-kernel/linux/linux-iot2050_5.x-ti-kernel.bb
AsuraZeng: it clearly express the intents, it based ti kernel.
or What better advice do you have?

recipes-bsp/u-boot/u-boot-iot2050-2021.x-upstream.inc
AsuraZeng: I don't know. I don't have tested the eth function under u-boot.
The problem needs to be investigated

recipes-kernel/linux/files/ti-kernel/0001-iot2050-add-iot2050-platform-support.patch
AsuraZeng: yes,but this makes it easier for the user to understand its intentions

jan-kiszka:
> Sure, that's why this pr is targeting to a sr-develop branch, other than the master branch.

Ah, sorry, missed that.

AsuraZeng:
> Why recipes-bsp/u-boot/files/upstream/0027-mmc-Add-a-saved_clock-member.patch and recipes-bsp/u-boot/files/upstream/0028-spl-mmc-Fix-spl_mmc_get_uboot_raw_sector-implementat.patch are empty

I have checked it,the change have modified in our branch. It only have commit message.
 possibly the upstream u-boot have the modification.

recipes-kernel/linux/linux-iot2050_5.x-ti-kernel.bb
BaochengSu: We already have a ti-kernel targeting at 4.19.94. The target of this work is for the sr2 soc, so I think `sr2` would be a better name, or something else that help people understand this is for sr2.

recipes-bsp/u-boot/u-boot-iot2050-2021.x-upstream.inc
BaochengSu: Sure, anyway atm eth is not on the critical path to sr2 development, we could address that later.

recipes-kernel/linux/files/ti-kernel/0001-iot2050-add-iot2050-platform-support.patch
BaochengSu: More or less...Anyway we will switch to the upstreaming kernel in the future, so looks fine at this moment.

recipes-kernel/linux/linux-iot2050_5.4.74+.bb
```diff
@@ -0,0 +1,34 @@
...
+    file://5.x-patch/0007-Workaround-to-correct-DP-clock-to-154MHZ.patch \
+    file://5.x-patch/0008-feat-add-io-expander-pcal9535-support.patch \
+    file://5.x-patch/0009-add-the-gpio-alias-support.patch"
+
+SRC_URI += " \
+    file://${KERNEL_DEFCONFIG} \
+    file://iot2050-upstream.cfg \
+    file://iot2050_defconfig_extra.cfg"
+
+KERNEL_REV = "07.01.00.006"
```
BaochengSu, line 30: Please check 9574bba32a1898794895ca3816e815154c80226d

kas/opt/5.x-kernel.yml
```diff
@@ -0,0 +1,13 @@
...
+# This file is subject to the terms and conditions of the MIT License.  See
+# COPYING.MIT file in the top-level directory.
+#
+
+header:
+  version: 10
+
+local_conf_header:
+  ti-kernel-versions: |
+    PREFERRED_VERSION_linux-iot2050 = "5.4.%"
```
BaochengSu, line 13: Better to enclose this into SOC_VERSION, so that we can use an unified option yml file for both u-boot and kernel.

PR #83, opened 2021-01-12
linux: Enable driver for Qualcomm MSM based 3G and LTE modems

jan-kiszka:
Requested by several users already.


BaochengSu:
Hi Jan,

We plan to freeze the code in next couple days, for the next release, do you have something else that need be merged into next release?

jan-kiszka:
Adding GPIO line names and enabling libgpiod support would have been nice, but I may not find enough time to do that.

Key element of the new release is the capability to boot from upstream-based firmware. As long as that is given, the release is on the right track.

PR #85, opened 2021-01-15
feat: enable soc device ID driver

AsuraZeng:
This allows Texas Instruments Keystone 3 SoCs to identify
specifics about the SoC in use

### DIFF
--- a/recipes-bsp/u-boot/files/upstream/0001-arm-dts-Add-IOT2050-device-tree-files.patch
+++ b/recipes-bsp/u-boot/files/upstream/0001-arm-dts-Add-IOT2050-device-tree-files.patch
@@ -1,7 +1,7 @@
-From 7ad97960dface3b377afcbb1e80fd99141943293 Mon Sep 17 00:00:00 2001
+From ef42ed1ce4ba33de0e883a6632a541248b31dabe Mon Sep 17 00:00:00 2001
 From: Jan Kiszka <jan.kiszka@siemens.com>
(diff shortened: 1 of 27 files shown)


jan-kiszka:
What use cases does this enable? I'd like to pick that up for upstream as well, just like to understand it first.

AsuraZeng:
hi jan
this is enable `soc_device_match` function in the u-boot to distinguish sr1 or sr2 soc.

BaochengSu (approved):

jan-kiszka:
Yes, but who is using this? Just a developer sitting in front of the U-Boot command line?

BaochengSu:
> Yes, but who is using this? Just a developer sitting in front of the U-Boot command line?

function `omap_usb2_phy_probe` in `drivers/phy/omap-usb2-phy.c`, this enables a workaround for PG1.0 silicon bug regarding usb, see details in the comment below: 

```c
	/*
	 * AM654x PG1.0 has a silicon bug that D+ is pulled high after
	 * POR, which could cause enumeration failure with some USB hubs.
	 * Disabling the USB2_PHY Charger Detect function will put D+
	 * into the normal state.
	 *
	 * Enable this workaround for AM654x PG1.0.
	 */
	if (soc_device_match(am65x_sr10_soc_devices))
		priv->flags |= OMAP_USB2_DISABLE_CHG_DET;
```

jan-kiszka:
OK, that makes it clear. Please write proper commit logs, always.

PR #87, opened 2021-01-19
Activate github workflows

jan-kiszka:
They lifted the time limit per job to 6 hours which even permits running the example image build followed by the RT image in a row (reusing build artifacts). This also enables keeping the artifacts (up to 90 days) for direct testing purposes.

In addition, this comes with an Isar update, getting us down to zero patches again.


BaochengSu:
Great work!

.github/workflows/main.yml
```diff
@@ -0,0 +1,41 @@
...
+
+jobs:
+  images:
+    name: Images
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Example image
+        run: ./kas-docker build kas-iot2050-example.yml
```
BaochengSu, line 13: Just curious, why not use kas-container directly?

.github/workflows/main.yml
```diff
@@ -0,0 +1,41 @@
...
+      - name: Upload example image
+        uses: actions/upload-artifact@v2
+        with:
+          name: iot2050-example-image
+          path: build/tmp/deploy/images/iot2050/iot2050-image-example-isar-iot2050.wic.img*
```
BaochengSu, line 18: How many disk spaces do we have?

.github/workflows/main.yml
jan-kiszka: Oops, legacy

.github/workflows/main.yml
jan-kiszka: No idea. I didn't find notes on that on github yet. Maybe you can try to search as well. But if there is a limit: There is also no hint how much you used already.

BaochengSu: > No idea. I didn't find notes on that on github yet. Maybe you can try to search as well. But if there is a limit: There is also no hint how much you used already.

I tried to search it but didn't find anything, I am wondering if we are using the organization storage space...

PR #88, opened 2021-01-20
u-boot: Add NOR protection support for Winbond chips

BaochengSu:
IOT2050 use a Winbond chip, and it supports lock/unlock function,
so that user can use `sf protect lock` and `sf protect unlock` commands
to lock/unlock the protection, so that the content of the flash could
be protected whenever needed.

This commit is to enable the NOR protect support for Winbond chips.


jan-kiszka:
NAK, please go upstream first. This is no critical feature, so we can wait for U-Boot to accept it.

PR #89, opened 2021-01-20
u-boot: Enable flash protection for IOT2050

BaochengSu:
The flash chip of IOT2050 is a Winbond chip and can reuse the same
logic of flash protection of STMICRO chips.

User can use `sf protect lock` and `sf protect unlock` commands to lock
& unlock the flash.

### DIFF
--- a/recipes-bsp/u-boot/files/upstream/0027-iot2050-Add-support-for-flash-protect.patch
+++ b/recipes-bsp/u-boot/files/upstream/0027-iot2050-Add-support-for-flash-protect.patch
@@ -1,0 +1,32 @@
+From fbb415a6ad15195146ef1f2720032ba17555d38c Mon Sep 17 00:00:00 2001
+From: Su Baocheng <baocheng.su@siemens.com>
+Date: Wed, 20 Jan 2021 16:52:19 +0800
+Subject: [PATCH 27/28] iot2050: Add support for flash protect
+
+IOT2050 use a Winbond flash that support lock & unlock, but it was not
+enabled.
+
+This winbond flash is a similar one to STMICRO, and can reuse the lock
+unlock logic for STMICRO chips. So by adding CONFIG_SPI_FLASH_STMICRO
+to iot2050_defconfig, the flash protection could be supported.
+
+Signed-off-by: Su Baocheng <baocheng.su@siemens.com>
+---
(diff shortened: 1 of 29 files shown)


jan-kiszka:
As I said: Let's pick up U-Boot review first by sending them the patch. The feature is not urgently needed, even more as we have the write-protect jumper in addition.

BaochengSu:
> As I said: Let's pick up U-Boot review first by sending them the patch. The feature is not urgently needed, even more as we have the write-protect jumper in addition.

According to [Winbond datasheet](https://www.winbond.com/resource-files/w25q128jv%20revg%2004082019%20plus.pdf), the write protect pin should work together with the BP{0-2} and TB to provide write protection. It doesn't work by its own. Have verified it on board.

This is a security feature and is written in our TRA. I am afraid we have to integrate this patch for next release.

I will start the upstreaming work for this patch, and let's see if it can be accepted by upstream before our next release.

recipes-bsp/u-boot/files/upstream/0027-iot2050-Add-support-for-flash-protect.patch
```diff
@@ -0,0 +1,32 @@
...
+Date: Wed, 20 Jan 2021 16:52:19 +0800
+Subject: [PATCH 27/28] iot2050: Add support for flash protect
+
+IOT2050 use a Winbond flash that support lock & unlock, but it was not
+enabled.
+
+This winbond flash is a similar one to STMICRO, and can reuse the lock
+unlock logic for STMICRO chips. So by adding CONFIG_SPI_FLASH_STMICRO
+to iot2050_defconfig, the flash protection could be supported.
+
```
jan-kiszka, line 12: If there is a technical dependency between CONFIG_SPI_FLASH_WINBOND and CONFIG_SPI_FLASH_STMICRO now (or after patch 28), that needs to be expressed via Kconfig instead. If we only use STMICRO now, WINBOND must be dropped.

BaochengSu: The flash on our board is Winbond, but it follows the same logic of flash protection with ST Micro flash.

Some in-tree boards also does things like this.

PR #92, opened 2021-01-28
Build fixes

jan-kiszka:
Follows the openssl update in buster and specifically resolves the npm breakages by updating the class as well as pulling most npm and node packages from bullseye. The latter may soon be changed again if buster-backports contains the same version and is generally fixed.

Note that this might be in conflict with a snapshot taken previously against the old npm scheme. Likely needs discussion how to handle that. Keep the npm-buildchroot recipe for now?


BaochengSu:
> Note that this might be in conflict with a snapshot taken previously against the old npm scheme. Likely needs discussion how to handle that. Keep the npm-buildchroot recipe for now?

snapshot taken by package-lock.yml?

jan-kiszka:
Yes, `kas/opt/package-lock.yml`.

BaochengSu:
Per my understanding, there are too much combinations of the snapshot timestamp + our commits, so it would be impossible to cover them all. 

I think we are safe as long as we can reproduce our release image with the package-lock.yml at our V tag.

So if ATM we take a snapshot, it can pass the build, then we are fine for next release.

BaochengSu (approved):

PR #96, opened 2021-02-01
Prepare package locking for next release

BaochengSu:
1. Debian packages are locked to snapshot of UTC 2021-01-28 00:00:00
2. node-red nodes are updated to latest version
3. tcf-agent is updated to latest master
4. op-tee is updated to latest version 3.12.0

npm-shrinkwrap.json files of nodes are generated with the 7.4.3 npm, so now they are using `"lockfileVersion": 2,`.


jan-kiszka:
checked also locally - looked good

PR #111, opened 2021-03-18
Chao/pg2 dev

AsuraZeng:


recipes-app/iot2050setup/files/iot2050setup.py
```diff
@@ -481,6 +481,8 @@ def configureExternalSerialMode(self):
             return
         terminateOpt = ' -t' if self.terminateStatus == 'on' else ''
         subprocess.call('switchserialmode -m ' + switchMode + terminateOpt, shell=True)
+        if self.topmenu.boardType == 'IOT2050 Advanced':
+            subprocess.call('switchserialmode -r ', shell=True)
```
BaochengSu, line 485: ATM the final board type for PG2 has not been locked yet, so I think it's ok to use the `Advanced` now.

However, I think we'd better to add some TBD mark here, just in case we lose it in the future.

BaochengSu: Not sure if we use the registers to determine SR2 soc is a good way here.

I think this work should be done by the u-boot then u-boot handle down the information to linux.

AsuraZeng: actually, I found no matter in the uboot or kernel, they take the similar way to identify sr1 or sr2 soc to do different actions.

AsuraZeng: And I think if we pass the info from kernel to user space.it needs more effort to expose the node to user space. Currently,I think this is a simple way to solve it.
In the future,we can distinguish the board by board-mode.I will mark this as TBD in order to check it in the future.

PR #131, opened 2021-05-03
u-boot-iot2050: Update upstream queue to v2021.04

jan-kiszka:
This comes with a couple of changes:
 - based on stable release
 - folded defconfig patches into main one
 - added support for boot-selection via USER button
 - cleaned up recipes

Even Ethernet works again (must have been an upstream issue, no related
changes on the patches).


jan-kiszka:
These patches are in line with what I'm planning to submit to u-boot upstream soon, at least regarding baseline and watchdog support (network depends on TI).

recipes-bsp/u-boot/files/upstream/0001-arm-dts-Add-IOT2050-device-tree-files.patch
```diff
@@ -545,6 +520,7 @@ index 0000000000..fcb45d1525
 +		&push_button_pins_default
 +		&db9_com_mode_pins_default
```
AsuraZeng, line 521: There is no related function about DB9 Port under uboot. Should we remove these nodes?

jan-kiszka: The idea is to use the kernel DT, just plus additional properties that U-Boot needs. Makes maintenance easier and was actually requested.

BaochengSu (approved):

PR #132, opened 2021-05-11
regen-rootfs-uuid: regen uuid after expand-on-first-boot.service

AsuraZeng:
Both regen-rootfs-uuid-on-first-boot.service and expand-on-first-boot.service
would operate the boot device.There may be a conflict.

Add dependency to start the regen-rootfs-uuid-on-first-boot.service after expand-on-first-boot.service


jan-kiszka (approved):
Let's give this a try and hope it fixes the seen issue.

PR #136, opened 2021-05-21
Chao/fix 4g module reconnect

AsuraZeng:


jan-kiszka (approved):
Just note - /wrt image release - that this obviously changes package sources.

PR #140, opened 2021-06-02
Rename op-tee-iot2050 to optee-os-iot2050

gylstorffq:
Rename recipe for consistency reasons with upstream.


jan-kiszka:
Something is indeed fishy, trying to understand why it builds here.

But the recipe should be renamed instead.

jan-kiszka:
That's why it works: https://github.com/ilbers/isar/blob/master/meta/recipes-bsp/optee-os/optee-os-custom.inc#L23. There must be some additional local change that breaks this.

Again, renaming the recipe and folder for consistency reasons is still worthwhile.

jan-kiszka (approved):

BaochengSu (approved):

PR #141, opened 2021-06-07
Feature/customizations base/cleanup

gylstorffq:
Split from swupdate branch:
- set hostname to 127.0.1.1
- move firmware to own package to allow reuse in downstream layers


jan-kiszka (approved):
LGTM.

PR #147, opened 2021-06-21
u-boot-iot2050: Switch to upstream 2021.04-based version

jan-kiszka:
The legacy SDK-based version no longer works with the current recipes
anyway. Time to move forward, dropping all related recipes and
artifacts, unifying the build target to a variant-agnostic one.

This unifies and completes the removal by @AsuraZeng.


jan-kiszka:
BTW, I'm planning to refresh the U-Boot baseline to 2021.07[-rcX] soon, also sync'ing the patch queue with the lasted versions submitted again.

Open issues with both current and upcoming version:
 - network only works after Linux booted once, not after cold boot
 - we need to fix the secure memory firewall (affects the binary artifacts only)

BaochengSu (approved):

PR #149, opened 2021-06-29
iot2050-firmware-update: Account for incomplete images

jan-kiszka:
Pad images that are not erase-size aligned and terminate earlier. This
avoids needless updates of the flash with zeros.


jan-kiszka:
Ping,  @BaochengSu

BaochengSu (approved):

PR #152, opened 2021-07-27
linux:add the watchdog support for linux image

AsuraZeng:
TI has removed the device node from k3-am65-mcu.dtsi
we are lacking of watchdog support,so we need adding the function back

the device node refer to ti sdk 06_03_00


jan-kiszka:
There is https://github.com/torvalds/linux/commit/6674a90bb141496f56ce840c22f2f2ddccc5d397 - not compatible with the TI kernel?

Also, did you check if the downstream wdt driver is up to speed with the mainline driver? There were several changes in upstream in order to make the essential pick-up of an already running watchdog (started by U-Boot) work.

jan-kiszka:
And `TI_SCI_PD_EXCLUSIVE` breaks over https://github.com/siemens/k3-rti-wdt.

AsuraZeng:
> There is [torvalds/linux@6674a90](https://github.com/torvalds/linux/commit/6674a90bb141496f56ce840c22f2f2ddccc5d397) - not compatible with the TI kernel?
> 
> Also, did you check if the downstream wdt driver is up to speed with the mainline driver? There were several changes in upstream in order to make the essential pick-up of an already running watchdog (started by U-Boot) work.

for the kernel 4.19.94, there is no this dts node.

downstream wdt driver has fall behind serval commit with upstream wdt driver.

is it necessary to pick up upstream wdt driver to the currently transitional kernel?

jan-kiszka:
Depends on what purpose the watchdog should fulfill. But if it shall serve e.g. #115, a clear yes.

AsuraZeng:
> Depends on what purpose the watchdog should fulfill. But if it shall serve e.g. #115, a clear yes.
> 
> In general: What is in upstream has settled. What is in the downstream kernel is still under development or even experimental. Even worse, if updating of a downstream branch was stopped.

for transitional kernel, as lacking of the wdt fucntion,just to add the fucntion back and keep the same with Released image.

so for this function, I think there is no need to pick up the difference from upstream.

AsuraZeng:
by the way, for SR2 soc, TI has solved the silicon bug for watchdog.
if so, does it mean it is not necessary rti-wdt firmware on r5f core to support watchdog?

jan-kiszka:
OK if this is just about restoring the (limitedly useful) support in the first release as-is. The transition kernel will not live very long, and the next kernel will be upstream based, code-wise aligned.

Regarding SR2 WDT: I strongly assume that we no longer need that firmware on the R5 core, but I didn't have the time yet to validate this. U-Boot will likely need some runtime detection, to avoid loading the firmware when that's not needed.

AsuraZeng:
> OK if this is just about restoring the (limitedly useful) support in the first release as-is. The transition kernel will not live very long, and the next kernel will be upstream based, code-wise aligned.
> 
> Regarding SR2 WDT: I strongly assume that we no longer need that firmware on the R5 core, but I didn't have the time yet to validate this. U-Boot will likely need some runtime detection, to avoid loading the firmware when that's not needed.


As far as I know, WDT function has not ready for SR2,so currently we still need R5F core firmware until TI has completed the hardware WDT fucntion

BaochengSu:
Echo to Jan, we have the watchdog enabled in upstream u-boot, does this conflict with that one, any test performed on the sysfw 3.0 firmware?

jan-kiszka:
We do not start the watchdog in U-Boot by default, but we prepare that our users could do that. In that case, the kernel must be ready for it. That's not yet true, at least due to the power settings. In any case, it should be tested.

AsuraZeng:
> We do not start the watchdog in U-Boot by default, but we prepare that our users could do that. In that case, the kernel must be ready for it. That's not yet true, at least due to the power settings. In any case, it should be tested.

for sr2 kernel, TI has removed the process loading the  rti_dwwdtest firmware to the R5F core on the kernel side 
can we started the watchdog under uboot and kernel just follow it?

if not, we need to  pick up  the rti_dwwdtest firmware loading process on the kernel side.

jan-kiszka:
`rti_dwwdtest` was a `test` application. https://github.com/siemens/k3-rti-wdt is what we use today (maybe time to clean up the image). If that is loaded there, Linux cannot load its own thing anymore.

AsuraZeng:
if we like to load the  rti_dwwdtest on the uboot, so it not necessary to load the application on the linux?
so in order to implement the watchdog function on the PG2, we need to start the watchdog on the uboot firstly?
Am I right?

jan-kiszka:
Again, forget about rti_dwwdtest - that was never intended to be used for what we used it. And was missing a lot of protection needed for a real watchdog reset mechanism.

If the RTI watchdog does not trigger a hardware reset without R5 help (and I still need to check that on PG2), yes, we need to load that l4-rti-wdt binary via rproc. That has to happen before the watchdog is started. If U-Boot starts it (required for SWUpdate scenarios), it will happen there. If only Linux starts the watchdog, the firmware must be loaded via Linux' rproc means. If Linux tries to load it while the watchdog is already running, it will get an error (firmwalls are up, blocking the R5 access).

BaochengSu:
Then if R5 WDG is started in u-boot, kernel will never able to use that watchdog in linux?

jan-kiszka:
No, why should that be the case? The watchdog register interface is not affected by the R5 code working around the missing reset wiring.

All what fails (with a warning) is loading R5 code. But that is not done by the Linux watchdog driver anyway - in contrast to the U-Boot watchdog driver with something like https://patchwork.ozlabs.org/project/uboot/patch/ed2a3a5622b5712d5f16a719104635fa31dc9f0c.1623440557.git.jan.kiszka@siemens.com.

BaochengSu:
Sure, we will test the kernel watchdog with u-boot watchdog loaded.

Meanwhile, I will merge this PR cause this one is urgent for next image release, also it will not make things more worse.

Then we need continue the discussion of the long term solutions.

jan-kiszka:
The patch is not correct as I explained.

PR #153, opened 2021-07-27
linux-iot2050: Add patch to restore 100MBit/s half duplex Ethernet

jan-kiszka:
Works fine here but likely needs systematic testing. To enable that, this PR.


AsuraZeng:
I have queried TI ago， they said they have some bugs on  the sysfw to support 100M half duplex, so they disabled this function.  it would has some problem to work with sysfw 3.x.  We need ti's reconfirmation

jan-kiszka:
Yeah, we definitely need feedback. 10M was added to 5.10-ti now, but all half-duplex modes remain disabled there. 1G half duplex didn't work for me, but 100M did.

We also need to validate our dependency on the PRU firmware version here. I think I checked both the one carried in master as well as the latest one from TI. But as the versions are not printed anywhere, I may have easily made mistakes in this.

BaochengSu:
Is this sufficient or do we need to merge trans-kernel to TI's 4.19 kernel? how do you think?

jan-kiszka:
I can't tell for sure yet. Merging that kernel will surely bring much more changes.

BaochengSu:
OK then, I couldn't find anything that prevent this one be merged. Also we may need check more regarding this issue from TI side.

PR #154, opened 2021-07-29
linux-iot2050: Rebase over 4.19 ti-processor-sdk

jan-kiszka:
This brings at least half-duplex support for prueth, possibly more. The
required patches for sysfw 3.x support now become part of our queue.
Also update the 8250 irq storm fix to the latest mainline version at
this chance.

Lightly tested so far, but HD seems fine. Therefore, I'm closing #153.


AsuraZeng:
dose ti has released the sdk8.0? and I don't know prufw from sdk8.0 could be compatible with currently transitional kernel. so we need more tests about the ethernet function.

jan-kiszka:
We need to re-test Ethernet anyway after this kernel update. At the same time, I'm concerned that our firmware version is getting too old for the changes we found in that kernel branch.

AsuraZeng:
Ok. there are some conflicts with master. needing some rebase with master to keep clean.

jan-kiszka:
oops, one moment...

jan-kiszka:
Watchdog is back with that sdk kernel, but it also carries the wrong power domain settings.

AsuraZeng:
another  topics,if we'd like load the rti-wdt.fw under uboot, where would we put the fw? 
resides in flash?

jan-kiszka:
Yes, will be part of the U-Boot proper fit image (thus signed as well in the future). About to send out v4 of the U-Boot patches.

AsuraZeng (commented):
looks good to me

BaochengSu (approved):

jan-kiszka:
We have a problem with that legacy watchdog driver in the kernel. It seems to not support hand-over from U-Boot. Will look into backporting the upstream version.

PR #159, opened 2021-08-12
iot2050setup:shield the info and debug level output

AsuraZeng:
as the info and debug level would flush the iot2050setup UI.
it is very hard to use. so shield the releated info
when we use this tool.


jan-kiszka:
Maybe make this dependent on whether the tool is running on a local console? We do not need this when connected via ssh, right?

BaochengSu:
Use `ps -o comm= -p $PPID` to check if the current session is child of sshd, or check the `SSH_TTY` or `SSH_CLIENT`, refer to `printenv | grep SSH`

recipes-app/autoresize-console-window/autoresize-console-window_0.1.bb
```diff
@@ -0,0 +1,18 @@
...
+#
+
+SRC_URI = "file://autoresize-console-window.sh"
+
+inherit dpkg-raw
+
+do_install() {
+    install -v -d ${D}/usr/bin/
+    install -v -m 755 ${WORKDIR}/autoresize-console-window.sh ${D}/usr/bin/autoresize-console-window
+}
```
jan-kiszka, line 18: Why as separate recipe? Is it used outside of iot2050setup?

recipes-app/autoresize-console-window/files/autoresize-console-window.sh
```diff
@@ -0,0 +1,15 @@
...
+#  Chao Zeng <chao.zeng@siemens.com>
+#
+# SPDX-License-Identifier: MIT
+
+current_settings=$(stty -g)
+stty raw -echo min 0 time 5
+printf '\0337\033[r\033[999;999H\033[6n\0338' > /dev/tty
+IFS='[;R' read -r _ rows cols _ < /dev/tty
+stty "$current_settings"
+stty cols "$cols" rows "$rows"
```
jan-kiszka, line 15: I wonder if that parsing etc. wouldn't be simpler in python directly. Possibly, there is termios API that even delivers you cols/rows directly.

recipes-app/iot2050setup/files/iot2050setup.py
```diff
@@ -590,14 +595,80 @@ def selectTerminate(self):
...
+                    break
+        finally:
+            # Reset the terminal back to normal cooked mode
+            termios.tcsetattr(fileno, termios.TCSAFLUSH, stty_sav)
+            fcntl.fcntl(fileno, fcntl.F_SETFL, fc_sav)
+
+        rows, cols = list(map(int, re.findall(r'\d+', output)))
+
+        fcntl.ioctl(ttyfd, termios.TIOCSWINSZ,
+                    struct.pack("HHHH", rows, cols, 0, 0))
```
jan-kiszka, line 643: close() for fd and ttyfd are missing.

jan-kiszka:
Looks good otherwise!

PR #160, opened 2021-08-17
iot2050-firmware-update: Update from tarball

BaochengSu:
Update the firmware using a tarball, which includes the firmware
image, its sha256 checksum and a update control file that indicates
which OS version the update package should be operated upon.


recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
```diff
@@ -118,61 +178,65 @@ class FirmwareUpdate(object):
...
+                        os.write(mtd_dev, firmware_content)
+                    else:
+                        print(".", end="")
+                        sys.stdout.flush()
+                    mtd_pos += mtd_erasesize
+                    firmware_size -= mtd_erasesize
+                print()
+                os.close(mtd_dev)
+
+                mtd_num += 1
```
jan-kiszka, line 232: This is the heart of the update tool. Can you refactor this separately for the new feature?

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
```diff
@@ -43,16 +72,47 @@ class FirmwareUpdate(object):
...
-            print("Reading {} failed: {}".format(self.firmware.name, e.strerror))
+            print("Reading {} failed: {}".format(self.firmware_tar.name, e.strerror))
             sys.exit(1)
 
         if file_sha256 != hash_code:
-            FirmwareUpdateException("Image checksum is not correct")
+            FirmwareUpdateException(
+                "Tarball {} checksum is not correct".format(self.firmware_tar.name))
+
+    def verify_tar(self):
```
jan-kiszka, line 87: This seems to miss a force mode.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
```diff
@@ -9,6 +9,30 @@
...
 # COPYING.MIT file in the top-level directory.
 #
+# To use this tool, a <firmware-update-package>.tar.xz and its sha256 checksum file
+# <firmware-update-package>.tar.xz.sha256 are needed.
+#
+# The <firmware-update-package>.tar.xz contains below files:
+#   - firmware.bin: The firmware to update.
+#   - firmware.bin.sha256: The sha256 checksum of the firmware.
+#   - update.conf.json: The update criteria. Currently only the OS image version
+#       is checked.
```
jan-kiszka, line 19: What about providing multiple firmware binaries for different targets? Should the user get a single tarball, and the tool will pick the fitting one? Or will we distribute multiple tarball, and the tool will only reject those that do not fit?

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
```diff
@@ -9,6 +9,30 @@
...
+#             "version_field": "BUILD_ID",
+#             "version_value": "V01.01.00"
+#         },
+#         {
+#             "image_type": "Industrial OS",
+#             "version_field": "VERSION_ID",
+#             "version_value": "2.1.1"
+#         }
+#     ]
+# }
```
jan-kiszka, line 35: Hardware properties are missing here.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
BaochengSu: There will be different binaries for different board, and also different sha256 checksums. All of these artifacts will be put into one tarball, and the tool will pick one depends on the physical board model.

This PR is only targeting next image release of PG1, I will work on the next version of this tool after transition kernel image released to provide these 'enhanced' functionalties.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
BaochengSu: Which one?

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
BaochengSu: I was not sure if it's a good idea to provide that function... 

However, without this, for some advanced users, it's might be not so handy to use this tool to flash a single firmware bin, provided they know what they are doing.

OK, I think it's good to do that, however before the effective opration on the flash, we need ask the user that they must know what they are doing.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
BaochengSu: Sorry, not fully follow you, do you suggest to separate the new tarball feature out of the `FirmwareUpdate` class?

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
jan-kiszka: We will have a chicken-egg issue here if we do not account for the PG2 case already: How do you want to ship a new firmware version for all hardware variants later on when this updater cannot match to PG1?

Also, checksumming, inside and outside of the taball, seems total overkill now, given that has validation mechanisms.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
jan-kiszka: Force it required, already for development purposes, but also for cases that we may have missed.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
jan-kiszka: I only request to factor our the change to this many function into a separate commit, explaining what it changes (replace variable name? or more?). It's hard to review.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
jan-kiszka: Match PG1/2. See above. This is mandatory, or we cannot ship any update for PG1 via this updater tool if that update also covers PG2.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
BaochengSu: We will have an new example release targeting PG2 board, and these change could be live in that version.

Regarding checksum, I think we will only provide one checksum of the tarball on the download page, and let user to do the verifying by themselves. Then it's reasonale to remove all the checksumming from this tool.

jan-kiszka: Let's better do this properly now, rather than regretting having had all measures in place already, ie. already add hw detection, description and binary selection today. Should be simple enough (via fw-env or /sys/firmware/devicetree).

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
```diff
@@ -9,6 +9,30 @@
...
+#   - update.conf.json: The update criteria. Currently only the OS image version
+#       is checked.
+#
+# Example of update.conf.json:
+# {
+#     "update_criteria":[
+#         {
+#             "image_type": "Example Image",
+#             "version_field": "BUILD_ID",
+#             "version_value": "V01.01.00"
```
jan-kiszka, line 27: What's the semantic of this version field? Exact match? Or "greater or equal"?

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
jan-kiszka: To make it more concrete what I'm concerned about: Customer has the upcoming image installed on a PG1 device. We prepared a firmware update for all devices, but our image tool in that version you propose now will either needlessly stumble over that new format ("best" case) or even pick the wrong image from the tarball (worst case) and brick the device.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
BaochengSu: startswith

jan-kiszka: Then let's call it 'min_version'.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
BaochengSu: I think the whitespace change make it hard to review, please toggle the `ignore white space` from the gear icon

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
BaochengSu: python startswith, so maybe "version_prefix"

jan-kiszka: OK, then this is kind of an exact match - why?

BaochengSu: In some cases we need to cover more than 1 minor versions, for example, we need to cover both 1.1.1 and 1.1.2, then we could use 1.1 as the match field.

Other way is to use the standard regex, maybe that's a better solution?

jan-kiszka: No, I mean why not using a min_version semantic? Without that, we may eventually have to release a new firmware package just declare it being compatible with installing it over a newer image version, without actually having to change its code for that.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
```diff
@@ -5,55 +5,90 @@
...
+# 			"description": "[optional] bla bla bla",
+# 			"name": "pg1-basic-vx.y.z.bin",
+# 			"version": "[optional] whatever",
+# 			"target_board": "pg1-basic"
+# 		},
+# 		{
+# 			"description": "[optional] bla bla bla",
+# 			"name": "pg2-advanced-vx.y.z.bin",
+# 			"version": "[optional] whatever",
+# 			"target_board": "pg2-advanced",
```
jan-kiszka, line 32: target_boards - making the value an array. Will save us from copy&pasting the entry for every board group.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
BaochengSu: That could be the case. Do you think to cover min_version semantic is enough here?

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
BaochengSu: OK

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
```diff
@@ -168,14 +184,158 @@ class FirmwareUpdate(object):
...
+
+                # Only choose the firmware that have the target_os configuration identifying with current 
+                # OS info, or the firmware w/o target_os node which means it doesn't care the OS info.
+                if len(target_os) == 0 or self._check_os(target_os, os_info):
+                    res.append(firmware.name)
+
+        return res
+
+
+class UpdateControl(object):
```
jan-kiszka, line 286: Would be nicer to make this a superclass of UpdateFirmware.
Maybe `FirmwareUpdate()` vs. `ManagedFirmwareUpdate(FimwareUpdate)`.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
jan-kiszka: It's a trade-off, true: Theoretically, a firmware version we release tomorrow could become incompatible with an image version we release later on. But, unless someone breaks the SYSFW ABI again, this may be rather unlikely from now on.

If we see a remaining risk, we would indeed have to define an upper limit of supported image versions already today, say `< V02.*`. But what if we feel like bumping the image version to a new major level for other reasons, e.g. due to a distro update? Then you also logically break compatibility with existing firmware update packages.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
BaochengSu: Done, please review again.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
BaochengSu: Done, please review again.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
```diff
@@ -137,7 +137,9 @@ class FirmwareUpdate(object):
             mtd_base_path = "{}/mtd".format(ospi_dev_path)
 
         self.firmware.seek(0)
-        firmware_size = os.stat(self.firmware.fileno()).st_size
+        self.firmware.seek(0, os.SEEK_END)
+        firmware_size = self.firmware.tell() 
+        self.firmware.seek(0)
```
jan-kiszka, line 142: As this is a regression fix for the previous commit, it should get folded into it.

jan-kiszka:
LGTM (except for the last patch split-up)

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
BaochengSu: Squeshed.

PR #163, opened 2021-08-27
customizations-example: Resolve renaming race when swapping eth0/eth1

jan-kiszka:
I think this should go in before the release, to avoid renaming twice.


AsuraZeng (approved):
the eth naming rule follow [this](https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames/). I think it's ok.

PR #164, opened 2021-08-31
iot2050setup: fix the issue that it would hang when using dp login

AsuraZeng:
as only take ssh login and serial login into consider, missing the
dp condition.

so only when login from serial, it would resize the window.
it would ignore other option

### DIFF
--- a/recipes-app/iot2050setup/files/iot2050setup.py
+++ b/recipes-app/iot2050setup/files/iot2050setup.py
@@ -643,13 +643,11 @@ class TerminalResize:
 
 
 def main():
-    default_console_login = True
-    # get the ssh releated env
-    ssh_client = os.getenv('SSH_CLIENT')
-    ssh_tty = os.getenv('SSH_TTY')
-
-    if ssh_client is not None or ssh_tty is not None:
-        default_console_login = False
+    default_console_login = False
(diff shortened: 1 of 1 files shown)


recipes-app/iot2050setup/files/iot2050setup.py
```diff
@@ -643,13 +643,12 @@ def resize(cls):
...
-    # get the ssh releated env
-    ssh_client = os.getenv('SSH_CLIENT')
-    ssh_tty = os.getenv('SSH_TTY')
+    default_console_login = False
+    # get the term releated env
+    tty_client = os.getenv('TERM')
 
-    if ssh_client is not None or ssh_tty is not None:
-        default_console_login = False
+    if tty_client == "vt220":
```
jan-kiszka, line 650: Is that type sufficient?

jan-kiszka:
```
root@iot2050-debian:~# tty
/dev/ttyS3
```

That works by reading where `/proc/self/fd/0` points to - anything /dev/ttyS or ttyUSB would indicate UART without having to probe all terminal types.

jan-kiszka (approved):

PR #169, opened 2021-09-03
Ignore vscode ide folder

BaochengSu:


jan-kiszka (commented):
Generally, this is better handled in `$HOME/.config/git/ignore`. Such lists can become rather long...

PR #170, opened 2021-09-03
Freeze bullseye code base

BaochengSu:
Freeze at Sep. 2 2021.


jan-kiszka (approved):
Looks good.

BTW, there is now also https://debian.notset.fr/snapshot/archive/, apparently more reactive and not throttling. Unfortunately, only with AMD64 packages...

BaochengSu:
> Looks good.
> 
> BTW, there is now also https://debian.notset.fr/snapshot/archive/, apparently more reactive and not throttling. Unfortunately, only with AMD64 packages...

visible ward

PR #172, opened 2021-09-04
update package versions

BaochengSu:
Update below packages:

1. tcf-agent
2. node-red
3. node-red nodes
4. Install python-is-python3


jan-kiszka (approved):
LGTM

PR #173, opened 2021-09-04
Update tiboot3.bin and sysfw

BaochengSu:
Update tiboot3.bin to D01.01.00.01

Update sysfw to 2021.05


BaochengSu:
This one should be merged after SR2 kernel ready

jan-kiszka:
Why only then? Does it break anything for the current kernel?

jan-kiszka:
And please also already queue pru firmware updates / extensions. Something like https://github.com/siemens/meta-iot2050/commit/836dd17108070b87a0fbfacccfb7ae10814c6fd6 (plus the links in recipes-bsp/u-boot/files/prebuild/).

BaochengSu:
> Why only then? Does it break anything for the current kernel?

Sorry, that's a mistake.

BaochengSu:
> And please also already queue pru firmware updates / extensions. Something like [836dd17](https://github.com/siemens/meta-iot2050/commit/836dd17108070b87a0fbfacccfb7ae10814c6fd6) (plus the links in recipes-bsp/u-boot/files/prebuild/).

Done.

jan-kiszka:
Too late for this round, but maybe we can rather pick the binary from https://git.ti.com/cgit/processor-firmware/ti-linux-firmware/tree/ti-pruss?h=ti-linux-firmware. Theoretically, the same versions should be available there as well.

PR #176, opened 2021-09-06
Firmware fixes and cleanups

jan-kiszka:
Resolves an issue after the bullseye update and drops self-hosted pruss firmware images.


BaochengSu (approved):

AsuraZeng (approved):
LGTM

PR #179, opened 2021-09-09
Su/hw reset cp210x

BaochengSu:
Add the hardware reset to the CP210X, origined from Chao.

### DIFF
--- a/recipes-app/iot2050setup/files/iot2050setup.py
+++ b/recipes-app/iot2050setup/files/iot2050setup.py
@@ -1,5 +1,9 @@
 #! /usr/bin/python3
-
+#
+# Copyright (c) Siemens AG, 2021
+#
+# This file is subject to the terms and conditions of the MIT License.  See
+# COPYING.MIT file in the top-level directory.
 import traceback
 from snack import *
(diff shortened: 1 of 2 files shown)


jan-kiszka:
Does this address the rs485 issue you mentioned? Or which scenario does it resolve?

AsuraZeng (approved):

AsuraZeng:
@jan-kiszka  it could no need to reboot the device  making  change the com port function among 232 485 422 take effect for PG2 board

AsuraZeng:
@BaochengSu  lack of  supporting for the `CP2102N-RESET` pin.
pls refer https://github.com/siemens/meta-iot2050/commit/20d582b03da29fa1e11429c9d18465cfe2a9768a
`main_gpio1` dts node to pick the `CP2102N-RESET` related setting

BaochengSu:
> Does this address the rs485 issue you mentioned? Or which scenario does it resolve?

RS485 issue is another topic. This one is regarding of the CP210x hardware design change for PG2.

Under the new design, it's not necessory to reboot the board to let the serial switching taking effect.

PR #181, opened 2021-09-09
Proposal for internationalization

SteffenBurkhardt:
For this approach choices must also have an identifier.


jan-kiszka:
Let's discuss this in kas context (mailing list) - meta-iot2050 is just one potential user.

PR #183, opened 2021-09-13
u-boot:sf:get the permission before operating the device

AsuraZeng:
When flash protect is enable,write or erase the flash would not reminder any error.
It would show "erase Ok",but actually erasing the flash is not success

Check the write protection status before operating the device


AsuraZeng:
no respond from u-boot.
this feature I think could keep at local firstly.

BaochengSu:
Keep it locally at this moment, due to need delivery to testing.

BaochengSu (approved):

jan-kiszka:
This has style issues (checkpatch is your friend).

What's the link to the upstream discussion?

PR #184, opened 2021-09-13
Chao/change boot order

AsuraZeng:


BaochengSu (approved):

jan-kiszka:
NACK (for the future): This is no separate patch, reasoning is wrong (you do NOT want to start from emmc), format is wrong. We are maintaining an upstream queue, so this needs to be folded in to stay in sync.

jan-kiszka:
In fact, this needs to be reverted: We have the button as control over booting externally. I will send a PR.

PR #185, opened 2021-09-13
Revert "u-boot:change the boot order"

jan-kiszka:
This reverts commit a5a1717df9104338c0c63eabdbc723485701b1e6.

We now have the user button to control the boot order, this no longer
the need to prioritize external media. This is safer when the user
accidentally plugs a bootable device.


AsuraZeng:
this should be integrated by industry os.
Pre-installed industry os in the emmc, so  it can not change the boot order through button.

jan-kiszka:
I can't follow your argument yet. Whatever is installed on the eMMC, you can always press and hold the user button to skip it from being booted. That is true since we updated the firmware. Thus, there is no need to flipping boot orders permanently.

BaochengSu:
> I can't follow your argument yet. Whatever is installed on the eMMC, you can always press and hold the user button to skip it from being booted. That is true since we updated the firmware. Thus, there is no need to flipping boot orders permanently.

I think chao's idea is that the `install-on-emmc` feature should exist in the Industrial OS as well, otherwise user cannot use the `install-on-emmc` feature to update the emmc.

jan-kiszka:
No, it is unrelated HOW you install from the external boot medium the eMMC. But that's in fact another reason to merge this: The changed boot order breaks the expected workflow of our installation script.

I would recommend giving this a try first.

jan-kiszka:
Just added a cleanup of the write-protect patch as well, the upstream submitted version.

BaochengSu:
Just checked that we already has the `remove-emmc-target` in u-boot, so this patch is not needed now.

BaochengSu (approved):

jan-kiszka:
More precisely, we have the `CONFIG_IOT2050_BOOT_SWITCH` feature now ("Disable eMMC boot via USER button") after updating U-Boot to the upstream queue, yes.

AsuraZeng:
do there a mechanism to restore the emmc back in boot_targets after installing emmc in order to boot from emmc next bootup

jan-kiszka:
Just read the code: https://github.com/siemens/u-boot/blob/jan/iot2050-v2021.04/board/siemens/iot2050/board.c#L66 - this is temporary, affecting the current boot only.

PR #186, opened 2021-09-13
u-boot: fix the string compare for the board name

BaochengSu:
Use strncmp to compare the board name to check if the board is
advanced or not.


jan-kiszka:
Looks good - just let me push the series here so that we avoid unrelated git footer changes (back and forth).

jan-kiszka:
I would also use strstr - to avoid the manually counted 16 chars.

jan-kiszka (approved):

PR #190, opened 2021-09-17
linux-iot2050-rt: Use stable patch URL

jan-kiszka:
RT patches are only available under "older" when newer releases appear.


AsuraZeng:
the patches are in the link always change, I suggest we use the patches in older folder

PR #191, opened 2021-09-17
Align kernel versions

jan-kiszka:
This stable update would allow aligning both kernels over one version.


jan-kiszka:
Outdated by now.

PR #193, opened 2021-09-24
iot2050setup: fix the serial port switch function

AsuraZeng:
1.The result of the regular expression match contains the null character,
which causes the subsequent comparison to fail

2.Wait for the sync to complete,then reset the bridge

### DIFF
--- a/recipes-app/iot2050setup/files/iot2050setup.py
+++ b/recipes-app/iot2050setup/files/iot2050setup.py
@@ -26,7 +26,7 @@ class ansicolors:
 class TopMenu:
     def __init__(self):
         self.gscreen = SnackScreen()
-        self.boardType = subprocess.check_output('grep -a -o -P "IOT2050 .*" /proc/device-tree/model',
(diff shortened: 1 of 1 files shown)


recipes-app/iot2050setup/files/iot2050setup.py
```diff
@@ -26,7 +26,7 @@ class ansicolors:
 class TopMenu:
     def __init__(self):
         self.gscreen = SnackScreen()
-        self.boardType = subprocess.check_output('grep -a -o -P "IOT2050 .*" /proc/device-tree/model',
+        self.boardType = subprocess.check_output('grep -a -o -P "IOT2050 [\w\s]+" /proc/device-tree/model',
```
BaochengSu, line 29: The \s also covers the preceding space as well I think.

AsuraZeng: yes, the space could be deleted

BaochengSu:
LGTM

BaochengSu (approved):

PR #194, opened 2021-09-26
linux:dts:re-correct usb dts configurations

AsuraZeng:
There is a problem with the configuration of the usb dts, which needs to be corrected

Signed-off-by: chao zeng <chao.zeng@siemens.com>

PR #195, opened 2021-09-26
linux-iot2050: Update patch queue to fix USB DT entry

jan-kiszka:
usb0_phy must have been usb0, the host controller. This truly enables
USB 3.0 mode now.

The update also brings a naming fix for cp2102n_reset_pin_default from
upstream.

Obsoletes #194.


BaochengSu (approved):

BaochengSu:
@jan-kiszka The `From hash` makes it hard to locate the real change in PRs, do we really need it or we can actually remove that line without impacting the isar building?

Same question applies to the git version number.

For example, this debian patch doesnot contain any of them: https://salsa.debian.org/debian/openssl/-/blob/debian/unstable/debian/patches/Set-systemwide-default-settings-for-libssl-users.patch

PR #197, opened 2021-09-27
Update tiboot3.bin to D01.01.00.04

BaochengSu:


AsuraZeng:
we will tag the tiboot3.bin using V version  not use the D version eventually?

BaochengSu:
> we will tag the tiboot3.bin using V version not use the D version eventually?

Yep, but now we are not that close to the V tag.

AsuraZeng:
could we leave the commit message to describe what's the detail change about this PR?

AsuraZeng (approved):

PR #198, opened 2021-09-30
Fix the omap RS485 issue

BaochengSu:
An issue is observed on 8250 tty that under the RS485 or RS422 mode, the
RX/TX does not work. For RS485 the RX does not work meanwhile for RS422
the TX does not work.

This issue was introduced in the kernel driver when Lukas Wunner added
commit f45709df7731 to prevent the userspace interfering w/ RS485.
However this commit breaks the RS485.

Meanwhile there are some mis-configuration in the switchserialmode tool,
which may covered some potential problems in the past.

### DIFF
--- a/recipes-app/switchserialmode/files/src/switchserialmode.c
+++ b/recipes-app/switchserialmode/files/src/switchserialmode.c
@@ -349,15 +349,17 @@ static E_STATUS ttyuart_switchto_rs485(const char *uartdev, int logicLevel)
 
     SET_UART_RTS_ACTIVE_LOGIC(rs485conf.flags, logicLevel);
 
-    rs485conf.flags |= SER_RS485_ENABLED | SER_RS485_RX_DURING_TX;
+    rs485conf.flags |= SER_RS485_ENABLED;
 
     return ttyuart_set_rs485conf(uartdev, &rs485conf);
 }
 
-static E_STATUS ttyuart_switchto_rs422(const char *uartdev)
+static E_STATUS ttyuart_switchto_rs422(const char *uartdev, int logicLevel)
 {
     struct serial_rs485 rs485conf;
     memset(&rs485conf, 0, sizeof(rs485conf));
+
+    SET_UART_RTS_ACTIVE_LOGIC(rs485conf.flags, logicLevel);
 
     rs485conf.flags |= SER_RS485_ENABLED | SER_RS485_RX_DURING_TX;
 
@@ -412,7 +414,7 @@ static void ttyuart_command_handle(int argc, char **argv)
     }
     else if(0 == strcasecmp(mode, "rs422"))
     {
-        ttyuart_switchto_rs422(devType);
+        ttyuart_switchto_rs422(devType, logicLevel);
     }
     else
     {
--- a/recipes-kernel/linux/files/patches-5.10/0117-Revert-serial-8250-Don-t-touch-RTS-modem-control-whi.patch
+++ b/recipes-kernel/linux/files/patches-5.10/0117-Revert-serial-8250-Don-t-touch-RTS-modem-control-whi.patch
@@ -1,0 +1,49 @@
+From df91a5f865fab96ed5c0c3f0e2b941d629bc845a Mon Sep 17 00:00:00 2001
+From: Su Bao Cheng <baocheng.su@siemens.com>
+Date: Mon, 22 Nov 2021 15:59:01 +0800
+Subject: [PATCH 117/117] Revert "serial: 8250: Don't touch RTS modem control
+ while in rs485 mode"
+
+This reverts commit f45709df7731ad36306a28a3e1af7309d55c35f5.
+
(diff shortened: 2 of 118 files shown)


recipes-kernel/linux/files/patches-5.10/0105-Revert-serial-8250-Don-t-touch-RTS-modem-control-whi.patch
```diff
@@ -0,0 +1,40 @@
...
+ rs485 mode"
+
+This reverts commit f45709df7731ad36306a28a3e1af7309d55c35f5.
+
+The `serial8250_do_set_mctrl` not only used by userspace but also the
+kernel itself.
+
+During tty_open, the omap8250_restore_regs first manipulates some
+registers then use the set_mctrl to restore the MCR register.
+
```
jan-kiszka, line 14: Reported upstream already?

BaochengSu:
I've tried the patch provided by Lukas, seems issue is resolved. However, further testing has been arranged. Let's see.

BaochengSu:
What's left to haunting me is that the RS485 config flags are definitely wrong in released V02 or V01.01.01 version, however the actual RS485 and RS422 behave correctly...

I suspect the hardware tranceveiver covers it, need dig more.

jan-kiszka:
You mean that `switchserialmode ... -d` is not returning what was configured?

If the patch works, please provide feedback on the mailing list. But I'm a bit reluctant to merging this "fix" because the new part in it needs to go through review first. I would just update the description of your revert commit for now.

BaochengSu:
> You mean that `switchserialmode ... -d` is not returning what was configured?
> 
Yep, on old V02 or triple 01 release.

> If the patch works, please provide feedback on the mailing list. But I'm a bit reluctant to merging this "fix" because the new part in it needs to go through review first. I would just update the description of your revert commit for now.
Definitely I will provide the feedback, that is also what I planned to do after the further testing.

Meanwhile, do you suggest for this PR we only need to rever the commit?

jan-kiszka:
The revert is the safest option for now, yes. We can update it once an official patch hit mainline.

BaochengSu:
@jan-kiszka reverted to the old reverting version.

recipes-app/switchserialmode/files/src/switchserialmode.c
```diff
@@ -412,7 +414,7 @@ static void ttyuart_command_handle(int argc, char **argv)
     }
     else if(0 == strcasecmp(mode, "rs422"))
     {
-        ttyuart_switchto_rs422(devType);
+        ttyuart_switchto_rs422(devType, logicLevel);
```
jan-kiszka, line 417: This file was probably committed by accident.

recipes-kernel/linux/files/patches-5.10/0117-Revert-serial-8250-Don-t-touch-RTS-modem-control-whi.patch
```diff
@@ -0,0 +1,45 @@
...
+Subject: [PATCH 117/117] Revert "serial: 8250: Don't touch RTS modem control
+ while in rs485 mode"
+
+This reverts commit f45709df7731ad36306a28a3e1af7309d55c35f5.
+
+The mcr register is not always reflect the significant value, so
+checking against it will mislead.
+
+The direct issue is RS485/422 does not fully work on 8250 serial, due to
+the MCR[1] is not set correctly with the to-be-converted commit.
```
jan-kiszka, line 13: The correct reasoning is that the original patch was trying to avoid user-space triggered changes but incorrectly blocked also valid kernel-originated changes because it hooked into the wrong function.

recipes-app/switchserialmode/files/src/switchserialmode.c
BaochengSu: No, it's on purpose. The `SER_RS485_RX_DURING_TX` is used to tell apart the RS422 from RS485. Also the other change is needed to let the RS422 fully functioning.

Actually this file is far away from what it should be, I was working on a refactory, however, at this moment maybe just to fix the issue is ok.

recipes-kernel/linux/files/patches-5.10/0117-Revert-serial-8250-Don-t-touch-RTS-modem-control-whi.patch
BaochengSu: ACK and revised.

recipes-app/switchserialmode/files/src/switchserialmode.c
jan-kiszka: It's not ok to merge different topics into the same commit.

BaochengSu: ACK w/ revision, please review

jan-kiszka (approved):

PR #205, opened 2021-10-22
kas update / kas menu support

jan-kiszka:
This should be another usability jump. It specifically helps to make image & build options easier discoverable.


Kconfig
```diff
@@ -0,0 +1,99 @@
...
+	  interface.
+
+config IMAGE_BOOT
+	bool "Firmware image"
+	help
+	  Build the firmware image that is responsible for booting the device.
+
+	  WARNING: Do not flash this image onto your device unless you know
+	  that it fits AND you have an external flash programmer at hand that
+	  allows to recover. Otherwise you risk to BRICK THE IOT2050!
```
BaochengSu, line 30: Seems this should be adapted to the pg1&pg2 split.

jan-kiszka: right, will have a look

Kconfig
```diff
@@ -0,0 +1,99 @@
...
+comment "Build options"
+
+config SDK
+	bool "Build SDK"
+	help
+	  Generate an SDK consisting of a cross-compiler and library headers
+	  needed to build applications for the device.
+
+	  Note: This will not generate the image for the device itself.
+
```
BaochengSu, line 65: Could this be a `Image Type` rather than a `Build opt`?

For example, the below combinations should make no difference on the built output:
- `Example Image` + `Build SDK`
- `Example Image` +  `Preempt-RT` + `Build SDK`
- `LXDE Image` + `Build SDK`
- `LXDE Image` +  `Preempt-RT` + `Build SDK`

If I am not fimilar w/ the code, then I would be confused.

jan-kiszka: Practically, the SDK is the same for example vs. LXDE. Preempt-RT, though, should install a different kernel header package. If not, we had a bug.

BaochengSu: OK, this means that for the whole picture, we actually should build 3 binaries for SDKs, or at least 2 since LXDE is not in our test plan...

README.md
```diff
@@ -13,25 +13,13 @@ For example under Debian Linux
 sudo apt install docker.io
 ```
 
-Then build the example image
+Then open the menu to select the desired image and options:
 
 ```shell
-./kas-container build kas-iot2050-example.yml
+./kas-container menu
```
BaochengSu, line 19: Doesn't work for my tmux configuration:

```shell
❯ ./kas-container menu
2022-01-07 08:18:42 - INFO     - kas 2.6 started
Unknown terminal: tmux-256color
Check the TERM environment variable.
Also make sure that the terminal is defined in the terminfo database.
Alternatively, set the TERMCAP environment variable to the desired
termcap entry.
```

jan-kiszka: What's your `TERM` on the console side?

jan-kiszka: Probably a kas-container topic: We may only propagate supported `TERM` values, similar to `SHELL`. Use a proper shell env, and it will work.

BaochengSu: ❯ echo $TERM
tmux-256color

jan-kiszka: And that is not ncurses-compatible. Check `/lib/terminfo` for supported settings. Not sure if Debian has an extension for tmux as well somewhere, maybe not.

jan-kiszka: Buster vs. bullseye topic: ncurses from the latter addresses this. Will be fixed by kas 3.0 (once released) and by us picking up that version eventually.

BaochengSu: > Probably a kas-container topic: We may only propagate supported `TERM` values, similar to `SHELL`. Use a proper shell env, and it will work.

Yep, another zsh shell works well.

Kconfig
jan-kiszka: ...or make the SDK build both kernels and include both kernel header packages. Our current SDK is corresponding to the example image, and that uses the regular kernel.

BaochengSu: That would be better. 

BTW, could we install two kernels in one example image? Does u-boot support multiple kernel such as grub2?

jan-kiszka: Nope, this is not supported u-boot-script. Would take an interactive boot loader - but that makes no sense for an embedded device.

You can always switch the kernel by building and installing the different package, though.

jan-kiszka:
Just updated the queue. Note that I changed my mind regarding how to model the boot image builds in kas best: top-level files are better because they are more explicit.

BaochengSu:
Probably off-topic, how is the download speed of `ghcr.io` from your side?

BaochengSu:
> 

makes sence to me.

jan-kiszka:
> Probably off-topic, how is the download speed of `ghcr.io` from your side?

Multiple MB/s. Another "Google is suspicious" issue behind the Great Firewall? Is there some alternative public container registry closer to you that we could fill as well? Then users would only have to set `KAS_CONTAINER_IMAGE_PATH`.

BaochengSu:
> > Probably off-topic, how is the download speed of `ghcr.io` from your side?
> 
> Multiple MB/s. Another "Google is suspicious" issue behind the Great Firewall? Is there some alternative public container registry closer to you that we could fill as well? Then users would only have to set `KAS_CONTAINER_IMAGE_PATH`.

Yep, I am not sure if something nearby. However, I don't think this is a serious issue for most developers here :).

BaochengSu (approved):

PR #207, opened 2021-10-28
linux: usb : serial: pick up some missing patches to fix software flo…

AsuraZeng:
…wcontrol

Since some patches are missing,resulting usb-to-serial software flowcontrol can not work

Pick up releated patches to fix the software flowcontrol


AsuraZeng:
> I think we are missing 2d9a00705910ccea2dc5d9cba5469ff2de72fc87. Otherwise, this looks good.

hi jan
yes，you are right. I have  picked this commit.

BaochengSu (approved):

PR #211, opened 2021-11-04
mraa: uart: allow change the uart attribute immediately

AsuraZeng:
When using the flow control, the rts asserts,
There is also data in the buffer on the sender,
So it can't change the termios attribute when
re-instantiate the device.

When the device is re-instantiated, discard
the input and output buffers

### DIFF
--- a/recipes-app/mraa/files/0001-aio.c-fix-mraa_aio_set_bit-for-result-scaling.patch
+++ b/recipes-app/mraa/files/0001-aio.c-fix-mraa_aio_set_bit-for-result-scaling.patch
@@ -1,7 +1,7 @@
-From 3b1222d0fc05107020e4f1e00e1d45b782c38c95 Mon Sep 17 00:00:00 2001
+From 95f32e87ca29fadeac35bc7f8925ba151d9eb770 Mon Sep 17 00:00:00 2001
 From: "le.jin" <le.jin@siemens.com>
 Date: Fri, 1 Nov 2019 18:02:54 +0800
-Subject: [PATCH 01/13] aio.c: fix mraa_aio_set_bit for result scaling
+Subject: [PATCH 01/14] aio.c: fix mraa_aio_set_bit for result scaling
 
 Signed-off-by: le.jin <le.jin@siemens.com>
 ---
@@ -28,5 +28,5 @@ index 0cc631f..7a94ac1 100644
  }
  
 -- 
-2.26.2
+2.32.0
 
--- a/recipes-app/mraa/files/0002-feat-iot2050-add-iot2050-platform-support.patch
+++ b/recipes-app/mraa/files/0002-feat-iot2050-add-iot2050-platform-support.patch
@@ -1,7 +1,7 @@
-From a1bee1d7bd044a7658e54d894b5ffbfb33f58360 Mon Sep 17 00:00:00 2001
+From 04d1329e410982dcf45bc71c12feaecac09b0472 Mon Sep 17 00:00:00 2001
 From: "le.jin" <le.jin@siemens.com>
 Date: Fri, 24 May 2019 14:33:42 +0800
-Subject: [PATCH 02/13] feat:iot2050:add iot2050 platform support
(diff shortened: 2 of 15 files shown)


jan-kiszka:
This is also an upstream topic. Unfortunately, I do not fully understand the patch description yet to assess whether this change can be applied unconditionally. Can you try to describe the failure scenario more verbosely? How can a user reproduce it now?

AsuraZeng:
First of all, needing to enable hardware flow control.
the receiving end data is full, assert the rts signal,the sender(arduino uart) stop send the data,but currently sender buffer is not empty.
if re-instantiated the serial port can not configured attribute successfully.

so empty the input and output buffer when re-instantiated serial port

jan-kiszka:
Shouldn't the kernel flush the buffer when closing a port? If not, shouldn't we (also) do that when we close the port?

Second, why the change TCSAFLUSH -> TCSANOW?

AsuraZeng:
there is no chance for user to close the port. like CTRL+C to exit the port. 

TCSAFLUSH is meaning waiting the buffer to clear. 
and if We clear the buffer before this operation, I think use the TCSANOW is more reasonable not to wait
the buffer clear

jan-kiszka:
`mraa_uart_stop()`? Your description is still missing the pieces that provide a full reproduction picture.

And the change for TSCANOW should also be explained in the patch commit message.

AsuraZeng:
I think I need to describe how to reproduce more detail.
1.enable arduino uart flow-control.
2. send data from ttyS1 device and using another device to receive data.
SEND SIDE
```
import serial
import time

ser=serial.Serial('/dev/ttyS1',9600, rtscts=True)  

nsendlen=0
nread=0
while 1:
    string_send="f"
    nsendlen+=len(string_send)
    ser.write(string_send.encode())
    print(nsendlen)
    time.sleep(0.01)
    print("cts:{}".format(ser.cts))
```

RECEIVE SIDE
```
import serial
import time

ser=serial.Serial('/dev/ttyUSB0',9600, rtscts=True)  

nsendlen=0
nread=0

while 1:
    print("ongoing...")
    time.sleep(0.5)
```

3. when receive buffer is almost full, the RTS would assert
 and send side could not send data
4. as the send side send data dead loop,exit the progress using CTRL+C
5. Now,using continue to re-instantiate the ttyS1 using mraa interface.
```
import mraa
mraa.Uart(0)
```
it can not instance  successfully. it would stall at here



# TCSANOW and TCSAFLUSH
why I change the command from TCSAFLUSH to TCSANOW.
as has cleared before using tcflush(dev->fd,TCIOFLUSH)

at here I think  there is no necessary to use TCSAFLUSH waiting the send buffer clear and discard the input buffer.

jan-kiszka:
So, that would mean a close() on the sender side with outgoing data in the queue, blocked by flow-control, will not flush the buffer. Therefore, the next user of the UART will find the buffer busy, and the tcsetattr() may not progress.

But I wonder if unconditional flushing is then the right thing. Can't the situation be cleanly resolved by the receiver finally accepting the data? Or are we in a dead-lock scenario now?

AsuraZeng:
> So, that would mean a close() on the sender side with outgoing data in the queue, blocked by flow-control, will not flush the buffer. Therefore, the next user of the UART will find the buffer busy, and the tcsetattr() may not progress.
> 
> But I wonder if unconditional flushing is then the right thing. Can't the situation be cleanly resolved by the receiver finally accepting the data? Or are we in a dead-lock scenario now?

Yes，you are right.
the sender continue send the data and the buffer should always contain the data.
re-open the port would failed.

I don't know if there are scenarios where data needs to be retained when re-instantiate the port
But I think the re-opening of the serial port  ,the buffer should be discarded

jan-kiszka:
I'm picking on this as this is an interface change. And if there is not issue resolving this situation by simply receiving the blocked data, I'm inclined to not change the interface.

AsuraZeng:
This can be processed from user program anyway.
modify the mraa interface is not properly.

so close the PR

PR #213, opened 2021-11-11
linux-iot2050:fix: couldn't connect to phy

AsuraZeng:
When booting up,it would show the error "couldn't connect to phy".
as it trigger the deferred probe mechanism,it would probe later

But there is a chance of error, causing the MAC to fail to link with
the phy normally.Resulting in the ethernet can not work

How to reproduce:
1.power on and get the ip then ping router
2.reboot
3.waiting bootup,and repeat 1~3

Repeat about 500 times,it would cause the error.
Compile the mdio into kernel to fix this issue.


BaochengSu:
LGTM at this moment.

Long term speaking, the defered probe should not trigger kernel oops even if the mdio is modularized.

jan-kiszka:
This makes no sense to me yet: We didn't build that MDIO driver, but Ethernet generally still worked?! Also, this driver seems to be involved only in the CPSW interface, not the PRU one. Otherwise, the prueth driver would do a `select TI_DAVINCI_MDIO`.

I strongly suspect this is papering over the actual problem.

BaochengSu:
> This makes no sense to me yet: We didn't build that MDIO driver, but Ethernet generally still worked?! Also, this driver seems to be involved only in the CPSW interface, not the PRU one. Otherwise, the prueth driver would do a `select TI_DAVINCI_MDIO`.
> 
> I strongly suspect this is papering over the actual problem.

looks actually Prueth is using this DAVINCI_MDIO: https://github.com/siemens/meta-iot2050/blob/65a3ba902b3ada8a34acba9331abf839d6073ca9/recipes-kernel/linux/files/patches-5.10/0085-net-ti-icssg-prueth-Add-ICSSG-ethernet-driver.patch#L119

jan-kiszka:
OK, now it now starts to make sense. Hope TI is fixing this before upstreaming that code....

BaochengSu:
Yep, we will inform TI regarding this issue.

BaochengSu (approved):

PR #214, opened 2021-11-16
iot2050-firmware-update: Use device tree board name

BaochengSu:
Use device tree board name to distinguish board type, instead of using
the soc id, due to the later is proven to be not a solid method.


jan-kiszka (approved):
Theoretically less safe (you can boot some devices with the wrong DT), but the practical reasons are also understood - looks good.

PR #216, opened 2021-11-17
u-boot-iot2050: Make dependencies compatible with downstream machine

gylstorffq:
If downstream creates a new machine based on conf/machine/iot2050.conf
the dependency to optee-os-iot2050 is no longer fulfilled. Rename
optee-os-iot2050 to optee-os-${Machine} to allow downstream defined
machines.

Signed-off-by: Quirin Gylstorff <quirin.gylstorff@siemens.com>

### DIFF
--- a/recipes-bsp/u-boot/u-boot-iot2050.inc
+++ b/recipes-bsp/u-boot/u-boot-iot2050.inc
@@ -26,9 +26,9 @@ SPI_FLASH_DEPLOY_IMG = "iot2050-image-boot.bin"
 DEPENDS += "swig"
 
 # Build environment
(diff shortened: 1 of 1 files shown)


jan-kiszka:
Nope, optee-os-iot2050 remains optee-os-iot2050, not optee-os-<random-machine>. Where do we refer to optee-os-${MACHINE}? Do we need to set `OPTEE_NAME`?

gylstorffq:
Close MR as OPTEE_NAME solves the problem.

jan-kiszka:
Will you send an alternative PR then?

PR #218, opened 2021-11-18
install-on-emmc: Add a case where the button has been pressed

AsuraZeng:
As we can press the button to remove the emmc from boot_targets,
besides they'd like to install image to emmc. It is hard for user
to press the button,release the button then again pressing the button
to trigger the install process.

The easiest way is to keep pressing the button without thinking
So add one condition to detect the key has alreay been pressed always.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,6 +127,9 @@ The installation can also be triggered automatically by creating the file
 `/etc/install-on-emmc` on the vanilla image by mounting it under Linux and
 executing, e.g., `touch <mountpoint>/etc/install-on-emmc`.
 
+The installation can also be triggered by keeping pressing the user button 
+during startup at the first boot.
+
 ## Selecting a boot device
 
 By default, the boot loader will pick the first bootable device. If that device
--- a/recipes-core/install-on-emmc/files/install-on-emmc.sh
+++ b/recipes-core/install-on-emmc/files/install-on-emmc.sh
@@ -84,12 +84,14 @@ if [ "${EMMC_DEV}" = "${BOOT_DEV}" ]; then
 fi
 
 # Presence of /etc/install-on-emmc skips button check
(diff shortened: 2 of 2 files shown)


jan-kiszka:
This is wrong, and the current behavior has been designed as it is for a reason: The user holds the button during boot-up in order to select a different boot device. When that is the case, the button must be released before boot continues, and the eMMC installation would start to evaluate the button. So, no problem in that case.

For the eMMC installation procedure itself: Once the user activated it via holding the button long enough while the orange LED was flashing, it can be released again and need not be touched again after that because boot defaults (eMMC first) already do the right thing. Thus, there is no issue that requires any adjustments.

AsuraZeng:
1.the key need to be release is not must when selecting different boot targets.
almost 5s it would continue to boot-up.
2.  Observing the state of the light increases the complexity of the operation

I think it’s necessary to provide an easy-to-operate way

jan-kiszka:
Right, there is a timeout to avoid that a stuck button would make the device unbootable. It's 10 seconds (https://github.com/u-boot/u-boot/blob/master/board/siemens/iot2050/board.c#L206).

The button sequence is a key element in the procedure because we must avoid accidental overwriting of the eMMC by users. Therefore: strong NACK.

BaochengSu:
> Right, there is a timeout to avoid that a stuck button would make the device unbootable. It's 10 seconds (https://github.com/u-boot/u-boot/blob/master/board/siemens/iot2050/board.c#L206).
> 
do we have this u-boot user-button feature documented somewhere? Otherwise I will submit a patch to record it somewhere.
> The button sequence is a key element in the procedure because we must avoid accidental overwriting of the eMMC by users. Therefore: strong NACK.

jan-kiszka:
Nope, this is not documented yet. README might be a good starting point.

Same for install-on-eMMC, in fact.

PR #223, opened 2021-11-26
HACK: tty: Flush any pending characters in the driver

AsuraZeng:
omap_8250 can't send the data when closing the port,it keeps the buffer.

then re-used the port, at least there are several bugs.
1. the receive side could receive these data again.
2. it would hang when change the termios.

Flush the buffer when send timeout or some errors.

### DIFF
--- a/recipes-kernel/linux/files/patches-5.10/0001-dmaengine-ti-k3-udma-glue-Add-function-to-get-device.patch
+++ b/recipes-kernel/linux/files/patches-5.10/0001-dmaengine-ti-k3-udma-glue-Add-function-to-get-device.patch
@@ -1,7 +1,7 @@
-From 82c840791fdb200ce51f4213f8ba015bc3bdaf2a Mon Sep 17 00:00:00 2001
+From 83202d53880f68384482db221d5b30327057459c Mon Sep 17 00:00:00 2001
 From: Peter Ujfalusi <peter.ujfalusi@ti.com>
 Date: Tue, 8 Dec 2020 11:04:24 +0200
-Subject: [PATCH 001/117] dmaengine: ti: k3-udma-glue: Add function to get
+Subject: [PATCH 001/118] dmaengine: ti: k3-udma-glue: Add function to get
  device pointer for DMA API
 
 Glue layer users should use the device of the DMA for DMA mapping and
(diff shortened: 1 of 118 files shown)


AsuraZeng:
I think this should be upstream too. When the sending fails, the output buffer needs empty, to enhance the flow of robust

jan-kiszka:
So, #211 was basically a symptom of what this tries to resolve.

Questions:
- Is this behavior different to an older kernel version? Or our old BSP kernel?
- Is this really generic (you patch generic code)? Or rather omap-specific as your comment above suggests?

AsuraZeng:
yes , #211  is just one of the symptoms. 

Questions:

    Is this behavior different to an older kernel version? Or our old BSP kernel?
     yes ,it is different with older kernel version. 
    Is this really generic (you patch generic code)? Or rather omap-specific as your comment above suggests?
    yes,this is generic code,not the omap specific code. I observed several other uart devices and found that the final buffers were all emptied. maybe this is a common behavior,buffer will be emptied anyway.When sending timeout or abnormal then empty the buffer.

jan-kiszka:
Can you specify the kernel versions that used to work?

And, no, the problem is NOT generic when it only affects the OMAP UART, and all others do that clearing anyway. In that case, the patch must target the driver, not the tty layer.

AsuraZeng:
Currently , I found kernel 4.19.94 works fine.

Yes,actually this should target to omap driver. For tty layer is just a workaround.

jan-kiszka:
Not a bug of the kernel, wrong test pattern: close() as well as TCSETSW (what stty uses) shall block until pending transmissions are flushed. If those can't make progress due to remote side not accepting them, the application must flush the buffers first - or use TCSETS.

AsuraZeng:
if so, #211   needs to re-open and patch to our currently  mraa.

jan-kiszka:
Yes, we need a patch to mraa, but not like #211. And not only there.

We need a new function in mraa in order to drop the output buffer, and that needs to be called explicitly from the application. As `mraa_uart_flush()` is already defined for draining output, a new `mraa_uart_clear()` could serve for dropping queues, maybe with a parameter to select input/output/both.

jan-kiszka:
And mraa API changes needs to be discussed upstream first, to avoid that we invent something that then becomes IOT2050-only.

PR #229, opened 2021-12-09
WIP: Compatibility with upcoming sbuild support in Isar

WiseLord:
This patchset makes meta-iot2050 compatible with upcoming sbuild support in Isar.
There are 5 patches here.

First 3 of them can be used without sbuild on current 'next':
- tcf-agent: fix debian/rules permissions
- npm: Add libnode-dev build depends to some npm packages
- u-boot-iot2050: Fix building with sbuild

The following patch is related to sbuild, but should also be 'next'-compatible:
- sbuild: npm: introduce and use npm_do_mounts and npm_undo_mounts

And the last patch uses sbuild-specific exports, so it is not compatible with 'next'
- sbuild: Use sbuild-export API for meta-iot2050 packages

The patchset is tested with the following changes in kas-container:

```
diff --git a/kas-container b/kas-container
index 787f65b..7277196 100755
--- a/kas-container
+++ b/kas-container
@@ -315,9 +315,13 @@ if [ "$(id -u)" -eq 0 ] && [ "${KAS_ALLOW_ROOT}" != "yes" ] ; then
        exit 1
 fi
 
+# We need to use external dirs for schroot overlay
+trace mkdir -p "${KAS_BUILD_DIR}/schroot_overlay"
+
 set -- "$@" -v "${KAS_REPO_DIR}":/repo:${KAS_REPO_MOUNT_OPT} \
        -v "${KAS_WORK_DIR}":/work:rw --workdir=/work \
        -v "${KAS_BUILD_DIR}":/build:rw \
+       -v "${KAS_BUILD_DIR}/schroot_overlay":/var/lib/schroot/union/overlay:rw \
        -e KAS_BUILD_DIR=/build \
        -e USER_ID="$(id -u)" -e GROUP_ID="$(id -g)" --rm
 
diff --git a/kas/iot2050.yml b/kas/iot2050.yml
index 901b164..1e8da00 100644
--- a/kas/iot2050.yml
+++ b/kas/iot2050.yml
@@ -27,12 +27,9 @@ repos:
 
   isar:
-    refspec: e274130b870b7e31532fe191b9932cde1d819b4b
+    refspec: 37a71e72c7b8e9ae7e9f6a402ecdc6ff66eab8c4
     layers:
       meta:
-    patches:
-      u-boot-fix:
-        path: isar-patches/0001-u-boot-Exclude-libubootenv.a-from-dh_strip.patch
 
 bblayers_conf_header:
   standard: |
```
Also, modified docker image currently should be used:

```
export KAS_CONTAINER_IMAGE_PATH=ghcr.io/wiselord
export KAS_IMAGE_VERSION=2.6.2-sbuild
```


recipes-bsp/u-boot/u-boot-iot2050.inc
```diff
@@ -51,7 +51,9 @@ dpkg_runbuild_prepend() {
 do_build[dirs] += "${DEPLOY_DIR_IMAGE}"
 
 do_deploy() {
-    cp -f "${S}/${SPI_FLASH_IMG}" "${DEPLOY_DIR_IMAGE}/${SPI_FLASH_DEPLOY_IMG}"
+     dpkg --fsys-tarfile ${WORKDIR}/${PN}_${PV}_${PACKAGE_ARCH}.deb | \
+        tar xOf - ./usr/lib/u-boot/${MACHINE}/flash.bin \
+        > "${DEPLOY_DIR_IMAGE}/${SPI_FLASH_DEPLOY_IMG}"
```
jan-kiszka, line 56: wrong indention

classes/npm.bbclass
```diff
@@ -19,6 +20,30 @@ ISAR_CROSS_COMPILE = "0"
...
+        if [ `expr $i % 100` -eq 0 ] ; then
+            bbwarn "${BUILDROOT}: Couldn't unmount ($i), retrying..."
+        fi
+        if [ $i -ge 10000 ]; then
+            bbfatal "${BUILDROOT}: Couldn't unmount after timeout"
+        fi
+        i=`expr $i + 1`
+    done
+    sudo rmdir ${BUILDROOT}
+}
```
jan-kiszka, line 45: This logic looks generic and should be provided by Isar, not copied into multiple downstream layers.

jan-kiszka: But I assume that Adriaan's mounting cleanup class will resolve that issue upstream. Maybe we should poke him to prioritize that work.

recipes-bsp/k3-rti-wdt/k3-rti-wdt_1.1.bb
```diff
@@ -26,5 +26,5 @@ do_prepare_build() {
 }
 
 dpkg_runbuild_prepend() {
-    export CC="gcc"
+    sbuild_export CC="gcc"
```
jan-kiszka, line 29: As already commented on upstream: this naming is wrong. Don't spread it to downstream layers, fix it in Isar first.

recipes-app/node-red/node-red_2.0.6.bb
```diff
@@ -13,6 +13,8 @@ inherit npm
 
 DESCRIPTION = "A visual tool for wiring the Internet of Things"
 
+DEBIAN_BUILD_DEPENDS =. "libnode-dev,"
+
```
jan-kiszka, line 17: For convenience reasons, I'm inclined to add this npm.bblcass, even if not all npm packages may contain native elements that require this. You could make that dep conditional, only adding it if NPM_REBUILD is set.

jan-kiszka:
Marked it "WIP" / "Draft" to clarify that this is not yet merge material but shall help with the Isar interface discussion.

PR #231, opened 2021-12-16
Revert "u-boot-iot2050: Add patch to re-write vars of invalid board i…

AsuraZeng:
…nfo"

This reverts commit 7d3eec8a9bf7c09e538bf43dcc0e3aacc1a8d012.


jan-kiszka:
For the commit log: "It turned out to not solve the problem completely. Instead, invalid environments will now be cleaned so that reprogramming is properly triggered on the next boot."

AsuraZeng:
ok.  I have added to the commit message

jan-kiszka (approved):

PR #232, opened 2021-12-16
linux-iot2050: The green light is out of control when panic occurs

AsuraZeng:
The green light is out of control when panic occurs.
it should be off then.

Indicates that the green light is off when panic.Add configuration
to the device tree to fix it.


jan-kiszka:
Should go into a separate patch: It does not use upstream mechanisms, and the modified patch is actually a clean upstream backport.

AsuraZeng:
Use one standlone patch to fix this issue.

jan-kiszka (approved):

PR #236, opened 2021-12-27
Split the firmware building for PG1 and PG2

BaochengSu:
Since the new signature mechenism is used on PG2, we have to split the firmware...


jan-kiszka:
If we model this directly via U-Boot, we can drop unused artifacts (sysfw, device trees) from the the no longer needed unified image. And we can convert runtime detections (`board_is_sr1()`) into compile-time selections. Same goes for the internals of SE-Boot.

BaochengSu:
> If we model this directly via U-Boot, we can drop unused artifacts (sysfw, device trees) from the no longer needed unified image. And we can convert runtime detections (`board_is_sr1()`) into compile-time selections. Same goes for the internals of SE-Boot.

poc @ https://github.com/BaochengSu/u-boot/commit/808c79050b0a1ee2e2b04e38f3d79e3ac2483cdc, how about it?

jan-kiszka:
Yes, that u-boot commit is going into the direction I had in mind.

As discussed privately, we can do that step-wise, and first merge a recipe split here. I'll comment on details of this approach now.

conf/machine/iot2050.conf
```diff
@@ -18,3 +18,5 @@ IMAGE_TYPE ?= "wic-img"
 WKS_FILE ?= "iot2050.wks.in"
 
 IMAGE_INSTALL += "u-boot-script"
+
+IOT2050_VARIANTS ??= "PG1"
```
jan-kiszka, line 22: This is better modeled via separate u-boot recipes and, thus, differences in PREFERRED_PROVIDER_u-boot-iot2050.

recipes-bsp/u-boot/u-boot-iot2050.inc
```diff
@@ -36,6 +36,21 @@ U_BOOT_CONFIG_PACKAGE = "1"
...
+            ;;
+        *"PG2"*)
+            echo -n "Linking SBL for PG2."
+            ln -sf ../prebuild/tiboot3.bin_PG2 ${S}/tiboot3.bin
+            ;;
+        *)
+            bbfatal "IOT2050_VARIANTS=${IOT2050_VARIANTS}, it must contain PG1 or PG2!"
+            ;;
+    esac
+
```
jan-kiszka, line 53: See my comment above: This would move into the specific recipe file (-pg1 vs. -pg2).

conf/machine/iot2050.conf
```diff
@@ -10,11 +10,13 @@
...
+PREFERRED_PROVIDER_u-boot-${MACHINE} = "u-boot-iot2050-${IOT2050_VARIANTS}"
 
 KERNEL_NAME ?= "iot2050"
 
 IMAGE_TYPE ?= "wic-img"
 WKS_FILE ?= "iot2050.wks.in"
 
 IMAGE_INSTALL += "u-boot-script"
+
+IOT2050_VARIANTS ??= "pg1"
```
jan-kiszka, line 22: `IOT2050_VARIANTS` is not needed. Default to one variant in this conf, likely pg2, and change PREFERRED_PROVIDER_u-boot-iot2050 in the kas option.

kas/opt/pg2.yml
```diff
@@ -0,0 +1,16 @@
...
+# This file is subject to the terms and conditions of the MIT License.  See
+# COPYING.MIT file in the top-level directory.
+#
+
+header:
+  version: 10
+
+local_conf_header:
+  board_variant: |
+    IOT2050_VARIANTS ?= "pg2"
```
jan-kiszka, line 16: One of these option files is not needed.

recipes-bsp/u-boot/u-boot-iot2050-pg1.bb
```diff
@@ -0,0 +1,16 @@
...
+#  Su Baocheng <baocheng.su@siemens.com>
+#
+# This file is subject to the terms and conditions of the MIT License.  See
+# COPYING.MIT file in the top-level directory.
+#
+
+require u-boot-iot2050_2021.04.inc
+
+do_prepare_build_append() {
+    cd ${S}
```
jan-kiszka, line 14: Not needed.

recipes-bsp/u-boot/u-boot-iot2050-pg2.bb
```diff
@@ -0,0 +1,16 @@
...
+#  Su Baocheng <baocheng.su@siemens.com>
+#
+# This file is subject to the terms and conditions of the MIT License.  See
+# COPYING.MIT file in the top-level directory.
+#
+
+require u-boot-iot2050_2021.04.inc
+
+do_prepare_build_append() {
+    cd ${S}
```
jan-kiszka, line 14: same here

recipes-bsp/u-boot/u-boot-iot2050.inc
```diff
@@ -20,7 +20,7 @@ U_BOOT_CONFIG = "iot2050_defconfig"
 U_BOOT_BIN = "flash.bin"
 
 SPI_FLASH_IMG = "${U_BOOT_BIN}"
-SPI_FLASH_DEPLOY_IMG = "iot2050-image-boot.bin"
+SPI_FLASH_DEPLOY_IMG = "iot2050-${IOT2050_VARIANTS}-image-boot.bin"
```
jan-kiszka, line 23: Set this var in the including recipe, then `IOT2050_VARIANTS` is not needed also for this use case.

recipes-bsp/u-boot/u-boot-iot2050_2021.04.inc
```diff
@@ -1,15 +1,17 @@
...
 #  Jan Kiszka <jan.kiszka@siemens.com>
+#  Su Baocheng <baocheng.su@siemens.com>
 #
 # This file is subject to the terms and conditions of the MIT License.  See
 # COPYING.MIT file in the top-level directory.
 #
 
 require u-boot-iot2050.inc
 
+PV = "2021.04"
```
jan-kiszka, line 14: Not needed. Will be derived from the including recipe.

BaochengSu: Seems needed, see: https://github.com/siemens/meta-iot2050/runs/4709773299?check_suite_focus=true

recipes-bsp/u-boot/u-boot-iot2050-pg2.bb
BaochengSu: `cd ${S}` deleted.

recipes-bsp/u-boot/u-boot-iot2050-pg1.bb
BaochengSu: cd ${S} deleted.

kas/opt/pg2.yml
BaochengSu: The one for PG2 is deleted.

conf/machine/iot2050.conf
BaochengSu: Done

BaochengSu:
```
local_conf_header:
  u-boot: |
    PREFERRED_PROVIDER_u-boot-${MACHINE} = "u-boot-iot2050-pg1"
```

Doesn't work for increment build, the `build/conf/local.conf` changed accordingly however isar doesn't hornor this change.

jan-kiszka: Missing weak assignment in the distro.conf?

recipes-bsp/u-boot/u-boot-iot2050_2021.04.inc
jan-kiszka: Yes, missed that you didn't properly versioned the recipes ("..._2021.04"). Fix that, and this will work.

BaochengSu:
Yep. Actually this comments has been deleted by me, seems the network didn't work as expected again.

BaochengSu:
I was thinking of add the PV to the filename of pg1.bb and pg2.bb, however, don't you think this seems tedious and kind of ugly to have triple `2021.04`?

conf/machine/iot2050.conf
BaochengSu: Yep. Actually this comments has been deleted by me, seems the network didn't work as expected again.

recipes-bsp/u-boot/u-boot-iot2050_2021.04.inc
BaochengSu: I was thinking of add the PV to the filename of pg1.bb and pg2.bb, however, don't you think this seems tedious and kind of ugly to have triple `2021.04`?

jan-kiszka: Nope, why? We do this for the kernel as well.

recipes-bsp/u-boot/u-boot-iot2050.inc
```diff
@@ -20,7 +20,7 @@ U_BOOT_CONFIG = "iot2050_defconfig"
 U_BOOT_BIN = "flash.bin"
 
 SPI_FLASH_IMG = "${U_BOOT_BIN}"
-SPI_FLASH_DEPLOY_IMG = "iot2050-image-boot.bin"
+SPI_FLASH_DEPLOY_IMG ??= "iot2050-image-boot.bin"
```
jan-kiszka, line 23: Does this make any sense? Nothing is producing that image. I would just drop this line.

BaochengSu: I think `u-boot-iot2050_202x.xx-upstream.bb`  may still need it. Or just delete this upstream bb?

jan-kiszka: In fact, upstream needs to be converted as well in order to still work on both variants... Good point.

BaochengSu: That could be going to another PR... not on the critial path.

There are some new upstreaming patches anyway, the second stage for this PR, maybe this upstream bb could be addressed in that stage.

jan-kiszka: OK, then break the upstream recipe again for now.

jan-kiszka (approved):

PR #240, opened 2022-01-07
Update Isar, enable caching

jan-kiszka:
This brings both ccache and (automatically) sstate caches. Both should accelerate partial rebuilds.


jan-kiszka:
Extracted from the meta-coral work. Will also be helpful for #115.

BaochengSu (approved):

PR #244, opened 2022-01-13
npm: Skip dh_strip_nondeterminism for npm packages

jan-kiszka:
As our packaging is not clean anyway and this step takes a long while
with packages that have many files, maybe even longer with bullseye now,
simply skip it. Saves several minutes build time for many packages.

### DIFF
--- a/classes/npm.bbclass
+++ b/classes/npm.bbclass
@@ -243,3 +243,11 @@ do_install() {
 
     dpkg_undo_mounts
 }
+
+do_prepare_build_append() {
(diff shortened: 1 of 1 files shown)


BaochengSu (approved):

jan-kiszka:
Because it's an evil hack for an evil problem: dependency explosion with NodeJS packages.

The proper solution would be individual packaging of dependencies, not the vendoring that this class does. But it's unrealistic to maintain such a huge pile of dependencies as Debian packages, often even for the Debian community.

PR #252, opened 2022-01-25
sbuild v4 - Compatibility with upcoming sbuild support in Isar

WiseLord:
This pull-request is a replacement for #229

This patchset makes meta-iot2050 compatible with upcoming sbuild support in Isar.
Patchset: [[PATCH v4 00/12] Sbuild/Schroot migration](https://groups.google.com/g/isar-users/c/1aMy8l8CtBU)

First 3 of patches can be used without sbuild on current 'next':
- tcf-agent: fix debian/rules permissions
- u-boot-iot2050: Fix building with sbuild
- npm: Add libnode-dev build depends

And the last patch uses sbuild-specific exports, so it is not compatible with 'next', but is compatible with "pre-sbuild" patchset [[PATCH v2 0/9] Avoid using shell environment during the build](https://groups.google.com/g/isar-users/c/lICFKdfsJbo)
- sbuild: Use dpkg_build_export API for meta-iot2050 packages

The patchset is tested with the [following changes](https://github.com/WiseLord/kas/commit/a554e4a1c650efeb0c6c60e98790bc9c1610cc47) in kas.


jan-kiszka:
Please split this PR so that part one can already be tested and merged.

Part two is missing the Isar revision update. But that can only be done once the feature hit upstream. BTW, I wonder if upstream will have some means to detect now failing exports, though I do not have a good idea on the "how" yet.

classes/npm.bbclass
```diff
@@ -35,6 +35,8 @@ OWN_NPM_CLASS_PACKAGE ?= "0"
 # needed as gyp from bullseye does not establish /usr/bin/python
 NPM_EXTRA_DEPS = "${@'python' if d.getVar('NPM_REBUILD') == '1' else ''}"
 
+DEBIAN_BUILD_DEPENDS =. "${@'libnode-dev,' if d.getVar('NPM_REBUILD') == '1' else ''}"
+
```
jan-kiszka, line 39: BTW, I cannot reproduce the problem: According to the issue description, `kas-container build kas-iot2050-example.yml --target node-red-node-serialport` should already trigger it today, without sbuilder, because there will be no mraa built in that case. But something else pulls it, indirectly. I wonder what that is.

WiseLord: I rechecked it, monitoring buildchroot's `/var/cache/apt/archives` content at the same time. It seems that `libnode-dev` is installed by `do_install_npm()` (and not comes with `mraa` package that I first thought) - at least, `libnode-dev_12.22.5~dfsg-2~11u1_arm64.deb` file appeared here at this time. That's why buildchroot-based build works OK.

But in case of sbuild, build environment doesn't include libnode-dev, that's why an error appears.

WiseLord: Here you can see the build log
[node-red-node-serialport_sbuild-log.txt](https://github.com/siemens/meta-iot2050/files/7948352/node-red-node-serialport_sbuild-log.txt)

WiseLord: Actually, it seems that missing `libnode.so.72` library belongs to another package, `libnode72`, so the dependency tree could be smaller.

jan-kiszka: Please clarify this before sending the split PR and also update the commit log of this patch accordingly.

WiseLord: I rechecked it and four packages mentioned in previous pull-request (`node-red-contrib-modbus`, `node-red-node-serialport`, `node-red-node-sqlite` and `node-red`) are really depend on `libnode.so.72` library provided by `libnode72` package in bullseye.
So, we can add this dependency either to 4 recipes, or to npm.bbclass.

jan-kiszka: Let's keep this injected globally via `NPM_REBUILD` condition. It only applies to nodes with native elements, but it's not worth to track that at node-level.

WiseLord: OK

PR #257, opened 2022-01-27
Ubely/pre sbuild v4

WiseLord:
A continuation of https://github.com/siemens/meta-iot2050/pull/252


WiseLord:
Last patch `sbuild: Use dpkg_build_export API for meta-iot2050 packages` is not included

recipes-bsp/u-boot/u-boot-iot2050.inc
```diff
@@ -48,16 +48,10 @@ dpkg_runbuild_prepend() {
...
-
-dpkg_build_sstate_finalize_append() {
-    ln -f ${DPKG_SSTATE}/${SPI_FLASH_IMG} -t ${S}
-}
-
 do_deploy() {
-    cp -f "${S}/${SPI_FLASH_IMG}" "${DEPLOY_DIR_IMAGE}/${SPI_FLASH_DEPLOY_IMG}"
+    dpkg --fsys-tarfile ${WORKDIR}/${PN}_${PV}_${PACKAGE_ARCH}.deb | \
+        tar xOf - ./usr/lib/u-boot/${MACHINE}/${SPI_FLASH_IMG} \
+        > "${DEPLOY_DIR_IMAGE}/${SPI_FLASH_DEPLOY_IMG}"
```
jan-kiszka, line 54: Looks like this was not tested against master. Please do so.

jan-kiszka:
```
ERROR: u-boot-iot2050-pg2-2021.04-r0 do_deploy: Execution of '/build/tmp/work/iot2050-debian-arm64/u-boot-iot2050-pg2/2021.04-r0/temp/run.do_deploy.30011' failed with exit code 2:
dpkg-deb: error: failed to read archive '/build/tmp/work/iot2050-debian-arm64/u-boot-iot2050-pg2/2021.04-r0/u-boot-iot2050-pg2_2021.04_arm64.deb': No such file or directory
tar: This does not look like a tar archive
tar: ./usr/lib/u-boot/iot2050/flash.bin: Not found in archive
tar: Exiting with failure status due to previous errors
WARNING: exit code 2 from a shell command.
```

jan-kiszka: Looking at it, this may actually be a bug in the existing recipe, just exposed now. Still, patches should be tested!

jan-kiszka: Nope, debian package name is `u-boot-${MACHINE}` and stays like this, independent of the recipe name (which controls the variant). A bit special in this case. So, please adjust.

WiseLord: I seem to miss this error due to the strange thing: running `./kas-container build kas-iot2050-example.yml` completed without errors and I didn't find logs for 'deploy' - it simply was not executed.

When I've run the shell and try to build u-boot in it - yes, I caught the same problem. Already fixed and will be sent a bit later.

I'm trying to understand why skipping deploy task could happen. Probably, removing dpkg_build_state_* stuff from the recipe could somehow make an impact.

WiseLord: Investigation through `bitbage -g` shows that `u-boot-iot2050-pg2.do_deploy` is not in task tree when building any image. So, flash.bin is never deployed when building any of `iot2050-image-*`.

jan-kiszka: That's true for disk image building which only happens to build u-boot for the purpose of u-boot-iot2050-config. The case above was triggered by building one of the firmware images.

jan-kiszka:
Obsoleted by #258. You could have pushed to the same branch to simply update this PR.

PR #258, opened 2022-01-28
Ubely/pre_sbuild v4 1

WiseLord:
A continuation of #257


jan-kiszka (dismissed):

WiseLord:
This patchset is compatible with upcoming sbuild patchset, where some 'export' related things were added.
Briefly, there is an environment comparison between `do_prepare_build` and `dpkg_runbuild` tasks is done, so any variable exported in `dpkg_runbuild_prepend` goes to sbuild environment.
So, there is no need to have one more patch (export => dpkg_build_export) like it was done in https://github.com/siemens/meta-iot2050/pull/252

jan-kiszka (commented):
I think we first need to clarify in Isar upstream if we want to keep `export` support or rather push toward a migration.

PR #261, opened 2022-02-10
Ubely/sbuild

WiseLord:
Some modifications for the recent sbuild (https://github.com/ilbers/isar/tree/ubely/sbuild) compatibility
- avoid shell exports in the recipes (move them to debian/rules)
- support DEB_BUILD_OPTIONS recently merged to isar/next
- modifications for npm.bbclass:
  - buildchroot is not used for pre-build tasks;
  - **host npm** used to fetch bundled packages;
  - target npm used (as build dependency) to install/compile npm packages;

Patches tested and work both with current `isar/next` and `ubely/sbuild`

First 3 patches in the chain were already used in previous pull-request https://github.com/siemens/meta-iot2050/pull/258


jan-kiszka:
Given that we still have problems with the sbuild design, I would focus on obvious cleanups first (tcf-agent, openssl, second half of u-boot-iot2050) and leave the rest aside until those topics have settled upstream. This PR remains WIP in my eyes.

jan-kiszka (commented):
How is this supposed to work? There is no npm package installed on a random host.

fetch_npm must run over a well-defined environment where the build triggered the installation of npm. OTW, we will continue to require some own chroot here.
@fmoessbauer

fmoessbauer:
As the sbuild support is already integrated for some time, we should close #261 and #258 . @BaochengSu

PR #259, opened 2022-01-29
Integrate support for Coral 

jan-kiszka:
Add Google Coral packages to the example image by default. This increases our storage demand, so include the github running VM cleanup that @gylstorffq developed for #115.

PR #264, opened 2022-02-17
Fix SWUpdate build

gylstorffq:
Fix the swupdate build error by bumping isar-cip-core to ed318f247beef10e0433c6ed699b4c1ab008fa6a.

Also bump ISAR and meta-coral refspec to drop the patches introduced by https://github.com/siemens/meta-iot2050/pull/259.

On my machine the build needs currently 52G disk memory.

An out-of-disk error occurred during building. Free disk space by purging the following packages:
- dotnet
- google-chrom-stable
-  temurin java
- google-cloud-sdk
- hhvm
- azure-cli
- llvm*

Also delete all content below /opt and /usr/local.

This leads to free disk space of 65G.

This is a workaround until a slim build environment is provided, see https://github.com/actions/virtual-environments/issues/5098.

### DIFF
--- a/.github/workflows/free-disk-space/action.yml
+++ b/.github/workflows/free-disk-space/action.yml
@@ -3,6 +3,11 @@ description: 'Remove content of the provided image to free disk space for the bu
 runs:
   using: "composite"
   steps:
-      - name: Remove /usr/local/* to free disk space
-        run:  sudo rm -rf /usr/local/*
+      - name:  Free disk space
+        run: |
+          sudo apt-get purge 'dotnet*' google-chrome-stable 'temurin*' google-cloud-sdk hhvm azure-cli 'mongodb*' powershell firefox chromium 'llvm*' 'libllvm*' 'adoptopenjdk*' 'mysql*' libgl1-mesa-dri apache2 'mono*'
+          sudo rm -rf /usr/local/* /opt/hostedtoolcache /opt/pipx* /opt/az
+        shell: bash
+      - name: Show disk space usage
+        run:  df -h
         shell: bash
--- a/kas/iot2050.yml
+++ b/kas/iot2050.yml
@@ -27,24 +27,17 @@ repos:
 
   isar:
     url: https://github.com/ilbers/isar
-    refspec: 74036ebaab5265d62cdf90f563acd5b4d3b8b6d2
+    refspec: cf3f18a649ff42726677a58b224b9bb58ea4ba46
     layers:
       meta:
-    patches:
-      deb-build-profiles:
-        repo: meta-coral
-        path: patches/isar/0001-add-support-for-debian-build-profiles-and-options.patch
-      adjust-kernel-recipe:
-        repo: meta-coral
-        path: patches/isar/0002-refactor-linux-custom.inc-to-use-ISAR-s-DEB_BUILD_PR.patch
 
   meta-coral:
     url: https://github.com/siemens/meta-coral
-    refspec: f0202eaabc39b167d3e86e8595c6f97331f29293
+    refspec: 9d0a1f60ae955ee2edca6caf98312e79accb98e2
 
   cip-core:
     url: https://gitlab.com/cip-project/cip-core/isar-cip-core.git
-    refspec: 0feddaa6dd65c5a1b56ff6fcb18b4e748e53972f
+    refspec: ed318f247beef10e0433c6ed699b4c1ab008fa6a
 
 bblayers_conf_header:
   standard: |


.github/workflows/free-disk-space/action.yml
```diff
@@ -6,3 +6,18 @@ runs:
...
+        shell: bash
+      - name: Remove temurin-jdk to free disk space
+        run:  sudo apt purge temurin*
+        shell: bash
+      - name: Remove google-cloud-sdk to free disk space
+        run:  sudo apt purge google-cloud-sdk
+        shell: bash
+      - name: Remove hhvm to free disk space
+        run:  sudo apt purge hhvm
+        shell: bash
```
jan-kiszka, line 23: Amazing - I wonder how long we can play this game.

BTW, 'shell: bash' is needed for all those calls?

kas/iot2050.yml
```diff
@@ -33,7 +33,7 @@ repos:
 
   meta-coral:
     url: https://github.com/siemens/meta-coral
-    refspec: f0202eaabc39b167d3e86e8595c6f97331f29293
+    refspec: 9d0a1f60ae955ee2edca6caf98312e79accb98e2
```
jan-kiszka, line 36: This removes patches - those that we still use at this stage...?

kas/iot2050.yml
```diff
@@ -27,16 +27,9 @@ repos:
...
+    refspec: cf3f18a649ff42726677a58b224b9bb58ea4ba46
     layers:
       meta:
-    patches:
-      deb-build-profiles:
-        repo: meta-coral
-        path: patches/isar/0001-add-support-for-debian-build-profiles-and-options.patch
-      adjust-kernel-recipe:
-        repo: meta-coral
-        path: patches/isar/0002-refactor-linux-custom.inc-to-use-ISAR-s-DEB_BUILD_PR.patch
```
jan-kiszka, line 39: This patch likely has to come before the meta-coral dump, see my earlier comment.

.github/workflows/free-disk-space/action.yml
gylstorffq: I think we are at the end of the game. We currently need 52G and this will increase with the next feature. I set it to draft to find solution which future prove instead of purging a pill  of packges.

kas/iot2050.yml
```diff
@@ -44,7 +44,7 @@ repos:
 
   cip-core:
     url: https://gitlab.com/cip-project/cip-core/isar-cip-core.git
-    refspec: 0feddaa6dd65c5a1b56ff6fcb18b4e748e53972f
+    refspec: ed318f247beef10e0433c6ed699b4c1ab008fa6a
```
jan-kiszka, line 40: Has to come in lock-step with the isar update - like we did in isar-cip-core as well.

.github/workflows/free-disk-space/action.yml
jan-kiszka: None of the calls rely in bash behavior, that's why I don't see that need.

.github/workflows/free-disk-space/action.yml
```diff
@@ -5,4 +5,31 @@ runs:
...
+        shell: /usr/bin/bash -e {0}
+      - name: Remove azure-cli to free disk space
+        run:  sudo apt purge azure-cli
+        shell: /usr/bin/bash -e {0}
+      - name: Remove llvm-* to free disk space
+        run:  sudo apt purge llvm-*
+        shell: /usr/bin/bash -e {0}
+      - name: Show disk space usage
+        run:  df -h
+        shell: /usr/bin/bash -e {0}
```
jan-kiszka, line 35:
```
run: |
  sudo apt-get purge 'dotnet*' google-chrome-stable 'temurin*' google-cloud-sdk hhvm azure-cli 'mongodb*' powershell firefox chromium 'llvm*' 'libllvm*' 'adoptopenjdk*' 'mysql*' libgl1-mesa-dri apache2 'mono*'
  sudo rm -rf /usr/local/* /opt/hostedtoolcache /opt/pipx* /opt/az
```
gives us 64G

jan-kiszka: And please drop those pointless `shell:` statements.

jan-kiszka: Interesting links:
https://github.com/ThewApp/free-actions/blob/main/.github/workflows/free.yml
https://github.com/actions/virtual-environments/discussions/3242

jan-kiszka: +4G if you also purge docker images.

.github/workflows/free-disk-space/action.yml
```diff
@@ -3,6 +3,11 @@ description: 'Remove content of the provided image to free disk space for the bu
...
   using: "composite"
   steps:
-      - name: Remove /usr/local/* to free disk space
-        run:  sudo rm -rf /usr/local/*
-        shell: bash
+      - name:  Free disk space
+        run: |
+          sudo apt-get purge 'dotnet*' google-chrome-stable 'temurin*' google-cloud-sdk hhvm azure-cli 'mongodb*' powershell firefox chromium 'llvm*' 'libllvm*' 'adoptopenjdk*' 'mysql*' libgl1-mesa-dri apache2 'mono*'
+          sudo rm -rf /usr/local/* /opt/hostedtoolcache /opt/pipx* /opt/az
+        shell: /usr/bin/bash -e {0}
```
jan-kiszka, line 10: Drop `shell:`.

.github/workflows/free-disk-space/action.yml
```diff
@@ -3,6 +3,11 @@ description: 'Remove content of the provided image to free disk space for the bu
...
+      - name: Show disk space usage
+        run:  df -h
+        shell: /usr/bin/bash -e {0}
```
jan-kiszka, line 13: Also here.

.github/workflows/free-disk-space/action.yml
jan-kiszka: Seems we need that in own actions, probably to avoid that they use a default set by the including workflow. Then `shell: bash` should be sufficient.

jan-kiszka (approved):

PR #278, opened 2022-03-10
u-boot: fix the issue that board can't load the correct dtb

AsuraZeng:
the name like k3-am6528-iot2050-basic-pg2 not the iot2050-basic-pg2.
so it can't find the correnct dtb.

correct this issue

### DIFF
--- a/recipes-bsp/u-boot/files/0001-iot2050-binman-add-missing-msg-for-blobs.patch
+++ b/recipes-bsp/u-boot/files/0001-iot2050-binman-add-missing-msg-for-blobs.patch
@@ -1,7 +1,7 @@
-From 4415d5cba8301616ba5fbd304f14425b787e830e Mon Sep 17 00:00:00 2001
+From f306b24341841d9d5bf3d1be5155f46e2c6c9d95 Mon Sep 17 00:00:00 2001
 From: Ivan Mikhaylov <ivan.mikhaylov@siemens.com>
 Date: Thu, 9 Dec 2021 16:10:53 +0000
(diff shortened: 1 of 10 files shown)


jan-kiszka:
Let me check the causing commit, and then we will fix that one up. No need for stacking things.

AsuraZeng:
> Let me check the causing commit, and then we will fix that one up. No need for stacking things.

this original commit has already upstream,not int the currently patches list.

jan-kiszka:
Unlikely - we added support for PG2 and tested that before upstreaming it.

AsuraZeng:
this is a new bug introduced after the commit 5ac8edaca2fd29c87c2bd4922b464b31a6dbd674.
and this bug is caused by using the fdt-SEQ.

jan-kiszka:
NACK: I've just built the image for PG2 via master and tested it on PG2 Basic. It's all fine. I suspect you accidentally built for PG1.

jan-kiszka:
OK, reproduced now: We better cut off that new prefix which confuses the match. strstr could eventually stumble over "name" vs. "name-suffix". The fix should go into the causing commit (https://github.com/siemens/u-boot/commit/5cf7130b85e9e76ffee765aabb56a972fbe050ed). I'll update the queues.

AsuraZeng:
> OK, reproduced now: We better cut off that new prefix which confuses the match. strstr could eventually stumble over "name" vs. "name-suffix". The fix should go into the causing commit ([siemens/u-boot@5cf7130](https://github.com/siemens/u-boot/commit/5cf7130b85e9e76ffee765aabb56a972fbe050ed)). I'll update the queues.

Ok, I  agree. I would drop this PR.

PR #279, opened 2022-03-10
u-boot-iot2050: Update patch queue to fix detection of Advanced devices

jan-kiszka:
The introduction of generated config nodes had another trap which is now
resolved.

Should obsolete #278.


AsuraZeng (approved):
looks good to me

BaochengSu (approved):

PR #283, opened 2022-03-17
openssl: Drop pinning to BASE_DISTRO_CODENAME

jan-kiszka:
This is neither needed anymore because we have no additional version in
out apt sources. Nor does it work when there is an update in -security
while the pinning pulls from the release repo:

The following packages will be DOWNGRADED:
  libssl1.1
0 upgraded, 580 newly installed, 1 downgraded, 0 to remove and 0 not upgraded.
E: Packages were downgraded and -y was used without --allow-downgrades.


BaochengSu (approved):

BaochengSu:
LGTM, let's wait for the CI results.

PR #269, opened 2022-02-23
Chao/bump version

AsuraZeng:

PR #285, opened 2022-03-17
Fix Node RED packages / cleanup customization postinst

jan-kiszka:
Primarily, this brings the correction of the npm-shrinkwrap files that became broken with the last update (see #269). The reason for the breakage was an outdated description of the steps to perform the generation of those files. This changed with npm 7.x, as can be found in Debian 11. This update was now performed following the fixed procedure under Debian 11.

This PR also cleans up the postinst script of cutomizations-example from unneed/wrong systemd service enablings.


classes/npm.bbclass
```diff
@@ -3,7 +3,8 @@
 # SPDX-License-Identifier: MIT
 
 # HOWTO generate an npm-shrinkwrap.json:
-#   npm install --global-style <my-favorite-package>
+#   npm config set global-style=true (you may want to reset this afterward)
```
AsuraZeng, line 6: I used -- global style to generate node red based on NPM 7.5.2
Then the shrink file is exported and tested, which seems to be correct

jan-kiszka: As `--global-style` was swallowed but had on effect, you created a flattened installation, and all dependencies landed outside of the node-red module (as it is the default). But that cause the packaging to ignore them, and Node RED was missing them on the target.

BaochengSu (approved):

PR #288, opened 2022-03-20
Drop version warning from stable as well

jan-kiszka:


BaochengSu:
I think maybe we should delivery the information regarding which example image version should work on which firmware version to users, otherwise such kind of issues would be more and more.

BaochengSu (approved):

PR #292, opened 2022-03-23
Add package mtd-utils

BaochengSu:
Prepare for the upcoming secure boot.

### DIFF
--- a/recipes-core/images/iot2050-image-example.bb
+++ b/recipes-core/images/iot2050-image-example.bb
@@ -59,6 +59,7 @@ IOT2050_DEBIAN_DEBUG_PACKAGES = " \
     rsyslog \
     net-tools \
     i2c-tools \
+    mtd-utils \
     sudo \
     "
 


jan-kiszka:
What's this for? No reasoning is given.

BaochengSu:
> What's this for? No reasoning is given.

Revised to:

Prepare for the upcoming secure boot.

The otp-related commands need be programmed into mtd partitions under
linux to provision customer keys, hence need this mtd-utils tools to
integrate these mtd tools into the example image so that user can use it
directly without install it manually.

jan-kiszka:
What more than `cat data > /dev/mtdX` do you need?

BaochengSu:
SPI Flash need be erased before writing, not sure of cat would do that erasing work...

jan-kiszka:
Sure, look at iot2050-firmware-update on how we do that so far.

BaochengSu:
there is a 

```
    @staticmethod
    def flash_erase(dev, start, nbytes):
        """This function erases flash sectors
        @dev: flash device file descriptor
        @start: start address
        @nbytes: number of bytes to erase
        """
        MEMERASE = 0x40084d02

        ioctl_data = struct.pack('II', start, nbytes)

        try:
            fcntl.ioctl(dev, MEMERASE, ioctl_data)
        except IOError:
            print("ioctl failed")
            sys.exit(1)
```

before writing.

jan-kiszka:
Exactly. What we are missing here now is what you plan to do with mtd-utils and why you don't write another python script instead.

BaochengSu:
Why not using exsiting tools instead of writing new scripts? That sounds `reinvent the wheel` to me.

jan-kiszka:
That's fine if the steps you have to do, the logic required here is much simpler than what we had to do in iot2050-firmware-update. I have no idea yet how it will look like, though, that's why I'm challenging you.

BaochengSu:
> That's fine if the steps you have to do, the logic required here is much simpler than what we had to do in iot2050-firmware-update. I have no idea yet how it will look like, though, that's why I'm challenging you.

Oh, if you are asking that then it's a very simple logic:

```
flash_erase /dev/mtd5 0 1
nandwrite /dev/mtd5  otp_data.bin
```

does everything.

jan-kiszka:
If it stays at this level, it's ok. But this is how firmware-update started back then as well.

In any case, please submit this along with the actual tools/scripts. Then we have the full picture of the workflow, documentation - and required dependencies.

jan-kiszka:
...and actually that package will obsolete this patch because it will state its dependency on mtd-utils. So: NACK.

BaochengSu:
> ...and actually that package will obsolete this patch because it will state its dependency on mtd-utils.

Why the FWU package will state its dependency on mtd-utils?

jan-kiszka:
Not the iot2050-firmware-update package. I mean the package that will allow the user to generate those OTP programming commands, sign them, and finally program them onto mtd5.

BaochengSu:
> Not the iot2050-firmware-update package. I mean the package that will allow the user to generate those OTP programming commands, sign them, and finally program them onto mtd5.

Ah, finally I get your meaning, yep, you are right, after all this is an `example` image, especially the secure boot functionality.

I will submit the whole tools together with its dependencies in to a single PR later so that user can cherry picking the commits within that PR.

PR #293, opened 2022-03-24
iot2050-firmware-update: preserve u-boot env variables across updates 

AsuraZeng:
there are some uboot env need to be reserved after firmware upgrade.
fill the reserved env at the reserved_uboot_env filed in the update.conf.json to
keep the uboot env after firmware upgrade

### DIFF
--- a/recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
+++ b/recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 #
-# Copyright (c) Siemens AG, 2020
+# Copyright (c) Siemens AG, 2020-2022
 #
 # Authors:
 #  Chao Zeng <chao.zeng@siemens.com>
@@ -127,14 +127,7 @@ class FirmwareUpdate(object):
             print("ioctl failed")
             sys.exit(1)
 
-    def update_firmware(self):
-        """Update Firmware"""
-        mtd_num = 0
-
-        print("===================================================")
-        print("IOT2050 firmware update started - DO NOT INTERRUPT!")
-        print("===================================================")
-
+    def get_mtd_info(self, mtd_num):
         ospi_dev_path = "/sys/bus/platform/devices/47040000.spi"
         if os.path.exists(ospi_dev_path + "/spi_master"):
             # kernel 5.9 and later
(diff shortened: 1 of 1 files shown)


recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
```diff
@@ -55,14 +55,18 @@
 # 			"key": "VERSION_ID",
 # 			"min_version": "2.1.1"
 # 		}
+# 	],
+# 	"reserved_uboot_env": [
```
BaochengSu, line 59: should be: `preserved_u-boot_env`

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
```diff
@@ -266,21 +282,28 @@ class UpdateConfiguration(object):
...
                 # OS info, or the firmware w/o target_os node which means it doesn't care the OS info.
                 if len(target_os) == 0 or self._check_os(target_os, os_info):
                     res.append(firmware.name)
 
         return res
 
+    def reserved_uboot_env(self):
+        try:
+            reserved_uboot_env = self._jsonobj.reserved_uboot_env
+            return reserved_uboot_env
```
BaochengSu, line 295: why not `return self._jsonobj.reserved_uboot_env`

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
```diff
@@ -266,21 +282,28 @@ class UpdateConfiguration(object):
...
+        except Exception as e:
+            return None
 
 class ManagedFirmwareUpdate(FirmwareUpdate):
     '''
     ManagedFirmwareUpdate will perform various checking upon both the update package and the 
     targeted board, before calling the underlying FirmwareUpdate() to flash the binaries to
-    the SPI flash.
+    the SPI flash.s
```
BaochengSu, line 303: ?

jan-kiszka:
"iot2050-firmware-update: preserve u-boot env variables across updates

Allow to preserve certain u-boot environment variables across firmware update.
This is useful to keep settings like the boot targets or the upcoming M.2 slot configuration.
The list of variables can be configured via the update.conf.json file."

BTW, the M.2 variable is probably too early to add at this stage. We still have to think this through again from a security perspective.

AsuraZeng:
> "iot2050-firmware-update: preserve u-boot env variables across updates
> 
> Allow to preserve certain u-boot environment variables across firmware update. This is useful to keep settings like the boot targets or the upcoming M.2 slot configuration. The list of variables can be configured via the update.conf.json file."
> 
> BTW, the M.2 variable is probably too early to add at this stage. We still have to think this through again from a security perspective.

just a template currently. we can decide later.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
```diff
@@ -289,13 +316,14 @@ class ManagedFirmwareUpdate(FirmwareUpdate):
         self._board_info = BoardInformation()
 
         # Parse the update configuration from the json file within the tarball
-        self._update_conf = UpdateConfiguration(
-            tarfile.open(fileobj=self.firmware).extractfile(self._update_conf_json))
+        self._update_conf = UpdateConfiguration(tarfile.open(
+            fileobj = self.firmware).extractfile(self._update_conf_json))
```
BaochengSu, line 320: IIRC, `=` in arguments should not have spaces around them, I might be wrong, you can do your own research.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
```diff
@@ -312,15 +340,66 @@ class ManagedFirmwareUpdate(FirmwareUpdate):
...
+            f.extract(member=self.uboot_default_env_file, path='.')
+            super().update_firmware()
+
+    def update_uboot_env(self, env_list):
+        '''
+        restore uboot env
+        retrive the uboot default env file belong to the upgrade firmware.
+        makeup the restore env binary and flash to the env mtd partition
+        '''
+        env_default_binary = 'uEnv_default.bin'
```
BaochengSu, line 360: This does not always work if the current dir is read-only.

Maybe try to use some dirs that alreadys has the write permission, such as `/tmp/`.

Also, if `mkenvimage` could have the output piped, you can redirect the output to the std_out and read it via subprocess, IIRC, then the local file could be avoided.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
```diff
@@ -312,15 +340,66 @@ class ManagedFirmwareUpdate(FirmwareUpdate):
...
             firmware_name = firmware_names[choice - 1]
 
         print("Will write {} to the SPI flash".format(firmware_name))
 
         self.firmware.seek(0)
-        self.firmware = tarfile.open(fileobj=self.firmware).extractfile(firmware_name)
-        super().update_firmware()
+        with tarfile.open(fileobj=self.firmware) as f:
+            self.firmware = f.extractfile(firmware_name)
+            f.extract(member=self.uboot_default_env_file, path='.')
```
BaochengSu, line 351: Same here, avoid to extract it, but use it via `f.extractfile()` just as the above line.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
```diff
@@ -312,15 +340,66 @@ class ManagedFirmwareUpdate(FirmwareUpdate):
...
+        with open(self.uboot_default_env_file, encoding="utf-8", mode="a") as file:
+            for value in env_list:
+                file.write(value)
+                file.write("\n")
+
+        for mtd_num in env_partition_list:
+            mtd_dev_path, mtd_size, mtd_erasesize, mtd_name = self.get_mtd_info(
+                mtd_num)
+            subprocess.call('mkenvimage -s {} -r -o {} {}'.format(mtd_size,
+                            env_default_binary, self.uboot_default_env_file), shell=True)
```
BaochengSu, line 373: Use `subprocess.run()` instead, and for the command, do not use the `-o {}` but direct to the `stdout` and capture it from the python script. Also check the `stderr` for robust (the `mkenvimage` command is not guaranteed to always work)

BaochengSu: Just as what you've done below with `subprocess.check_output`.

However, I am not sure if this works with the binary stream. You can have a try.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
AsuraZeng: I am not sure `mkenvimage` has the output piped. maybe the output file is necessary.

BaochengSu: Just verified via 

```
mkenvimage -s 1024 -r input.txt | hexdump
```

It works.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
AsuraZeng: extractfile would return [io.BufferedReader] object

what would we want here is a writing ability.
so for this intent, extract the file and do some operation on this file

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
```diff
@@ -312,15 +340,66 @@ class ManagedFirmwareUpdate(FirmwareUpdate):
...
+            preserved_uboot_env_name = self._update_conf.preserved_uboot_env()
 
+        preserved_uboot_env_value = []
+
+        if not preserved_uboot_env_name:
+            return None
+
+        for env_name in preserved_uboot_env_name:
+            env_value = subprocess.check_output(
+                'fw_printenv %s' % env_name, shell=True).decode('utf-8').lstrip().rstrip()
```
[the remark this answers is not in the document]
AsuraZeng: this is a empty value. would not keep the value

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
AsuraZeng: has removed the redundant blank

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
AsuraZeng: new version use  `/tmp` folder, when reboot, these tmp file would not exist

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
AsuraZeng: use `subprocess.run()` API to handle the command and add exception mechanism

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
```diff
@@ -312,15 +343,79 @@ class ManagedFirmwareUpdate(FirmwareUpdate):
...
+        uboot_env_assemble_file =  os.path.join(self.extract_path, "env_assemble_file")
+        assert os.path.isfile(uboot_default_env_file)
+        # assemble the env
+        shutil.copy(uboot_default_env_file, uboot_env_assemble_file)
+        with open(uboot_env_assemble_file, encoding="utf-8", mode="a") as file:
+            for value in env_list:
+                file.write(value)
+                file.write("\n")
+
+        with tempfile.NamedTemporaryFile(mode="rb") as env_default_binary:
```
BaochengSu, line 376: Why `mode="rb"`? shouldn't be `mode="wb"`?

AsuraZeng: remove the mode "rb" use the default `w+b` mode

BaochengSu (approved):

AsuraZeng:
@jan-kiszka  Please provide your comments, appreciate your feedback

AsuraZeng:
> Strictly spoken, this should be split into style changes/fixes, refactorings and feature additions.

thanks. Has separated  the code according to the modification

BaochengSu (approved):

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
```diff
@@ -330,19 +432,26 @@ def main(argv):
...
     try:
         args = parser.parse_args()
     except IOError as e:
         print(e.strerror, file=sys.stderr)
         exit(1)
 
-    erase_env_input = input("\nWarning: All U-Boot environment variables will be reset to factory settings. Continue (y/N)? ")
+    erase_env_input = input(
+        "\nWarning: All U-Boot environment variables will be reset to factory settings.\n"
+        "Unless preserve env in the update.conf.json or using -p argument.Continue(y/N)?")
```
jan-kiszka, line 447: This could be made more concrete: You do not need to warn if there are no other user modifications noticed and `-p` is provided.

PR #299, opened 2022-04-11
iot2050-image-example: add armhf compat libraries

AsuraZeng:
As could not compile the 32bit application on the native machine.
it should use cross-compile method.
Some 32bit applications need to be transplanted and applied on iot2050,
so add armhf compat libraries.
besides enable compt support,can get corresponding supported sdk

### DIFF
--- a/recipes-core/images/iot2050-image-example.bb
+++ b/recipes-core/images/iot2050-image-example.bb
@@ -81,16 +81,25 @@ IOT2050_DEBIAN_BT_PACKAGES = " \
     bluez \
     pulseaudio-module-bluetooth \
     "
+
 # alsa support
 IOT2050_DEBIAN_ALSA_PACKAGES = " \
     alsa-utils \
     alsa-tools \
     "
+
+# multiarch support
+IOT2050_DEBIAN_MULTIARCH_PACKAGES = " \
+    libc6:armhf \
+    libstdc++6:armhf \
+    "
+
 IMAGE_PREINSTALL += " \
     ${IOT2050_DEBIAN_DEBUG_PACKAGES} \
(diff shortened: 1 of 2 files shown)


recipes-core/images/iot2050-image-example.bb
```diff
@@ -81,16 +81,25 @@ IOT2050_DEBIAN_BT_PACKAGES = " \
...
 # alsa support
 IOT2050_DEBIAN_ALSA_PACKAGES = " \
     alsa-utils \
     alsa-tools \
     "
+
+# multiarch support
+IOT2050_DEBIAN_MULTIARCH_PACKAGES = " \
+    libc6:armhf \
+    libstdc++6:armhf \
```
jan-kiszka, line 94: This should only enable the very basic dependencies in the SDK and on the target. Which cases do we want to address with it?

jan-kiszka (changes requested):
Let's try to polish the commit message a bit:

"To allow transplanting 32-bit applications to the iot2050, enable support for building for the compat arch and add armhf libraries. This also adds this feature to the SDK."

recipes-core/images/iot2050-image-example.bb
AsuraZeng: for pn driver,currently it only supports 32bit.user could port the pn to iot2050 and also could create their pn application by themselves.

AsuraZeng:
update the commit message

jan-kiszka (approved):

BaochengSu:
@AsuraZeng, you need fix your date-time on your PC, this commit was from the future by the time submitting.

PR #304, opened 2022-04-24
Chao/fix partuuid

AsuraZeng:


jan-kiszka:
Sorry, the commit message for patch 2 does not parse for me yet. The purpose of regen-uuid is that the eMMC will get its own, unique UUID that is _different_ from the one of the SD-card. Otherwise, we risk to mount the SD-card on eMMC boot with the card is still plugged.

And patch 1 lacks a reason in its commit message. Every patch must carry a reason for its chance, not a description of it. The latter may only come _on top_ if the code is not obvious.

AsuraZeng:
the bug is that when we flash to emmc, the partuuid of emmc and sd should be the same at first.then we get a chance to replace the `NEW_UUID` using `sed -i 's/'${OLD_UUID}'/'${NEW_UUID}'/i' /etc/default/u-boot-script` in the script
regen-rootfs-uuid.sh

but actually, the uuid of the emmc  and  is not the same sd when dd sd to emmc sometimes.
Could not find the `OLD_UUID` , it would not replace the`OLD_UUID` use  `NEW_UUID` in the boot.scr.

so we pass the boot dev uuid as the original emmc uuid, make the replace work.

jan-kiszka:
OK, interesting finding. But the solution is not yet optimal. Does this already help?
```
diff --git a/recipes-core/install-on-emmc/files/install-on-emmc.sh b/recipes-core/install-on-emmc/files/install-on-emmc.sh
index e73f1978..a598c49d 100755
--- a/recipes-core/install-on-emmc/files/install-on-emmc.sh
+++ b/recipes-core/install-on-emmc/files/install-on-emmc.sh
@@ -124,8 +124,8 @@ blink ${LED_GREEN} 0.25
 SECTORS="$(($(sfdisk -d ${BOOT_DEV} 2>/dev/null | tail -1 | sed 's/.*start=[[:space:]]*\([^,]*\), size=[[:space:]]*\([^,]*\).*/\1+\2/')))"
 
 echo "Writing ${SECTORS} sectors to eMMC ${EMMC_DEV}..."
-dd if=${BOOT_DEV} of=${EMMC_DEV} count=${SECTORS}
-sync
+dd if=${BOOT_DEV} of=${EMMC_DEV} count=${SECTORS} conv=fsync
+partprobe ${EMMC_DEV}
 
 echo "Updating partition UUID of eMMC rootfs"
 partx -a ${EMMC_DEV}
```

AsuraZeng:
I would try to verify the solution.

AsuraZeng:
I have verified this. it works. I would creat a new PR.
then further verification would handle over to our system tester.

jan-kiszka:
Cool, great!

PR #305, opened 2022-04-26
customizations-base: Do not package /etc/hosts, modify it

AsuraZeng:
This avoids conflicts with other packages that want to adjust this file,
e.g. docker.io. And it's simpler because we can now let netbase deliver
the foundation.


jan-kiszka (approved):
Thanks for taking care!

PR #306, opened 2022-04-27
Update README.md

lyxsiemens:
Increase the command to match the size of uboot file and flash chip

### DIFF
--- a/recipes-bsp/u-boot/README.md
+++ b/recipes-bsp/u-boot/README.md
@@ -59,6 +59,14 @@ apt-get install flashrom
 
 Attach the grabber to the flash chip, then run the following on the host machine:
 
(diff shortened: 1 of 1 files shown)


AsuraZeng:
should meet DCO
you should use `git commit -s`

jan-kiszka:
Thanks for your contribution. Besides the formal DCO thing: Can you explain also why this is needed? I'm flashing 8M images for ages. And if it was needed, it should be addressed in U-Boot directly, in the flash layout description.

jan-kiszka:
OK, now I see the context (that not-so-much recommendable ch341a flasher). Please fold both commits into one (`git rebase -i ...`) and mention that context in the commit message.

PR #307, opened 2022-04-27
Chao/fix waiting for partuuid

AsuraZeng:


recipes-core/images/iot2050-image-example.bb
```diff
@@ -60,6 +60,7 @@ IOT2050_DEBIAN_DEBUG_PACKAGES = " \
     net-tools \
     i2c-tools \
     sudo \
+    parted \
```
jan-kiszka, line 63: Belongs as dependency into the package that needs it. Fold it over there into the patch that adds the feature.

AsuraZeng: OK. Have added this package belong to the feature patch.

jan-kiszka: I'm not seeing this yet. It's still only in the image recipe, not in DEBIAN_DEPENDS of install-on-emmc.

recipes-core/install-on-emmc/install-on-emmc_1.0.bb
```diff
@@ -11,7 +11,7 @@ inherit dpkg-raw
 DESCRIPTION = "This service provides the option to install on eMMC during first boot"
 
 DEPENDS = "regen-rootfs-uuid"
-DEBIAN_DEPENDS = "systemd, fdisk, util-linux, regen-rootfs-uuid"
+DEBIAN_DEPENDS = "systemd, fdisk, util-linux, regen-rootfs-uuid parted"
```
jan-kiszka, line 14: Syntax error

jan-kiszka (changes requested):

recipes-core/install-on-emmc/install-on-emmc_1.0.bb
jan-kiszka: (it's a subtle bug I make too often as well, but it is quickly found when running the build)

jan-kiszka (approved):

PR #308, opened 2022-04-27
Update README.md

lyxsiemens:
Increase the command to match the size of uboot file and flash chip

CH341A does not match the size of the uboot file with the size of the flash chip.
When the size of the uboot file is not equal to 16M, if the flashing step is executed directly, an error of size dosen't match will be reported, and the flashing fails

Signed-off-by: lyx <yanxia.li@siemens.com>


recipes-bsp/u-boot/README.md
```diff
@@ -59,6 +59,15 @@ apt-get install flashrom
 
 Attach the grabber to the flash chip, then run the following on the host machine:
 
+While the the size of the uboot file doesn't match the flash chip's size (16777216 B)
```
jan-kiszka, line 62: "As the size ..." - or "if", but I think we don't ship or produce 16M images anywhere.

recipes-bsp/u-boot/README.md
```diff
@@ -59,6 +59,15 @@ apt-get install flashrom
...
+execute this command:
+
+```shell
+truncate -s 16M iot2050-image-boot.bin
+```
+
+Then, upgrade  uboot:
```
jan-kiszka, line 69: Strictly spoken, the update will not only be U-Boot but the complete firmware. I would write: "Then update the firmware:"

lyxsiemens: Thanks a lot, I will update the md to be more accurate

recipes-bsp/u-boot/README.md
lyxsiemens: Yes, i got it. I will use "if" instead of "while". Thank you very much.

recipes-bsp/u-boot/README.md
```diff
@@ -59,6 +59,15 @@ apt-get install flashrom
 
 Attach the grabber to the flash chip, then run the following on the host machine:
 
+if the the size of the uboot file doesn't match the flash chip's size (16777216 B)
```
jan-kiszka, line 62: Nit: "If..." (capitalization at the beginning of a sentence) - and maybe write "(16 MB)" here.

And could you squash the second commit into the first one (git rebase...)? Thanks.

lyxsiemens: I have completed the modification, please check.
Thanks.

jan-kiszka (approved):
Small change, many iteration - but now you are truly familiar with the workflow :smiley:

PR #313, opened 2022-05-09
Update SEBoot

BaochengSu:
SEBoot is updated to D01.02.02.04 to fix:
1. The AVS issue
2. Some secure boot issues


jan-kiszka:
Are we (or the trusted world) secured now? A more telling change log (impact of both changes) would be good.

BaochengSu:
The changelog is recorded internally on our internal version control, here is only the signed artifacts.

jan-kiszka:
I know, and that is exactly why I'm asking for a more details, impact oriented change log for our external users. It's time to provide that now, we have customers doing own integrations that require this.

BaochengSu:
The details will be provided together with the secure boot tooling integration to this repo but now this PR.

This PR is only regarding quality issue fixing but not targeting `releasing a feature to customer`.

BaochengSu:
> The details will be provided together with the secure boot tooling integration to this repo but now this PR.
> 
> This PR is only regarding quality issue fixing but not targeting `releasing a feature to customer`.

In other words, this PR is only regarding quality fix that customer does not need to pay attention to.

BaochengSu:
> we have customers doing own integrations that require this.

Or, if I understand correctly, is there now customers already developing their security solution based on secure boot which we have not formally released?

jan-kiszka:
Let's not wait until we forget to write an exhaustive changelog when tagging a release here and write proper changelogs when commiting binaries that depend on us to be commented on.

BaochengSu:
There will not be changelog if there is not even version 1, such changelong related to secure boot will be reasonable only after we formally declared secure boot feature.

jan-kiszka (approved):

PR #317, opened 2022-05-19
Chao/fix partuuid

AsuraZeng:
After lots of testing,the error `waiting for partuuid` can not be resolved

when install image to emmc may show `the primary gpt table corrupt,the backup gpt table appears ok`
in this situation, using `sfdisk --part-uuid` can not get the changed partuuid instead of get the previous original emmc
partuuid not the new one(same as boot media part-uuid)

so in the script regen-rootfs-uuid `sed -i 's/'${OLD_UUID}'/'${NEW_UUID}'/i' /etc/default/u-boot-script` 
replacement fails

finally, it would result in `waiting for partuuid.....`

### DIFF
--- a/recipes-core/install-on-emmc/files/install-on-emmc.sh
+++ b/recipes-core/install-on-emmc/files/install-on-emmc.sh
@@ -69,6 +69,7 @@ blink()
 
 ROOT_DEV="$(findmnt / -o source -n)"
 BOOT_DEV="$(echo "${ROOT_DEV}" | sed 's/p\?[0-9]*$//')"
+ROOT_PART="$(echo "${ROOT_DEV}" | sed 's/.*[^0-9]\([0-9]*\)$/\1/')"
 EMMC_DEV="$(ls /dev/mmcblk*boot0 2>/dev/null | sed 's/boot0//')"
 
 trap terminate 0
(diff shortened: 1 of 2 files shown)


jan-kiszka:
This doesn't address why partprobe is not resolving the race.

But I still believe we should drop this altogther an go for a proper installer image.

AsuraZeng:
> This doesn't address why partprobe is not resolving the race.
> 
> But I still believe we should drop this altogther an go for a proper installer image.

partprobe update the partition table but would not recovery partition table.
so if the partition table corrupt, it would  get unexpected behavior.
 
yes, I agree with you.  currently install-on-emmc would take long time to flash the emmc and
not easy to operate.

But I think we should provide a workaround at least for this issue before we finish the new proper installer image.

jan-kiszka:
OK, but then make this here much simpler by matching and replacing any currently set PARTUUID. There is not need to pass in the old one if you make the sed pattern more flexible.

recipes-core/regen-rootfs-uuid/files/regen-rootfs-uuid.sh
```diff
@@ -16,7 +16,12 @@ if [ "${ROOT_DEV}" = "${BOOT_DEV}" ]; then
 	exit 1
 fi
 
-OLD_UUID=$(sfdisk --part-uuid ${BOOT_DEV} ${ROOT_PART})
+if [ $# -lt 1 ];then
```
chombourger, line 19: `${#}` for consistency with other variables

recipes-core/regen-rootfs-uuid/files/regen-rootfs-uuid.sh
```diff
@@ -16,7 +16,12 @@ if [ "${ROOT_DEV}" = "${BOOT_DEV}" ]; then
...
+	OLD_UUID=$(sfdisk --part-uuid ${BOOT_DEV} ${ROOT_PART})
+else
+	OLD_UUID="$1"
```
chombourger, line 22: `${1}`

recipes-core/install-on-emmc/files/install-on-emmc.sh
```diff
@@ -137,13 +138,16 @@ if ! test -b ${EMMC_DEV}p1; then
...
 fi
+
+BOOT_DEV_UUID=$(sfdisk --part-uuid ${BOOT_DEV} ${ROOT_PART})
+
 mount ${EMMC_DEV}p1 /mnt
 mount -o bind /dev /mnt/dev
 mount -t proc proc /mnt/proc
-chroot /mnt sh -c ' \
-    /usr/share/regen-rootfs-uuid/regen-rootfs-uuid.sh
+chroot /mnt sh -c " \
```
chombourger, line 147: if I am not mistaken there are no error checks throughout this code
I worry that we could end-up with a corrupted install

jan-kiszka: Yup, it could - and it did already in some cases. But that is a pre-existing issue.

AsuraZeng:
alternate similar solution has merged into the master. close it

PR #318, opened 2022-05-23
regen-rootfs-uuid: Remove dependency on reading correct old UUID

jan-kiszka:
This both simplifies the logic and makes it robust against temporary
inconsistency after writing the rootfs via install-on-emmc.

As an alternative to #317.


AsuraZeng:
it looks more concise.

AsuraZeng (approved):

PR #322, opened 2022-05-31
kernel: mdio: pick up fix from upstream

AsuraZeng:
pick up commit from upstream to avoid potential ethernet problem.


jan-kiszka (changes requested):
This patch is not in stable 5.10. Either it is a real fix, then it should (also) be proposed for stable, or it is not - then it should not be merged. Reading your commit message above, I cannot judge yet if it fixes a real problem or not. You do not describe what issue you saw, how to that was reproduced and if the patch has helped to resolve it.

Issue #315, opened 2022-05-13
Reading digital input with node-red-intel-gpio


wjomo:
Hello,
I am using a IOT2050 with the Example Image V1.2.2. On this system I am using node red and the Siemens IO Shield for reading some digital inputs. The digital input node reads the status of the signal, if its changing the node will send a message. Thank works fine.
I have checked the option **Send initial message with level of pin**, because I would like to read the digital input when the box is starting or node red is restarting. But at startup there is no message sent from the node, only when I deploy my flow there is a message sent. I thinks it is very important to read the status at startup or restart the flow, otherwise there will be a undefined status.

I am thinking about two useful features for that node:

1. Send the status of the pin at startup or restart the flow.
2. Reading the input periodically with a scaleable cycle-time.

But I have no idea if that features are possible to implement.

I would be happy if that were possible.

Thanks for your help an have an nice day.

fr0st61te:
1. Once you'll deploy the flow, you'll see that will be started at startup or after Node-Red restart.
2. Probably you need 'inject' functions from Node-Red nodes, as example https://theautomationplace.com/2019/10/30/time-based-automations-with-node-red/

wjomo:
Hi,
thanks for your answer.
I am using the pre installed nodes on the Example Image V1.2.2 for reading the digital inputs with node red.
At my use-case I tested the behaviour of the digital input nodes with checking the checkbox **Send initial message with level of pin** and without checking these option. With checking that option the node only send an status of the pin when I deploy the flow. At startup or restart Node-Red nothing happens.

I can t work with inject nodes, becuase there is no possibility to connect the digital input node with an inject node.

It would be great, if there is a way to adjust a time for a cyclic reading the input as it works with the analog input node. There is an option for adjusting the cycle time for reading the value.

I don t know what could I do to solve that problem.

Thanks for your help.

wjomo:
@fr0st61te 
Hi Ivan, 
do you have some further information for me about my issue?

Thank you an have an nice day.

fr0st61te:
> @fr0st61te Hi Ivan, do you have some further information for me about my issue?

@wjomo there is issue as you stated about messages on startup from DIN node, trying to figure out how to solve it.

wjomo:
Hi Ivan,
thanks for your quick answer.
If I understand correctly, you are trying to find a solution for that problem? Please let me know, if I can support you.

Thanks

fr0st61te:
[...]

PR #323, opened 2022-06-02
node-red-gpio: DIN's behavior change

fr0st61te:
Send explicitly status of pin when initialmsg checkbutton is on.
Eventhandler call will not send any DIN's status, it will send only on
edge trigger which doesn't happen on start.

Add the timer for sending out status because some nodes may be not ready
to get message from send call, same ideology using inject node, there is
no state of node from which it can be operated.

[f374dfb](https://github.com/fr0st61te/meta-iot2050/commit/f374dfba3ceba17ab2e1cd431221f5b634397862) didn't fix the situation, it didn't send status on startup for nodes.

### DIFF
--- a/recipes-app/node-red-gpio/files/0006-node-red-gpio-Add-support-for-initial-message-from.patch
+++ b/recipes-app/node-red-gpio/files/0006-node-red-gpio-Add-support-for-initial-message-from.patch
@@ -1,4 +1,4 @@
-From 03f27bff237021159536e5e5ea9e4e5d3421d521 Mon Sep 17 00:00:00 2001
+From 58e98769cdbf92fc4de9721fceb099b2e4129808 Mon Sep 17 00:00:00 2001
 From: Ivan Mikhaylov <ivan.mikhaylov@siemens.com>
 Date: Tue, 2 Nov 2021 17:20:04 +0000
 Subject: [PATCH] node-red-gpio: Add support for initial message from digital
@@ -12,8 +12,8 @@ Based on siemens/meta-iot2000@5fc2bbe patch 0003.
 Signed-off-by: Ivan Mikhaylov <ivan.mikhaylov@siemens.com>
 ---
  hardware/intel/mraa-gpio-din.html |  8 +++++++-
- hardware/intel/mraa-gpio-din.js   | 20 ++++++++++++++++----
- 2 files changed, 23 insertions(+), 5 deletions(-)
+ hardware/intel/mraa-gpio-din.js   | 24 ++++++++++++++++++++----
+ 2 files changed, 27 insertions(+), 5 deletions(-)
 
 diff --git a/hardware/intel/mraa-gpio-din.html b/hardware/intel/mraa-gpio-din.html
 index 58a1363..1c2903d 100644
@@ -42,16 +42,17 @@ index 58a1363..1c2903d 100644
          <label for="node-input-name"><i class="fa fa-tag"></i> Name</label>
          <input type="text" id="node-input-name" placeholder="Name" style="width: 250px;">
 diff --git a/hardware/intel/mraa-gpio-din.js b/hardware/intel/mraa-gpio-din.js
-index 9fc3a12..a8f6a83 100644
+index 9fc3a12..cfb7961 100644
 --- a/hardware/intel/mraa-gpio-din.js
 +++ b/hardware/intel/mraa-gpio-din.js
-@@ -8,13 +8,14 @@ module.exports = function(RED) {
+@@ -8,13 +8,15 @@ module.exports = function(RED) {
          this.pin = n.pin;
          this.interrupt = n.interrupt;
          this.mode = n.mode;
 +        this.initialMsg = n.initial;
          this.x = new m.Gpio(parseInt(this.pin));
          this.board = m.getPlatformName();
++        this.defaultTimeout = 100;
          var node = this;
          node.x.mode(parseInt(this.mode));
          node.x.dir(m.DIR_IN);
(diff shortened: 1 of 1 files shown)


recipes-app/node-red-gpio/files/0006-node-red-gpio-add-input-for-DIN-node.patch
```diff
@@ -0,0 +1,48 @@
+From 42cfd1e45afe8a092161540e365799fe83e26d57 Mon Sep 17 00:00:00 2001
+From: Ivan Mikhaylov <ivan.mikhaylov@siemens.com>
+Date: Tue, 31 May 2022 18:19:04 +0000
+Subject: [PATCH] node-red-gpio: add input for DIN node
+
+Add improvement for DIN node which adds a possibility to get input
+connection with different nodes, as example, inject node which may
+control when DIN node reports about current status.
```
jan-kiszka, line 8: Is this pattern found elsewhere as well? On first sight, it does not look very intuitive to me.

recipes-app/node-red-gpio/node-red-gpio_0.0.6-IOT2050.bb
```diff
@@ -20,8 +20,7 @@ SRC_URI = " \
     file://0002-extend-gpio-to-D19.patch \
     file://0003-Clean-up-mraa-objects-on-node-closing.patch \
     file://0004-Add-USER-button.patch \
-    file://0005-Add-configurable-GPIO-mode-for-DIN.patch \
-    file://0006-node-red-gpio-Add-support-for-initial-message-from.patch"
+    file://0005-Add-configurable-GPIO-mode-for-DIN.patch"
```
jan-kiszka: Can you provide an explanation why the pattern no longer works here? Did something in Node Red change since the version that the IOT2000 is using (where this definitely worked)?

fr0st61te: First of all - I didn't test it well, I've looked through physical behavior of this thing which is working except that it didn't send anything on startup from node-red pov, it just calls the handler which just send on GPIO behavior change from falling to raising and from raising to falling. So, I've missed that point with node.send.

recipes-app/node-red-gpio/files/0006-node-red-gpio-add-input-for-DIN-node.patch
fr0st61te: It's common pattern for node-red, as I see from their guides how they trigger functions/other things via timer interval or on startup. Plus of this thing, that DIN shouldn't care about timers or intervals or even about startup checkbuttons, inject node care about such things, anyways, even without connected input node it should work as was before.

jan-kiszka: Looking at https://github.com/node-red/node-red-nodes/blob/master/hardware/BBB/145-BBB-hardware.js#L228, it still uses that on-startup pattern, and I'm not seeing any inputs for DIN nodes. Is that node broken as well?

fr0st61te: No, it is ok. As I mentioned in [for node-re](https://github.com/siemens/meta-iot2050/pull/323#discussion_r888875710) , in that port I didn't add that send which should send it further, it only triggers on GPIO behavior change which doesn't happen on start.

recipes-app/node-red-gpio/node-red-gpio_0.0.6-IOT2050.bb
fr0st61te: > Ok, but then you do not need to add an extra trigger input to the DIN node, only fix the "send once on startup" path.

I thought it first too but then you have to control some needs, like sending status over time or on startup and this possible to make it with inject node which provides exact this stuff related to timings without any enhancements to GPIO and it still be possible to use without input.

jan-kiszka: You only need the additional on-startup trigger, and that could be internal. No need for a timer, all inputs should have edge triggers which will send out new messages automatically.

fr0st61te: > You only need the additional on-startup trigger, and that could be internal. No need for a timer, all inputs should have edge triggers which will send out new messages automatically.

But it the same as for inject nodes which can provide such functionality and even bigger, you can control status reporting via this node which provide you more than just trigger on startup.

jan-kiszka: I'd still like to see some concrete GPIO node which does the same. I would not try to attach an input to something that conceptually only delivers outputs.

fr0st61te: > I'd still like to see some concrete GPIO node which does the same. I would not try to attach an input to something that conceptually only delivers outputs.

Ok, I'll just fix that port then with sending out on startup but what with proposition about cycle-timers from #315 ?

jan-kiszka: Why do we need a timer? Is any of the input not sending events on falling or rising edges?

fr0st61te: > Why do we need a timer?

As example, if client wants to report about status somewhere else by timer or event at startup both of these things could be provided by inject node. So, it's like role distribution between nodes, inject nodes provides any time triggers(interval, once on startup, etc.), gpio node provides edge triggers and doesn't have any other internal logic about time, inject node provides that stuff.

> Is any of the input not sending events on falling or rising edges?

Yes, it should send on edge event.

fr0st61te:
not relevant due to upstream process

PR #325, opened 2022-06-09
Improved README.md and CONTRIBUTING.md

robang74:
Just a tip to add in the README.md, which might be useful for those are not used to use docker/containers. Without this suggestion, the user at its first try will receive a permission error. Trying to run the script as root, it will be said to do not do that.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -24,4 +24,44 @@ We use the
 and review all changes we merge to master.
 
 We appreciate any contributions, so please use the [Forking Workflow](https://www.atlassian.com/git/tutorials/comparing-workflows/forking-workflow)
-and send us `Merge Requests`
+and send us `Merge Requests`. 
+
+## Developer's Certificate of Origin 1.1
+
+In order to pass the checks, commit with signoff option
+
+```shell
+git commit --signoff
+```
+
+When signing-off a patch for this project like this
+
+    Signed-off-by: Random J Developer <random@developer.example.org>
+
+using your real name (no pseudonyms or anonymous contributions), you declare the
+following:
+
+    By making a contribution to this project, I certify that:
+
+        (a) The contribution was created in whole or in part by me and I
+            have the right to submit it under the open source license
+            indicated in the file; or
+
+        (b) The contribution is based upon previous work that, to the best
+            of my knowledge, is covered under an appropriate open source
+            license and I have the right under that license to submit that
+            work with modifications, whether created in whole or in part
+            by me, under the same open source license (unless I am
+            permitted to submit under a different license), as indicated
+            in the file; or
+
+        (c) The contribution was provided directly to me by some other
+            person who certified (a), (b) or (c) and I have not modified
+            it.
+
+        (d) I understand and agree that this project and the contribution
+            are public and that a record of the contribution (including all
+            personal information I submit with it, including my sign-off) is
+            maintained indefinitely and may be redistributed consistent with
+            this project or the open source license(s) involved.
+
--- a/README.md
+++ b/README.md
@@ -13,6 +13,14 @@ For example under Debian Linux
 sudo apt install docker.io
 ```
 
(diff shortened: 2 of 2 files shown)


jan-kiszka:
Please have a look at https://github.com/siemens/meta-iot2050/blob/master/CONTRIBUTING.md to resolve the DCO error. Change makes sense to me otherwise.

robang74:
Now commit [e0b8675](https://github.com/siemens/meta-iot2050/pull/325/commits/e0b8675817cadbaf382b40bc87526a6e386b34e1)  has been signed-off properly.

jan-kiszka:
Please rebase over master and fold into a single commit again. You can force-push the result here. We can also discuss the steps later if you are not familiar with git.

robang74:
Thanks, Jan. I did.

CONTRIBUTING.md
```diff
@@ -24,4 +24,11 @@ We use the
 and review all changes we merge to master.
 
 We appreciate any contributions, so please use the [Forking Workflow](https://www.atlassian.com/git/tutorials/comparing-workflows/forking-workflow)
-and send us `Merge Requests`
+and send us `Merge Requests`. 
+
+In order to pass the checks, commit with signoff option
+
+```shell
+git commit --signoff
```
jan-kiszka, line 32: Hmpf, now I realized that this repo is not yet documenting what we described e.g. here about the signed-off:

https://github.com/siemens/efibootguard/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11

Should be added as well.

CONTRIBUTING.md
```diff
@@ -24,4 +24,11 @@ We use the
...
+```
+
```
jan-kiszka, line 34: extra new-line at EOF - not needed.

CONTRIBUTING.md
robang74: Added also the Developer's Certificate of Origin 1.1

README.md
```diff
@@ -13,6 +13,12 @@ For example under Debian Linux
 sudo apt install docker.io
 ```
 
+If you want to run docker as non-root user then you need to add your user to the docker group:
+
+```shell
+sudo usermod -aG docker $USER
+```
+
```
BaochengSu, line 21: Then a logout is required or `newgrp`?

robang74: > Then a logout is required or `newgrp`?

That are details left to the user. Who does the action, should behave accordingly: if the group docker does not exist probably the system should be configured in such a way docker service run with its own group. If the $USER is not into docker group then a logout/login should be performed. Otherwise, nothing.

BaochengSu: Thanks for the explanation. :)

I was just thinking maybe it would be more user friendly to add some hints here, for example: `You may need to re-login or issue newgrp...`, something like that.

robang74: @BaochengSu: I did it with commit 428cd9d

CONTRIBUTING.md
```diff
@@ -24,4 +24,44 @@ We use the
...
+        (c) The contribution was provided directly to me by some other
+            person who certified (a), (b) or (c) and I have not modified
+            it.
+
+        (d) I understand and agree that this project and the contribution
+            are public and that a record of the contribution (including all
+            personal information I submit with it, including my sign-off) is
+            maintained indefinitely and may be redistributed consistent with
+            this project or the open source license(s) involved.
```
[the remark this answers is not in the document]
jan-kiszka: We do not have a template for that yet, also because some project were pushed toward CLA while other (like ours) are luckily fine with DCO. So these lines were basically taken from EFI Boot Guard, and that took them from Jailhouse once.

BaochengSu:
@robang74 I think everything else LGTM, except as Jan mentioned, please rebase all the commits to a single one.

After that I think this PR could be approved & merged.

robang74:
@BaochengSu: all changes has been merged into a single commit 39c07ef as requested.

jan-kiszka (approved):

PR #326, opened 2022-06-09
optee-os: add optee rpmb support

fr0st61te:
This patch provides rpmb support for IOT2050 on am65 which includes:
* TISCI protocol support
* DKEK retrieval
* DDR setup for DYNSHM
* compile option changes for optee

Signed-off-by: Ivan Mikhaylov <ivan.mikhaylov@siemens.com>


recipes-bsp/optee-os/optee-os-iot2050_3.16.0.bb
```diff
@@ -22,7 +28,8 @@ OPTEE_NAME = "iot2050"
 OPTEE_PLATFORM = "k3-am65x"
 OPTEE_EXTRA_BUILDARGS = " \
     CFG_ARM64_core=y CFG_TEE_CORE_LOG_LEVEL=2 CFG_USER_TA_TARGETS=ta_arm64 \
-    CFG_CONSOLE_UART=1"
+    CFG_CONSOLE_UART=1 CFG_RPMB_FS=y CFG_RPMB_FS_DEV_ID=1 CFG_CORE_DYN_SHM=y \
```
fr0st61te, line 31: What about CFG_RPMB_WRITE_KEY in which case it should be added?

recipes-bsp/optee-os/files/0001-plat-k3-drivers-Add-secure-proxy-driver-for-communic.patch
```diff
@@ -0,0 +1,460 @@
+From cf20f0a4f77eddf47f278f3d20dc786793542ab4 Mon Sep 17 00:00:00 2001
```
fr0st61te, line 1: I didn't test it yet just bring it up for discussion.

recipes-bsp/optee-os/optee-os-iot2050_3.16.0.bb
jan-kiszka: https://github.com/OP-TEE/optee_os/issues/5338
"CFG_RPMB_WRITE_KEY=y is very dangerous to use a on a production platform."

But let's try to pull in @apalos here as we left our private discussion with the question whether it takes a special firmware version (or rather a special op-tee build) to do the pairing with RPMB. Ilias, can you comment on this?

Looks like we want this `WRITE_KEY` switch to be build-time configurable and tell our users to never ship (or even generate) a _signed_ build which has this switch enabled. Pairing could also be done in a safe environment before enabling secure boot, but it does take a special firmware in my understanding.

BaochengSu: > Pairing could also be done in a safe environment before enabling secure boot, but it does take a special firmware in my understanding.

Then when the device is leaving the safe environment without secure boot enabled, attacker could burn any OPTee to steal the RPMB key.

BaochengSu: Or rather any software running at S-EL1. Anyway device need regenerate the RPMB key in secure memory with OPTee code, hacker could reverse engineering on the shipped S-EL1 (normally OPTee) code to get the key-generating logic and replay in their own S-EL1 software to dump the RPMB key.

I think most of the users will use the upstream OPTee code directly, which make it more easier, the key regenerating logic is already there in source code format.

BaochengSu: My point is when the device is leaving the safe environment for a non-safe circumstance, the secure boot must be enabled.

BaochengSu: > tell our users to never ship (or even generate) a _signed_ build which has this switch enabled.

What could happen if this *signed* build is leaked?

recipes-bsp/optee-os/files/0001-plat-k3-drivers-Add-secure-proxy-driver-for-communic.patch
BaochengSu: Didn't checked it but I notice this patch already acked by Linaro, so I guess this is already upstreamed or under upstreaming... So I believe somebody else had already tested it...

I think this patch is the prerequisite for the next SCI api patch so it must be picked.

jan-kiszka: Sure, if we can pick the patches via upstream, that would be better. But the question is also if there is already a release with the patch included...

recipes-bsp/optee-os/optee-os-iot2050_3.16.0.bb
jan-kiszka: That is exactly what we should understand better. It could be "very dangerous", see above.

BaochengSu: OK, I thought this is the real risk:

```
+OP-TEE can program this key when built with ``CFG_RPMB_WRITE_KEY=y``. One
+must be well warned that this configuration makes OP-TEE to send the RPMB key
+in clear to the non-secure side that relays the RPMB key programming request to
+the eMMC hardware device.
```

From https://github.com/OP-TEE/optee_docs/pull/157

recipes-bsp/optee-os/files/0001-plat-k3-drivers-Add-secure-proxy-driver-for-communic.patch
BaochengSu: I am ok with a OPTee version bump as long as it does not generate clearing delta.

jan-kiszka: More important would be a careful check if security fixes need to be picked up - we are starting to put OP-TEE on the critical path now for real.

BaochengSu: Agree. We have CERT report to refer to, maybe that could be helpful.

fr0st61te: > Sure, if we can pick the patches via upstream, that would be better. But the question is also if there is already a release with the patch included...

there is no release, it's not under any tags, it's just directly from master which could be to dangerous to bump into that state.

fr0st61te: > Didn't checked it but I notice this patch already acked by Linaro, so I guess this is already upstreamed or under upstreaming... So I believe somebody else had already tested it...
> 
> I think this patch is the prerequisite for the following SCI api patch so it must be picked.

Yes, TI and me both did testing but not for 3.16.0, I'll test it out to check if all works from upstream.

recipes-bsp/optee-os/optee-os-iot2050_3.16.0.bb
apalos: @BaochengSu usually this takes place in production once, so although it will happen in clear text it should at least be in a controlled safe environment. 

The main cause of problems here is how OP-TEE generates the key to begin with.  If the platform does not provide it's own key (ideally per hardware), OP-TEE will end up programming the default key.  However the default key is known and should only be used for testing.  

Unless your platform has it's own implementation of `tee_otp_get_hw_unique_key` it's probably using the default OP-TEE key

apalos: Yes.  I'll go update the op-tee documentation but basically you got two threats. The first is what I mentioned above.  The second one is related to how OPP-TEE accesses the eMMC to begin with.

OP-TEE uses a supplicant to communicate with the non secure world eMMC driver (since the hardware resides in non-secure world to begin with). If someone replaces the supplicant with a rogue application always replying "rpmb key not programmed", OP-TEE will then send the key in clear text,  asking for the supplicant to program it.

recipes-bsp/optee-os/optee-os-iot2050_3.16.0_rpmbw.bb
```diff
@@ -0,0 +1,14 @@
...
+# Authors:
+#  Ivan Mikhaylov <ivan.mikhaylov@siemens.com>
+#
+# This file is subject to the terms and conditions of the MIT License.  See
+# COPYING.MIT file in the top-level directory.
+#
+
+require optee-os-iot2050_3.16.0.inc
+
+OPTEE_EXTRA_BUILDARGS += " CFG_RPMB_WRITE_KEY=y"
```
jan-kiszka, line 13: Let's control that via an `OVERRIDE`, something like `OPTEE_EXTRA_BUILDARGS_append_rpmb-setup = '...'`. Then you can fold both recipes together again. The override would be set by a kas option file, similar to the secure boot enabling.

recipes-bsp/optee-os/optee-os-iot2050_3.16.0_rpmbw.bb
```diff
@@ -0,0 +1,14 @@
...
+DEFAULT_PREFERENCE = "-1"
```
jan-kiszka, line 14: Where is this used?

jan-kiszka:
We also need that kas option file to enable the special firmware build. And that setup should exclude signing the firmware, ie. detect and reject secure boot mode.

recipes-bsp/optee-os/optee-os-iot2050_3.16.0_rpmbw.bb
fr0st61te: It is for default preference for optee-os-iot2050_3.16.0.bb, it possible to use PREFERRED_PROVIDER instead of this option.

recipes-bsp/optee-os/optee-os-iot2050_3.16.0_rpmbw.bb
fr0st61te: ok, I'll check how it will goes with OVERRIDE.

recipes-bsp/optee-os/optee-os-iot2050_3.16.0_rpmbw.bb
jan-kiszka: This is first of all a bitbake variable, and at that level no one is using it. If it has a meaning for OP-TEE, you need to bring it into its (Debian) build environment. That maybe require a Debian build profile here to be future-proof.

fr0st61te:
> We also need that kas option file to enable the special firmware build.

kas-iot2050-boot-pg1-optee-rpmbw.yml
kas-iot2050-boot-pg2-optee-rpmbw.yml ? Or something else?

jan-kiszka:
Do we need pg1/2 differentiation for that? Officially supported will be pg2 only anyway, but we can leave that door technically open. Otherwise maybe `kas/opt/rpbm-setup.yml`?

Kconfig
```diff
@@ -125,11 +125,24 @@ config SECURE_BOOT
 	  This uses the public custMpk.key and certificate by default. Do not
 	  use this key in production, it is for demonstration purposes only.
 
+config RPMB_SETUP
+	bool "OPTEE RPMB setup for OTP key write"
+	depends on IMAGE_BOOT_PG2 && !SECURE_BOOT
+	help
+	   Enable CFG_RPMB_WRITE_KEY=y for OPTEE which provides
+	   OTP(one-time-programmable) RPMB key write into EMMC for initial
+	   HUK/DKEK setup for board.
```
fr0st61te, line 134: Do we need to put a warning here?

jan-kiszka: Maybe rather like this: "Enable one-time pairing between processor and secure storage (RPMB on eMMC). Do not sign and distribute this version. Run it only in a secure environmenton the target device, then replace it with signed production firmware and enable secure boot." Less technical as our audience may not be familiar with all the details you mentioned.

jan-kiszka:
What is now the actual testing status of these OP-TEE patches in the backported environment (3.16.0, rather than master)? Can we trust that? Or should we move at least to 3.17.0 (upstream merged them not that many commits after the release)?

fr0st61te:
> What is now the actual testing status of these OP-TEE patches in the backported environment (3.16.0, rather than master)? Can we trust that? Or should we move at least to 3.17.0 (upstream merged them not that many commits after the release)?

3.16.0 works fine with backported patches. 3.17.0 and 3.17.0-rc1 don't have these patches, probably in new rc or .18 it will be available.

fr0st61te:
Last version of commits is not tested with 3.18.0, will do it on the current week.

recipes-bsp/optee-os/optee-os-iot2050_3.18.0.bb
```diff
@@ -22,7 +22,18 @@ OPTEE_NAME = "iot2050"
...
+
+OPTEE_EXTRA_BUILDARGS_append_rpmb-setup = " CFG_RPMB_WRITE_KEY=y"
+
+python do_patch_prepend() {
+    import re
+
+    overrides = d.getVar('OVERRIDES')
+    if re.search("rpmb-setup", overrides) and re.search("secureboot", overrides):
+        bb.fatal("Not possible to use Secure Boot and RPMB setup for OPTEE")
+}
```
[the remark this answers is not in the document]
jan-kiszka: I think that check already exists there. This is to prevent user combining that on the command line without using kconfig.

kas/opt/rpbm-setup.yml
```diff
@@ -0,0 +1,16 @@
...
+# This file is subject to the terms and conditions of the MIT License.  See
+# COPYING.MIT file in the top-level directory.
+#
+
+header:
+  version: 10
+
+local_conf_header:
+  optee-os-iot2050_override: |
+    OVERRIDES .= ":rpmb-setup"
```
BaochengSu, line 16: `s/rpbm-setup.yml/rpmb-setup.yml/`

BaochengSu:
Testing on PG2, below error print could be confusing.

```
Security ID 0xc8beb00d-0x7f55c02e
Security policy: none
No options found, skip.
Validating FIT...
Loading image atf...
Loading image tee...
Loading image spl...
Loading image k3-am65-iot2050-spl.dtb...
NOTICE:  BL31: v2.6(release):
NOTICE:  BL31: Built : 02:56:00, Sep 14 2022
I/TC:
I/TC: OP-TEE version: 3.18.0 (gcc version 10.2.1 20210110 (Debian 10.2.1-6)) #1 Mon Sep 26 02:11:06 UTC 2022 aarch64
I/TC: WARNING: This OP-TEE configuration might be insecure!
I/TC: WARNING: Please check https://optee.readthedocs.io/en/latest/architecture/porting_guidelines.html
I/TC: Primary CPU initializing
I/TC: SYSFW ABI: 3.1 (firmware rev 0x0015 '21.9.1--v2021.09a (Terrific Lla')
I/TC: HUK Initialized
I/TC: Activated SA2UL device
E/TC:0 0 ti_sci_get_response:120 Message not acknowledged
E/TC:0 0 ti_sci_do_xfer:149 Failed to get response (-65536)
I/TC: Enabled firewalls for SA2UL TRNG device
I/TC: SA2UL TRNG initialized
I/TC: SA2UL Drivers initialized
I/TC: Primary CPU switching to normal world boot
```

This happens during Line 64 of `core/arch/arm/plat-k3/drivers/sa2ul.c`
```c
	/* Try to claim the SA2UL firewall for ourselves */
	ret = ti_sci_change_fwl_owner(fwl_id, sa2ul_region, owner_index,
				      &owner_privid, &owner_permission_bits);
	if (ret) {
		/*
		 * This is not fatal, it just means we are on an HS device
		 * where the DMSC already owns the SA2UL. On GP we need
		 * to do additional setup for access permissions below.
		 */
		DMSG("Could not change SA2UL firewall owner");
	} 
```

So it is expected on HS device. However, it is still confusing if I am a end user. And strictly speaking, the above code would cover some error if something wrong happens within `ti_sci_change_fwl_owner` on HS device.

BaochengSu:
Seems `CONFIG_CMD_OPTEE_RPMB` is missing from u-boot?

BaochengSu:
I think we should have some wikis regarding how to program the RPMB key, just as what we've done for secure boot otp key programming.

BaochengSu:
Test passed on PG2 device.

BaochengSu (approved):

PR #327, opened 2022-06-13
classes/npm: let gyp use python3

chombourger:
While gyp needs /usr/bin/python, it will happily work with python3
(removing the need to to pull python2): use Debian's package
python-is-python3 instead of python

Signed-off-by: Cedric Hombourger <cedric.hombourger@siemens.com>


jan-kiszka (approved):

BaochengSu:
@chombourger , this commit is only let gyp use python3 instead of python2, my understanding is that it does not impact the functionalities?

jan-kiszka:
CI passed, so none of our self-installed npm packages has problems with that. Not really a surprise, given that python2 is dead, and we do not have old versions of the packages in use.

jan-kiszka (approved):

PR #329, opened 2022-06-29
Secure Boot related refine

BaochengSu:
- Re-organize the secure boot keys
- Add OTP provisioning to menu
- Warn usage of dummy key


jan-kiszka (commented):
file://dir is fine with links inside dir. file://dir/* had problems, no? I don't see the need for explicit key listing yet.

Kconfig
```diff
@@ -181,3 +203,53 @@ config FIRMWARE_SECURE_VER
 	help
 	  Use specific anti-rollback secure version rather than the default 0.
 	  Range 0 - 127.
+
+choice
+	prompt "*** OTP provisioning command type ***"
```
jan-kiszka, line 208: This `***` separator does not look very beautiful IMHO.

BaochengSu: image

BaochengSu: image

jan-kiszka: Ah, was me in fact :smile:

jan-kiszka:
You need to describe better the boundary conditions when this starts to break. Currently, it works. I suppose it fails when the links we have in keys/ are pointing outside of the keys/ folder, right?

BaochengSu:
> You need to describe better the boundary conditions when this starts to break. Currently, it works. I suppose it fails when the links we have in keys/ are pointing outside of the keys/ folder, right?

Yep.

jan-kiszka (approved):

PR #332, opened 2022-07-08
u-boot-iot2050: Fix the PE image validation

BaochengSu:
1. Backport the upstream fix of the PE image validation issue of UEFI secure
boot:

https://patchwork.ozlabs.org/project/uboot/cover/20220705054815.30318-1-
takahiro.akashi@linaro.org/

2. Upgrade efibootguard to v0.12 to pick up the `bg_gen_unified_kernel` change to drop the symbol  table from the PE Image so that u-Boot can calculate the correct hash value.

3. the fix to turn on the RED stat LED once boot failed in secure boot mode.

### DIFF
--- a/isar-patches/0001-efibootguard-Update-to-release-0.12.patch
+++ b/isar-patches/0001-efibootguard-Update-to-release-0.12.patch
@@ -1,0 +1,77 @@
+From 4e7ea3e468760b19ab3a48faf8acae1a0849a41c Mon Sep 17 00:00:00 2001
+From: Jan Kiszka <jan.kiszka@siemens.com>
+Date: Mon, 11 Jul 2022 15:07:15 +0200
+Subject: [PATCH] efibootguard: Update to release 0.12
+
+Allows to drop our patch. And it's the precondition to add ARM support.
+
+Signed-off-by: Jan Kiszka <jan.kiszka@siemens.com>
+---
+ ...bootguard_0.11.bb => efibootguard_0.12.bb} |  5 ++--
+ ...efile-Drop-nostdinc-for-EFI-binaries.patch | 28 -------------------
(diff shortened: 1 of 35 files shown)


recipes-bsp/u-boot/files/0032-efi_loader-PE-hashing-for-image-with-gap-between-sec.patch
```diff
@@ -0,0 +1,56 @@
...
+ 			  len - (bytes_hashed + authsz));
+ 		efi_image_region_add(regs, efi + bytes_hashed,
+-				     efi + len - authsz, 0);
++				     efi + len - authsz, 1);
+ 	}
+ 
+ 	/* Return Certificates Table */
+-- 
+2.30.2
+
```
jan-kiszka, line 56: I'm still not confident about this one, and we resolved the issue in EFI Boot Guard by avoiding the gap. So this should not be included, and it may also not go into upstream. Let's rather backport https://github.com/siemens/efibootguard/commit/e9d982336bd5cc651b45f5fddc6e9f47568e78d1 for now until we can update isar-cip-core.

BaochengSu: By patch cip-core, or adding new `recipes-bsp/efibootguard/efibootguard-iot2050_0.11.bb`?

jan-kiszka: I fact, we should consider moving to the upcoming 0.12 release of EBG as that will fix other issues.

jan-kiszka (approved):

PR #335, opened 2022-07-17
iot2050-image-example: Add systemd-timesyncd

AsuraZeng:
Currently,there is no ntp service by default,so can not
sync time automatic.

Using systemd-timesyncd daemon to do the time sync.

### DIFF
--- a/recipes-core/images/iot2050-image-example.bb
+++ b/recipes-core/images/iot2050-image-example.bb
@@ -61,6 +61,7 @@ IOT2050_DEBIAN_DEBUG_PACKAGES = " \
     i2c-tools \
     sudo \
     usb-modeswitch \
+    systemd-timesyncd \
     "
 
 # wifi support


jan-kiszka:
I could swear that worked already - otherwise `apt update` shouldn't work as well (broken clock generally means invalid apt database). But I do not have a box and image in reach to confirm.

jan-kiszka (approved):

AsuraZeng:
it should work, but my several board need to use the `date` modify time every time, otherwise apt update can't be successful.

BaochengSu:
> I could swear that worked already - otherwise `apt update` shouldn't work as well (broken clock generally means invalid apt database).

+1

AsuraZeng:
I have checked, the latest version don't have the package `systemd-timesyncd`.
the release image 1.2.2 has this package by default. 

perhaps there are some changes on the debian package repo.

jan-kiszka:
Unlikely as we did not update the major Debian version in between. Unless we happened to pull the package indirectly before, this remains a mystery for now.

BaochengSu:
Checked with @AsuraZeng , this does not impact the upcoming release since the modification had not happened in the package freezing snapshot.

So my suggestion is to merge it after the upcoming release.

BaochengSu (approved):

jan-kiszka:
Good catch - and ack for postponing this.

PR #336, opened 2022-07-19
Update SEBoot to D01.02.02.07

BaochengSu:
This fixes the commit ca571f028abb682bb4bc25cc80d51ad668964090

The issue is that the firewall configuration on PG1 device cause the dma
transferring of SPI flash not working. Since we don't have secure boot
on PG1 device, there is no need to configure the firewall on these
devices.


AsuraZeng (approved):
LGTM

jan-kiszka (approved):

PR #343, opened 2022-07-29
install-on-emmc: Improve the robustness

BaochengSu:
Main change point:

1. When using Win32 Disk Imager to make the boot-able SD card or USB
   stick that bigger than the eMMC size (16GB), the `Backup LBA` of the
   GPT header is set to the end of the device, which is beyond the 16GB
   range. After copying such image to the eMMC, linux does not treat the
   GPT partition table as a correct one and some wired problems happen,
   such as the new p1 could not be informed to kernel.

   The solution is, the `Backup LBA` must be adjusted to the end of the
   eMMC after copying the image to the eMMC. `gdisk` provides an easy
   way to handle that.

2. If the eMMC has old partition data, (which is true for the Advanced
   device with Industrial OS), then better to wipe the eMMC partition
   table before copying the data to eMMC. This is because according to
   UEFI spec, there is a GPT backup header and backup partition table at
   the end of the device. When copying the new image via dd, normally
   the tail part of the device will not be touched. This is ok when
   everything goes smoothly, however, if problem happen during the
   install-on-emmc, the old backup partition data could be recovered
   during next boot.

   For example in the above problem, when `Backup LBA` is beyond the
   eMMC size, the new partition p1 could not be informed to kernel.
   however, the since there is already the old p1 before
   install-on-emmc, the script still continue to reboot the device,
   however, during next booting, the `Backup LBA` offset error is
   detected and the old backup partition table is recovered to overwrite
   the new GPT table.

   It is very hard to notice this issue especially when the old image
   and the new image have similar partition layout.

   The solution is, before install-on-emmc, wipe the old GPT
   partitioning data, including the backup LBA. `sgdisk` also provides
   an easy way to wipe it, meanwhile `sfdisk` could not do it easily.

### DIFF
--- a/recipes-core/install-on-emmc/files/install-on-emmc.sh
+++ b/recipes-core/install-on-emmc/files/install-on-emmc.sh
@@ -119,18 +119,23 @@ fi
 
 blink ${LED_GREEN} 0.25
 
+echo "Destroying the old GPT of eMMC ${EMMC_DEV}..."
+sgdisk -Z ${EMMC_DEV}
+
 # Calculate number of sectors to write:
 # Start of boot partition + sectors in that partition
 SECTORS="$(($(sfdisk -d ${BOOT_DEV} 2>/dev/null | tail -1 | sed 's/.*start=[[:space:]]*\([^,]*\), size=[[:space:]]*\([^,]*\).*/\1+\2/')))"
 
 echo "Writing ${SECTORS} sectors to eMMC ${EMMC_DEV}..."
 dd if=${BOOT_DEV} of=${EMMC_DEV} count=${SECTORS} conv=fsync
-partprobe ${EMMC_DEV}
+# fix the possible GPT header problem if BOOT_DEV is bigger than EMMC_DEV
+sgdisk -e ${EMMC_DEV}
 
 echo "Updating partition UUID of eMMC rootfs"
-partx -a ${EMMC_DEV}
-udevadm settle
 if ! test -b ${EMMC_DEV}p1; then
+	partprobe ${EMMC_DEV}
+	partx -a ${EMMC_DEV}
+	udevadm settle
 	echo "Waiting for ${EMMC_DEV}p1 to appear"
 	while ! test -b ${EMMC_DEV}p1; do
 		echo -n "."
--- a/recipes-core/install-on-emmc/install-on-emmc_1.0.bb
+++ b/recipes-core/install-on-emmc/install-on-emmc_1.0.bb
@@ -11,7 +11,7 @@ inherit dpkg-raw
 DESCRIPTION = "This service provides the option to install on eMMC during first boot"
 
 DEPENDS = "regen-rootfs-uuid"
-DEBIAN_DEPENDS = "systemd, fdisk, util-linux, regen-rootfs-uuid, parted"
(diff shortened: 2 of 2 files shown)


BaochengSu:
If this PR is got approved, the target merge point is after the 1.3.1 release, since the new introduced `gdisk` tool.

BaochengSu:
For handling GPT, `gdisk` is doing better than `fdisk`

recipes-core/install-on-emmc/files/install-on-emmc.sh
```diff
@@ -119,18 +119,23 @@ fi
...
-partprobe ${EMMC_DEV}
+# fix the possible GPT header problem if BOOT_DEV is bigger than EMMC_DEV
+sgdisk -e ${EMMC_DEV}
 
 echo "Updating partition UUID of eMMC rootfs"
-partx -a ${EMMC_DEV}
-udevadm settle
 if ! test -b ${EMMC_DEV}p1; then
+	partprobe ${EMMC_DEV}
+	partx -a ${EMMC_DEV}
```
AsuraZeng, line 137: what's the difference partprobe  and partx -a?
seems they are doing similar work

BaochengSu: Not sure, this is the old code, I just move the `partprobe ${EMMC_DEV}` from outside the `if` to inside.

And git blame shows it's you that added this code: [24d8479edb86da8e94dc17c315d3e85748cc04fd](https://github.com/siemens/meta-iot2050/commit/24d8479edb86da8e94dc17c315d3e85748cc04fd)

AsuraZeng: OK, could remove one of this

jan-kiszka:
There more have to tweak on this, there clearer it gets that we rather need an flasher stick instead.

recipes-core/install-on-emmc/files/install-on-emmc.sh
BaochengSu: I am afraid these two were added to improve some stability, so better to not remove any of them. We can think better in the new solution of a flash sticker.

recipes-core/install-on-emmc/files/install-on-emmc.sh
```diff
@@ -119,18 +119,23 @@ fi
...
+	udevadm settle
```
jan-kiszka, line 138: In fact, I would not over-optimize here and leave those steps out of the conditional branch. We don't tell the kernel about partition layout changes otherwise.

BaochengSu: So just delete all three lines here?

BaochengSu: `sgdisk -e` will tell the kernel regarding the layout change anyway.

jan-kiszka: Does it officially? If so, the partprobe and partx calls would be redundant. If not, my comment applies.

BaochengSu:
Let's have a thorough rethinking of this feature first, before that, close this PR.

PR #344, opened 2022-07-29
Chao/fix setup tool

AsuraZeng:


BaochengSu (approved):

recipes-app/iot2050setup/files/iot2050setup.py
```diff
@@ -506,7 +506,7 @@ def configureExternalSerialMode(self):
         self.config['User_configuration']['External_Serial_Current_Mode'] = switchMode
         self.saveConfig(self.config)
         subprocess.call('sync', shell=True)
-        if self.topmenu.boardType == 'IOT2050 Advanced PG2':
+        if not self.topmenu.boardType.startswith('IOT2050 Basic') and self.topmenu.boardType != 'IOT2050 Advanced':
```
jan-kiszka, line 509: Is that negative list now shorter than a positive list?

AsuraZeng: Yes,negative list  shorter than positive list.
For advanced variant, only PG1 don't need to do this operation.

now we have PG2/M2 variant and later we would have another variant.
for long term consider, so using this method to match the devices.

jan-kiszka (approved):

PR #345, opened 2022-07-29
Chao/fix update tool

AsuraZeng:


BaochengSu (approved):

AsuraZeng:
@jan-kiszka hi jan, Can you please provide your comments?

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
```diff
@@ -461,8 +461,12 @@ def main(argv):
     else:
         updater = ManagedFirmwareUpdate(args.firmware)
 
-    envlist = updater.get_preserved_uboot_env(args.preserve_list)
+    envlist = None
+    if not args.force:
+        envlist = updater.get_preserved_uboot_env(args.preserve_list)
```
jan-kiszka, line 464: While should "-f" not preserve the env if all needed information is provided?

AsuraZeng: From my perspective , '-f' would ignore all the checking.so would not keep the env.
It's better to keep the env through our Managed Firmware Update method.

jan-kiszka: Why is that better? This change makes the interface inconsistent.

AsuraZeng: Forced update itself is a higher risk operation. Based on this, providing more interfaces or options, I think the risk will be higher

It is a safer way to keep environment variables in the managed way of updating firmware

The forced update function should be simple

AsuraZeng: Hi jan. 
if we'd like to keep the environment of the target board, so  we'd like to have the builtin environment (u-boot-initial-env) of the target firmware. But currently we plan to integrate this file into our upgrade tar package,this would public to forum. But neither  public u-boot-initial-env  file nor can find the file in our build/tmp/deploy folder.
so for the "-f" option, current what we do is to erase all environment.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
```diff
@@ -461,14 +458,21 @@ def main(argv):
...
     updater.update_firmware()
+
     if envlist:
         print("\nRestore preserved uboot env")
         updater.update_uboot_env(envlist)
 
-    print("\nCompleted. Please reboot the device\n")
-
+    reboot_input = input(
+        "\nWarning: Completed. Please reboot the device. Continue (y/N)? ")
```
BaochengSu, line 472: `Continue` sounds to me too vague. More clear to me is to use `Reboot now (y/N)?`

AsuraZeng: This prompt is better, it has been modified

PR #346, opened 2022-07-30
u-boot: Report fdt error for loading u-boot from SPL

BaochengSu:
u-boot commit 71551055cbdb ("spl: fit: Load devicetree when a Linux
payload is found") made a change to not report the spl_fit_append_fdt
error at all if next-stage image is u-boot.

However for u-boot image without CONFIG_OF_EMBED, the error should be
reported to up-level caller. Otherwise, up-level caller would think the
fdt is already loaded which is obviously not true.

The reported issue is that, when secure boot is enabled, by tampering
the fdt data, although the device hangs, the red LED does not light. The
reason is that when spl parsing the fdt, bad hash value error was not
reported to the up-level caller, which would light the red led if
received any error.


BaochengSu:
Has submitted the patch to upstream.

jan-kiszka (approved):

PR #350, opened 2022-09-09
update mraa and node-red-gpio from upstream

fr0st61te:
Update mraa and node-red-gpio from upstream with latest versions, removing all unneeded patchsets which were ported.


recipes-app/mraa/mraa_2.2.0.bb
```diff
@@ -12,21 +12,8 @@ inherit dpkg
...
-            file://0007-iot2050-Switch-to-runtime-detection-of-gpiochip-numb.patch \
-            file://0008-iot2050-Add-USER-button.patch \
-            file://0009-led-Fix-and-cleanup-initialization.patch \
-            file://0010-uart-Fix-software-flow-control-management.patch \
-            file://0011-gpio-Avoid-spurious-value-reset-without-output-mode-.patch \
-            file://0012-version.h-Fix-multiple-definitions-of-gVERSION-_SHOR.patch \
-            file://0013-Support-for-swig-4.1.0.patch \
             file://rules"
-SRCREV = "7786c7ded5c9ce7773890d0e3dc27632898fc6b1"
+SRCREV = "8b1c54934e80edc2d36abac9d9c96fe1e01cb669"
```
jan-kiszka, line 16: Cool, patch-free. We are are no longer "2.2.0", so please also make that clear in the recipe version.

recipes-app/node-red-gpio/node-red-gpio_0.2.1-IOT2050.bb
```diff
@@ -14,15 +14,9 @@ inherit dpkg-raw
...
-    file://0001-add-the-board-info-add-the-led-control-node.patch \
-    file://0002-extend-gpio-to-D19.patch \
-    file://0003-Clean-up-mraa-objects-on-node-closing.patch \
-    file://0004-Add-USER-button.patch \
-    file://0005-Add-configurable-GPIO-mode-for-DIN.patch \
-    file://0006-node-red-gpio-Add-support-for-initial-message-from.patch"
-SRCREV="3087e8e2a1ea189f394bca0a2af159ad859d7722"
+SRC_URI = "git://github.com/node-red/node-red-nodes;protocol=https"
+
+SRCREV="e4dc3a2bcfd38dfd18eff1bfa5542ff56c40f27c"
```
jan-kiszka, line 19: Also here: recipe version is now no longer "-IOT2050" but actually some upstream version + git.

fr0st61te:
> When will it be possible to (also) pull the nod-red nodes via the palette from npmjs.com? 

I think it's not possible due to difference between mraa ids inside types. It's possible to update to latest version of package in palette manager unfortunately.

> In case users want to update to a custom Node RED version and then install the IO nodes themselves.

I've checked update from 0.0.6 version to 0.2.0, seems works fine for me. It is possible with manual update:
`cd /usr/lib/node_modules/node-red/node_modules/; npm install node-red-node-intel-gpio@0.2.0 -g;`

recipes-app/mraa/mraa_2.2.0.bb
fr0st61te: Ok, but how it should be spelled, like `mraa_master.bb` ?

jan-kiszka: ...or "2.2.0+git", clarify the baseline.

jan-kiszka (approved):

PR #353, opened 2022-09-15
iot2050-firmware-update: Add integrity check

AsuraZeng:
After the update is completed, read back the content of the flash, calculate the hash value of the content and the write value, compare the two hash value to ensure the correctness of the written value.

### DIFF
--- a/recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
+++ b/recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
@@ -112,6 +112,15 @@ import tempfile
 import hashlib
 
 
+class UpgradeError(Exception):
+    def __init__(self, ErrorInfo):
+        super().__init__(self)
+        self.erroinfo = ErrorInfo
+
+    def __str__(self):
+        return self.erroinfo
+
+
 class FirmwareUpdate(object):
     def __init__(self, firmware):
         self.firmware = firmware
@@ -122,8 +131,7 @@ class FirmwareUpdate(object):
             with open(path, "r") as f:
                 return f.read()
         except IOError as e:
-            print("Reading {} failed: {}".format(path, e.strerror))
-            sys.exit(1)
+            raise UpgradeError("Reading {} failed: {}".format(path, e.strerror))
 
     @staticmethod
     def flash_erase(dev, start, nbytes):
@@ -139,8 +147,7 @@ class FirmwareUpdate(object):
         try:
             fcntl.ioctl(dev, MEMERASE, ioctl_data)
         except IOError:
-            print("ioctl failed")
-            sys.exit(1)
+            raise UpgradeError("Flash erasing failed")
 
     def get_mtd_info(self, mtd_num):
         ospi_dev_path = "/sys/bus/platform/devices/47040000.spi"
@@ -170,8 +177,7 @@ class FirmwareUpdate(object):
         try:
             mtd_dev = os.open(mtd_dev_path, os.O_SYNC | os.O_RDWR)
         except IOError as e:
-            print("Opening {} failed: {}".format(mtd_dev_path, e.strerror))
-            sys.exit(1)
+            raise UpgradeError("Opening {} failed: {}".format(mtd_dev_path, e.strerror))
 
         while mtd_pos < mtd_size and file_size > 0:
             mtd_content = os.read(mtd_dev, mtd_erasesize)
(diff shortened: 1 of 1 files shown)


recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
```diff
@@ -218,6 +220,54 @@ class FirmwareUpdate(object):
...
+
+        return firmware_md5.hexdigest()
+
+    def firmware_integrity_check(self):
+        firmware_md5_digest = self.firmware_md5_digest()
+        firmware_flash_md5_digest = self.flash_md5_digest(self.firmware_len)
+
+        if firmware_md5_digest != firmware_flash_md5_digest:
+            raise Exception('Upgrade Failed! Please do not reboot. Please try updating firmware again!\n')
+
```
BaochengSu, line 270: Per the convention used in other place of this script, I would rather use a print and a exit(!0).

A pythonic exception printing seems more like a internal error of the application, at least to me.

Or do we have somewhere in the script to capture this exception?

BaochengSu: By the way, better to inform the user if they reboot the device, it could be bricked.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
```diff
@@ -402,6 +452,21 @@ class ManagedFirmwareUpdate(FirmwareUpdate):
...
+        env_md5 = hashlib.md5()
+        env_md5.update(file_obj.read())
+        env_file_md5_digest = env_md5.hexdigest()
+
+        env_file_size = os.path.getsize(file_obj.name)
+        env_flash_md5_digest = self.flash_md5_digest(env_file_size, mtd_num=3)
+
+        if env_file_md5_digest != env_flash_md5_digest:
+            raise Exception('Upgrade Failed! Please do not reboot. Please try updating firmware again!\n')
+
```
BaochengSu, line 469: Same here as above.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
AsuraZeng: Just printing and then exiting, I'm afraid the user will ignore this message. At least we need obvious error messages

AsuraZeng: For the prompt language, it can be optimized

BaochengSu: > Just printing and then exiting, I'm afraid the user will ignore this message. At least we need obvious error messages

We can use all CAPTION words, or triple `!!!` or a big `WARNING:`.

BaochengSu: Please also pay attention to something like `flash_erase`, such kind of error also makes the device un-bootable.

BaochengSu: After a closer look into the current code, the whole exception logic should be refactored to capture the programming failed event.

Better to throw all the exceptions to the most up level called and let the most upper level to handle them to report the warning.

AsuraZeng: there is something wrong and this error is Fatal Error.
I think the better way is to abort and prompt. Just print is not enough

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
```diff
@@ -475,27 +549,39 @@ def main(argv):
...
+                "\nWarning: Preserve the above env variables. Confirm(Y/n)? ")
+            if not preserved_env_input == "n":
+                updater.update_uboot_env(envlist)
+    except UpgradeError as e:
+        print(e)
+        intput_reminder = '''
+        ====================================
+        Upgrade Failed! Please do not reboot!!
+        The Device may become brick.
+        Please try upgrade firmware again!
```
jan-kiszka, line 579: How likely is it that retrying will resolve? I would also recommend re-downloading.

BaochengSu: Flash programming error is not so rare, if the downloading has no error, then most likely the error is programming, such retrying should helps.

Of course, a re-downloading recommendation should also helps. Or we can suggest the user to calculate the checksum after downloading, before upgrading, in the document.

jan-kiszka: Do we have a checksum as part of the managed update package? If so, my comment is indeed unneeded.

BaochengSu: No, since a broken tar.xz should not be extract-able. Or is it?

However, if we think a checksum is helping then since we have the so-called `managed update`, we can  enhance the `update.conf.json` to have the checksum included. However, I suggest to have it later since the current status is sufficient to detect at least the programming error.

jan-kiszka (approved):

BaochengSu (approved):

PR #355, opened 2022-09-15
Chao/export image

AsuraZeng:


BaochengSu:
I think the final goal for firmware update deploying is that the firmware update package is generated by the build and copied into deploy, not only the .deb for the updating tool.

However, this could be the first step towards the final goal.

BaochengSu (approved):

recipes-bsp/u-boot/u-boot-iot2050.inc
```diff
@@ -95,6 +95,7 @@ dpkg_build_sstate_finalize_append() {
 
 do_deploy() {
     cp -f "${S}/${SPI_FLASH_IMG}" "${DEPLOY_DIR_IMAGE}/${SPI_FLASH_DEPLOY_IMG}"
+    cp -f "${S}/u-boot-initial-env" "${DEPLOY_DIR_IMAGE}/"
```
jan-kiszka, line 98: You also need to update sstate caching.

AsuraZeng: HI jan
I have no expericing how to modify this.
and for the deploy stage, do there need to add to the sstate caching?

jan-kiszka: Just study the code a few lines above and/or the commit that added the already existing line. And, yes, it's mandatory to not break local builds.

AsuraZeng: has modified, pls help to review

jan-kiszka (approved):

BaochengSu (approved):

PR #358, opened 2022-09-21
customizations-example: create a symbolic link for serial port

AsuraZeng:
For advanced board and basic board, serial port have different names For advanced board serial port named ttyUSB0, whereas for basic board, serial port named ttyS2

Another problem is when using 4G or 5G card go through USB signal, according to the probe sequence, ttyUSB0 may not represent serial port for advanced board.

So create a dev symbolic link ttyX30 represent serial port.


BaochengSu (approved):

BaochengSu:
> Very welcome! We should also update/add documentation and make our customer support aware of this.

Time to write some wiki.

PR #359, opened 2022-09-26
iot2050-firmware-update: Fix env that preserved is not right

AsuraZeng:
Issue:
we used this tool on the previous release image found that if there were tow identical values will be used to sort higher. However, the backward value is used on the latest image.

It is unreasonable to have two identical environment variables in the same uboot env file. So remove the same environment variables from the default list to fix this issue.

### DIFF
--- a/recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
+++ b/recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
@@ -111,7 +111,6 @@ import subprocess
 import tempfile
 import hashlib
 
-
 class UpgradeError(Exception):
     def __init__(self, ErrorInfo):
         super().__init__(self)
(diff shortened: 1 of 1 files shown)


BaochengSu:
As we have discussed offline, if the env does not exist, for 1.1.1 the error must be handled.

jan-kiszka:
Sorry, the commit message does not fully parse for me. Was this observed in a real setup?

AsuraZeng:
> 

Yes. it was observed on a real setup. 
for the latest release image, this tool works well. but in the previous image(transitional image), it would fail.

in the default env file `u-boot-init-env`,  there exists one env  `boot_targets`,

if we want to preserve that env `boot_targets`, for the original design, we would append the env item to the last of  `u-boot-init-env`, then using `mkenvimage`  to create a new binary then flash to the env mtd partitions.

the problem is that if we append one that exists in the `u-boot-init-env`, so there would two identical env items in the same file.

Through experiment,  in the Debian bullseyes, using `fw_printenv` it would select the last one, so it could work as our expectation. But in the buster, it would select the previous one, we would get the wrong env.

Has two same environments in the file is not correct, so remove one

jan-kiszka:
OK, makes more sense. Maybe enhance to commit log a bit (also: tow -> two).

jan-kiszka (approved):

PR #362, opened 2022-09-27
enable OPTee secure storage and RPMB key paring  in u-boot

BaochengSu:
Enable eMMC RPMB support and OPTee driver in u-boot so that secure storage could be usable in u-boot.

Also add the RPMB key paring support in u-boot.


recipes-bsp/u-boot/files/0033-iot2050-Enable-eMMC-RPMB-support-and-OPTee-driver.patch
```diff
@@ -0,0 +1,47 @@
...
+Subject: [PATCH 33/33] iot2050: Enable eMMC RPMB support and OPTee driver
+
+By enabling the eMMC RPMB support and the OPTee driver, UEFI secure boot
+can take advantage of the RPMB based secure storage provided by OPTee.
+
+This also prepares the RPMB otp key programming via U-Boot by enabling
+the `CONFIG_CMD_OPTEE_RPMB`.
+
+For security concerns, by default disable the RPMB operation in the
+`mmc` command.
```
jan-kiszka, line 13: What does this mean? Note that there is no console in secure boot mode. And if things should be off in that mode, there is an extra config snippet to flip controls.

recipes-bsp/u-boot/files/0033-iot2050-Enable-eMMC-RPMB-support-and-OPTee-driver.patch
```diff
@@ -0,0 +1,47 @@
...
+--- a/configs/iot2050_pg1_defconfig
++++ b/configs/iot2050_pg1_defconfig
+@@ -140,3 +140,7 @@ CONFIG_WDT=y
+ CONFIG_WDT_K3_RTI=y
+ CONFIG_WDT_K3_RTI_LOAD_FW=y
+ CONFIG_OF_LIBFDT_OVERLAY=y
++CONFIG_TEE=y
++CONFIG_OPTEE=y
++CONFIG_SUPPORT_EMMC_RPMB=y
++# CONFIG_CMD_MMC_RPMB is not set
```
jan-kiszka, line 32: Do we want to expose this to PG1?

recipes-bsp/u-boot/u-boot-iot2050.inc
```diff
@@ -48,6 +50,14 @@ DEBIAN_BUILD_DEPENDS_append_otpcmd = ", secure-boot-otp-provisioning"
 
 U_BOOT_CONFIG_PACKAGE = "1"
 
+python do_patch_prepend() {
```
jan-kiszka, line 53: Should be an anonymous python block.

recipes-bsp/u-boot/files/0033-iot2050-Enable-eMMC-RPMB-support-and-OPTee-driver.patch
BaochengSu: I think no harm for PG1. The advantage is that we could keep a unified baseline for both variants.

recipes-bsp/u-boot/u-boot-iot2050.inc
```diff
@@ -68,6 +78,14 @@ do_prepare_build_append_secureboot() {
     cat ${WORKDIR}/secure-boot.cfg >> ${S}/configs/${U_BOOT_CONFIG}
 }
 
+do_prepare_build_append_rpmb-setup() {
+    if [ "${PRODUCT_GENERATION}" = "pg1" ]; then
+        bbwarn "PG1 devices are not supported in storing securely"
```
jan-kiszka, line 83: This is too vague.

recipes-bsp/u-boot/README.md
```diff
@@ -71,3 +71,73 @@ Then upgrade the firmware:
 ```shell
 flashrom -p ch341a_spi -c W25Q128.V -w iot2050-image-boot.bin
 ```
+
+## eMMC RPMB otp key provisioning
+
+To utilize the eMMC RPMB as the backend for secure storage, an otp key is
+required to be programmed into the eMMC RPMB. This is a per-device key, each
+device has its unique key.
```
jan-kiszka, line 79: "This key is unique per device."

recipes-bsp/u-boot/README.md
```diff
@@ -71,3 +71,73 @@ Then upgrade the firmware:
...
+> on the device.
+>
+> This special firmware build must not be distribute outside of the secure
+> operating environment, even with the secure boot enabled. This is due to when
+> OPTee checks if the RPMB key is programmed , it uses a supplicant to
+> communicate with the non secure world eMMC driver (since the hardware resides
+> in non-secure world to begin with). If someone replaces the supplicant with a
+> rogue application always replying "RPMB key not programmed", OP-TEE will then
+> send the key in clear text, asking for the supplicant to program it, which
+> could lead to the leakage of the key.
```
jan-kiszka, line 106: The recommendation should be:
 - program RPMB key using UNSIGNED special firmware in secure operating environment
 - flash SIGNED regular image
 - program secure boot keys and enable secure boot

jan-kiszka: And: NEVER sign RPMB setup firmware version.

recipes-bsp/u-boot/files/0033-iot2050-Enable-eMMC-RPMB-support-and-OPTee-driver.patch
jan-kiszka: Also: The U-Boot RPMB commands are purely containing operations for initial deployment? If not, that would be another reason to keep them enabled in non-secure builds.

BaochengSu: There is nothing to do with secure boot.

the u-boot `mmc` command also exposes some rpmb sub-commands for read/write/key the RPMB by enabling the `CONFIG_CMD_OPTEE_RPMB` which is implied by `CONFIG_SUPPORT_EMMC_RPMB` :

```
config SUPPORT_EMMC_RPMB
	bool "Support eMMC replay protected memory block (RPMB)"
	imply CMD_MMC_RPMB
```

Esp. for the `mmc rpmb key` sub command, it can program anything as the otp key into the RPMB, if such command is exposed to our pre-installed u-boot, then in some cases (such as on the way shipping), somebody could easily take the advantage of this command to flash anything into the RPMB otp, to induce a RPMB subsystem Denial-of-Service.

BaochengSu: A better way is to use a separate CONFIG_ for the `mmc rpmb key` sub-command.

recipes-bsp/u-boot/u-boot-iot2050.inc
BaochengSu: Reworded to `PG1 devices do not supported RPMB based secure storage`

recipes-bsp/u-boot/README.md
BaochengSu: The whole warning reworded to:

> :warning:
> This special firmware must NOT be signed.
>
> This special firmware must run in a secure operating environment.
>
> It is recommended that after programming the RPMB otp key, flash the signed
> regular images then program the secure boot keys and enable secure boot.

recipes-bsp/u-boot/files/0033-iot2050-Enable-eMMC-RPMB-support-and-OPTee-driver.patch
BaochengSu: > Also: The U-Boot RPMB commands are purely containing operations for initial deployment? If not, that would be another reason to keep them enabled in non-secure builds.

If RPMB key is not programmed, then this `mmc rpmb` command could not do anything but programming the key. However,  this command does not generate key but read the key from somewhere in the memory, this is not what we suggest to do.
 
U-Boot provides two command to manipulate the RPMB, what we should use is the other `optee-rpmb` command which generates the key inside optee.

BaochengSu: So even for non-secure build, to reduce the chance of misoperation, the rpmb sub command provided by `mmc` should be disabled by default.

BaochengSu: For the read/write operations, since this command use memory key instead of the key generated in OPTee, it is totally useless on our board.

jan-kiszka: Accepted as safety measure, rejected as pseudo security measure (every attacker can install anything as long as the device is not in secure mode).

recipes-bsp/u-boot/README.md
jan-kiszka: Maybe even "To protected the RPMB key, it is required to flash a signed image, program secure boot keys and enable secure boot."

recipes-bsp/u-boot/files/0033-iot2050-Enable-eMMC-RPMB-support-and-OPTee-driver.patch
BaochengSu: what about change `For security concerns, by default disable the RPMB operation in the
mmc command.` 

to `Disable the mmc rpmb sub command to aviod mis-programming of the RPMB key `

BaochengSu: Or `Disable the mmc rpmb sub command to reduce the chance of mis-programming of the RPMB key`

jan-kiszka: Ack.

BaochengSu:
Change in V2:

1. Fix the symbolic links in secure-boot-otp-provisioning
2. Make PRODUCT_GENERATION global to reuse it in OPTee for warning PG1 devices on RPMB setup.
3. Integrate edk2 to isar-cip-core.
4. Integrate edk2 StandaloneMM to OPTee.
5. Enable StandaloneMM managed UEFI variable operation in u-boot.
6. Make `optee-rpmb` command default to u-boot config, since optee build is controlled, expose this command would not bring risks.
7. Rewords u-boot README regarding RPMB key provisioning.

jan-kiszka:
Let me finish the isar-cip-core review and integration of Sven's patches. We don't want to rebase again here (or worse) but rather consume an upstream-accepted revision.

BaochengSu:
> Let me finish the isar-cip-core review and integration of Sven's patches. We don't want to rebase again here (or worse) but rather consume an upstream-accepted revision.

Fine, schedule-wise I will continue work on this branch for the further topics, such as Key enrolling. To me, the EDK2 part of Sven's patch seems acceptable except the git clone takes too long and occupies too much space. So once there is a final state of the upstream patch I will update here accordingly, since that part takes only one commit in my patch set, it should not produce too many troubles by then.

BaochengSu:
Change in V3:

- Disable the REE FS for secure storage in optee, RPMB based secure storage is mandatory.
- Add optee-client package, preparing for UEFI key enrolling under linux user-land.
- Rename download tarball name for optee-os for avoiding name collision.

UEFI key enrolling is on the way.

disk-enc-on-iot2050.md
```diff
@@ -0,0 +1,64 @@
...
+keyctl add encrypted enckey "load `cat enckey.blob`" @u
+
+### create the dm-crypt target at /dev/mapper/x
+dmsetup create x --table "0 31080448 crypt capi:xts(aes)-plain64 :32:encrypted:enckey 0 /dev/mmcblk1 0"
+partprobe
+
+### mount the device and dump the test file
+mount /dev/mapper/x1 test
+hexdump test/test.bin
+```
```
stormc, line 64: I assume this is eventually going to be done automatically in the `initramfs`, right? How do you envision this to happen? 

As you're doing the integration manually here (using the plumbing `dmsetup` command, e.g.), is there a reason you're not using LUKS2 Tokens, see [cryptsetup FAQ :: 10.12 What is a LUKS2 Token](https://gitlab.com/cryptsetup/cryptsetup/-/wikis/FrequentlyAskedQuestions#10-luks2-questions)? 
[systemd](https://github.com/systemd/) is for example providing a [cryptsetup](https://gitlab.com/cryptsetup) external LUKS token plugin to integrate a TPM with the kernel keyring and by this extension to cryptsetup. 
Granted, for your way of storing the passphrase it's going to need a similar plugin or manual plumbing for the time being but there's merit in adopting the token scheme: it's following a paved track and is not a custom solution you have to maintain...

Given that the integration is already done in systemd (which does not require systemd to be running as the `initramfs`'s `init`), have you considered using a fTPM instead of this specific scheme?

BaochengSu: This is only a POC to demostrate one possible way to do the pure disk encryption on IOT2050, with the help of the kernel trusted/encrypted key backed by OPTee trusted key TA. The advantage is that the disk encryption key never leaves the kernel space.

However, as you've pointed out, this method does not have the fancy features that LUKS could provide, such as the re-encryption. And at this moment it is not possible to bind it together with LUKS2/systemd solution, due to there is no kernel trusted/encryption key support in cryptsetup/systemd. As you've supported, it's not likely possible to maintain such plugin for us.

So now I am already working on the fTPM solution as described in https://www.linaro.org/blog/securing-a-device-with-trusted-substrate/.

stormc: Ah, OK, thanks for the heads-up! So, just for clarification, you're PoC'ing the fTPM schema and adopt the plumbing/integration that systemd has laid out?

BaochengSu: Yep.

BaochengSu:
This PR is partially obsoleted by #426, the rest will be obsoleted by upcoming disk encryption integration.

PR #364, opened 2022-10-10
Update SEBoot to D01.03.01.01

BaochengSu:
Fix the false negative reports when POLICY_SOFT is set.


AsuraZeng (approved):
Look through the seboot code, LGTM.

PR #365, opened 2022-10-11
Rework M.2 support

jan-kiszka:
This changes the interface between firmware and kernel for configuring
and communicating the M.2 mode.

Previously, a different device tree was selected by the firmware when a
different mode was set, including repeating updates of the U-Boot
environment in flash on every boot. This approach here uses the device
tree fix-up of U-Boot to adjust a given DT to the selected mode.

Further changes:
 - permit manual configuration in secure boot mode
 - report selected mode and if it was chosen manually or as fallback
 - align U-Boot DT to kernel DT
 - document manual configuration
 - cleanups and simplifications


recipes-bsp/u-boot/files/0027-arm-dts-iot2050-Add-support-for-M.2-variant.patch
```diff
@@ -114,14 +122,34 @@ index 0000000000..741b9177da
 +	>;
 +};
 +
++/*
++ * Default configuration for B-key with 2 PCIe lanes, E-key disabled.
```
AsuraZeng, line 126: Default is B-key with usb3.0 and E-key with pcie one lane.
even if B-key with two lane, E-key could also use 2.0

recipes-bsp/u-boot/files/0028-iot2050-Add-support-for-configuring-M.2-connector.patch
```diff
@@ -0,0 +1,534 @@
...
++				 <&k3_clks 151 8>;  /* set PIPE3_TXB_CLK to WIZ8B2M4VSB */
++	phys = <&serdes0 PHY_TYPE_USB3 0>;
++	phy-names = "usb3-phy";
++};
++
++&usb0 {
++	maximum-speed = "super-speed";
++	snps,dis-u1-entry-quirk;
++	snps,dis-u2-entry-quirk;
++};
```
AsuraZeng, line 193: As mentioned above, default should be bkey usb3.0 + ekey pcie

recipes-bsp/u-boot/files/0028-iot2050-Add-support-for-configuring-M.2-connector.patch
```diff
@@ -0,0 +1,534 @@
...
++static void m2_overlay_prepare(void)
++{
++#if defined(CONFIG_OF_LIBFDT) && defined(CONFIG_OF_BOARD_SETUP)
++	const char *overlay_path;
++	void *overlay;
++	u64 loadaddr;
++	ofnode node;
++	int ret;
++
++	if (connector_mode == BKEY_PCIEX2_EKEY_NONE)
```
AsuraZeng, line 356: same as above

recipes-bsp/u-boot/files/0028-iot2050-Add-support-for-configuring-M.2-connector.patch
```diff
@@ -0,0 +1,534 @@
...
++					       CONNECTOR_MODE_INVALID);
++	const char *mode_info = "";
++	struct m2_config_pins config_pins;
++	unsigned int n;
++
++	/* enable M.2 connector power */
++	set_pinvalue("gpio@601000_17", "P3V3_M2_EN", 1);
++	udelay(4 * 100);
++
++	m2_manual_config = env_get_ulong("m2_manual_config", 10,
```
AsuraZeng, line 402: 392 line, twice read

recipes-bsp/u-boot/files/0028-iot2050-Add-support-for-configuring-M.2-connector.patch
```diff
@@ -0,0 +1,534 @@
...
+From: Jan Kiszka <jan.kiszka@siemens.com>
+Date: Tue, 11 Oct 2022 13:23:21 +0200
+Subject: [PATCH 28/32] iot2050: Add support for configuring M.2 connector
+
+The M.2 slots of the related IOT2050 variant need to be configured
+according to the plugged cards. This tries to detect the card using the
+M.2 configuration pins of the B-key slot. If that fails, a U-Boot
+environment variable can be set to configure manually. This variable is
+write-permitted also in secure boot mode as it is not able to undermine
+the integrity of the booted system.
```
AsuraZeng, line 11: manual configuration prior to auto-detection.
so it's not "If that fails, a U-Boot environment variable can be set to configure manually"

whether or not fails, manual configuration should be first priority if existed

recipes-bsp/u-boot/files/0028-iot2050-Add-support-for-configuring-M.2-connector.patch
```diff
@@ -0,0 +1,534 @@
...
++
++	/* enable M.2 connector power */
++	set_pinvalue("gpio@601000_17", "P3V3_M2_EN", 1);
++	udelay(4 * 100);
++
++	m2_manual_config = env_get_ulong("m2_manual_config", 10,
++					 CONNECTOR_MODE_INVALID);
++	if (m2_manual_config < CONNECTOR_MODE_INVALID) {
++		mode_info = " [manual mode]";
++		connector_mode = m2_manual_config;
```
AsuraZeng, line 406: should check range

recipes-bsp/u-boot/files/0028-iot2050-Add-support-for-configuring-M.2-connector.patch
```diff
@@ -0,0 +1,534 @@
...
++			if (!memcmp(config_pins.config,
++				    m2_config_table[n].config_pins.config,
++				    sizeof(config_pins.config))) {
++				connector_mode = m2_config_table[n].mode;
++				break;
++			}
++		}
++		if (connector_mode == CONNECTOR_MODE_INVALID) {
++			mode_info = " [fallback, card unknown/unsupported]";
++			connector_mode = BKEY_USB30_EKEY_PCIE;
```
AsuraZeng, line 423: this is the real default mode.

recipes-bsp/u-boot/files/0028-iot2050-Add-support-for-configuring-M.2-connector.patch
```diff
@@ -0,0 +1,534 @@
...
++		for (n = 0; n < ARRAY_SIZE(config_pins.config); n++)
++			config_pins.config[n] =
++				get_pinvalue(m2_bkey_cfg_pin_info[n].gpio_name,
++					     m2_bkey_cfg_pin_info[n].label);
++		connector_mode = CONNECTOR_MODE_INVALID;
++		for (n = 0; n < ARRAY_SIZE(m2_config_table); n++) {
++			if (!memcmp(config_pins.config,
++				    m2_config_table[n].config_pins.config,
++				    sizeof(config_pins.config))) {
++				connector_mode = m2_config_table[n].mode;
```
AsuraZeng, line 417: mode_info should also be set

recipes-bsp/u-boot/files/0028-iot2050-Add-support-for-configuring-M.2-connector.patch
```diff
@@ -0,0 +1,534 @@
...
++	int err;
++
++	if (!connector_overlay)
++		return;
++
++	overlay = malloc(connector_overlay_size);
++	if (!overlay)
++		goto fixup_error;
++
++	memcpy(overlay, connector_overlay, connector_overlay_size);
```
AsuraZeng, line 473: why memcpy again？

recipes-bsp/u-boot/files/0028-iot2050-Add-support-for-configuring-M.2-connector.patch
```diff
@@ -0,0 +1,534 @@
...
++	err = fdt_overlay_apply_verbose(blob, overlay);
++	free(overlay);
```
AsuraZeng, line 475: two malloc one free

recipes-bsp/u-boot/files/0028-iot2050-Add-support-for-configuring-M.2-connector.patch
```diff
@@ -0,0 +1,534 @@
...
++
++	if (!err)
++		return;
```
AsuraZeng, line 478: no error print?

recipes-bsp/u-boot/files/0028-iot2050-Add-support-for-configuring-M.2-connector.patch
```diff
@@ -0,0 +1,534 @@
...
++----------------------
++
++The M.2 variant of the IOT2050 comes with one B-key and one E-key slot. These
++are configured by U-Boot depending on the detected usage (auto configuration).
++The device tree loaded later on for the OS will then be fixed up according to
++this configuration.
++
++If this auto configuration does not work reliably, the U-Boot environment
++variable "m2_manual_config" can be set to select the mode manually:
++
```
AsuraZeng, line 516: better to add: manual config always prior to auto-detect

recipes-kernel/linux/files/patches-5.10/0120-arm64-dts-ti-iot2050-Add-support-for-M.2-variant.patch
```diff
@@ -0,0 +1,161 @@
...
++&main_gpio0 {
++	pinctrl-names = "default";
++	pinctrl-0 = <&main_m2_pcie_mux_control>;
++};
++
++&main_gpio1 {
++	pinctrl-names = "default";
++	pinctrl-0 = <
++	    &main_m2_enable_pins_default
++		&main_pmx0_m2_config_pins_default
```
AsuraZeng, line 117: indention

recipes-kernel/linux/files/patches-5.10/0120-arm64-dts-ti-iot2050-Add-support-for-M.2-variant.patch
```diff
@@ -0,0 +1,161 @@
...
++		&main_pmx1_m2_config_pins_default
++	>;
++};
++
++/*
++ * Default configuration for B-key with 2 PCIe lanes, E-key disabled.
```
AsuraZeng, line 123: default is bkey usb3.0 ekey pcie

AsuraZeng:
why no dtbo in kernel?

AsuraZeng:
in secure mode, ebg will load the overlay or not?

recipes-bsp/u-boot/files/0027-arm-dts-iot2050-Add-support-for-M.2-variant.patch
jan-kiszka: Nope, this was changed in order to make overlays possible: U-Boot does not support removing properties via overlays. Therefore, the baseline DT must start in a way that we only add or change properties.

jan-kiszka: Oh, that the E-key can have a role when the B-key uses two PCI lanes, that was nowhere mentioned so far. Can you elaborate please?

recipes-bsp/u-boot/files/0028-iot2050-Add-support-for-configuring-M.2-connector.patch
jan-kiszka: Nope.

recipes-bsp/u-boot/files/0028-iot2050-Add-support-for-configuring-M.2-connector.patch
jan-kiszka: It is meant as "if auto-configuration does not work for you, you can set up manual configuration". I will try to reword this to make it clearer.

recipes-bsp/u-boot/files/0028-iot2050-Add-support-for-configuring-M.2-connector.patch
jan-kiszka: oops

recipes-bsp/u-boot/files/0028-iot2050-Add-support-for-configuring-M.2-connector.patch
jan-kiszka: "For the case auto configuration does not work reliably, it is possible to set
the U-Boot environment variable "m2_manual_config" to select the mode manually:"

recipes-bsp/u-boot/files/0028-iot2050-Add-support-for-configuring-M.2-connector.patch
jan-kiszka: Connector mode can be any value if the user does not read the documentation.

recipes-bsp/u-boot/files/0028-iot2050-Add-support-for-configuring-M.2-connector.patch
jan-kiszka: It's changed, as I wrote.

recipes-bsp/u-boot/files/0028-iot2050-Add-support-for-configuring-M.2-connector.patch
jan-kiszka: Apply the overlay will unfortunately modify that blob. So we need to preserve the original for future fix-ups.

recipes-bsp/u-boot/files/0028-iot2050-Add-support-for-configuring-M.2-connector.patch
jan-kiszka: The first malloc is (almost) persistent: we only free the first copy implicitly when U-Boot boot-time services terminate (on hand-over to OS during legacy boot or when UEFI boot-time services are terminated). Thus no free.

recipes-bsp/u-boot/files/0028-iot2050-Add-support-for-configuring-M.2-connector.patch
jan-kiszka: We fall through to it.

recipes-bsp/u-boot/files/0028-iot2050-Add-support-for-configuring-M.2-connector.patch
jan-kiszka: See above.

recipes-kernel/linux/files/patches-5.10/0120-arm64-dts-ti-iot2050-Add-support-for-M.2-variant.patch
jan-kiszka: Fixed

recipes-kernel/linux/files/patches-5.10/0120-arm64-dts-ti-iot2050-Add-support-for-M.2-variant.patch
jan-kiszka: Nope.

jan-kiszka:
> why no dtbo in kernel?

The kernel has not use case for that overlay.

jan-kiszka:
> in secure mode, ebg will load the overlay or not?

No, firmware will load it (it's embedded) and apply it when the UEFI bootloader (EBG, systemd-boot) requests EFI_DT_FIXUP_PROTOCOL.

recipes-bsp/u-boot/files/0028-iot2050-Add-support-for-configuring-M.2-connector.patch
jan-kiszka: Empty mode_info implies "auto" so far. I don't think we need to write this out explicitly.

recipes-bsp/u-boot/files/0028-iot2050-Add-support-for-configuring-M.2-connector.patch
jan-kiszka: Oh, misread as "why are we checking the range" - we are checking it to be >= 0 and < CONNECTOR_MODE_INVALID.

recipes-bsp/u-boot/files/0028-iot2050-Add-support-for-configuring-M.2-connector.patch
AsuraZeng: misread the err.

AsuraZeng:
> > why no dtbo in kernel?
> 
> The kernel has not use case for that overlay.

yes, but we intend to overlay kernel's dtb.However the dtb is produced under uboot, I think it's better to place copies in the kernel also.

jan-kiszka:
>

jan-kiszka:
> > > why no dtbo in kernel?
> > 
> > 
> > The kernel has not use case for that overlay.
> 
> yes, but we intend to overlay kernel's dtb.However the dtb is produced under uboot, I think it's better to place copies in the kernel also.

Again: what for?

recipes-bsp/u-boot/files/0027-arm-dts-iot2050-Add-support-for-M.2-variant.patch
jan-kiszka: OK, will update the comment and outputs accordingly.

recipes-bsp/u-boot/files/0028-iot2050-Add-support-for-configuring-M.2-connector.patch
jan-kiszka: We write out the detected mode already.

recipes-bsp/u-boot/files/0028-iot2050-Add-support-for-configuring-M.2-connector.patch
jan-kiszka: Then we hard-code in the code the max size of the overlay.

recipes-bsp/u-boot/files/0027-arm-dts-iot2050-Add-support-for-M.2-variant.patch
BaochengSu: The title in iot2050setup should be refined to `M.2 PCIE/USB3.0 Configuration`

AsuraZeng:
> > > > why no dtbo in kernel?
> > > 
> > > 
> > > The kernel has not use case for that overlay.
> > 
> > 
> > yes, but we intend to overlay kernel's dtb.However the dtb is produced under uboot, I think it's better to place copies in the kernel also.
> 
> Again: what for?

Can this dtbo be used in the kernel? Put dtbo in the file system.
The advantage of this is that there is no need to forcibly bind the dtbo of the kernel to uboot.
The disadvantage may also require additional consideration of the secure aspect

recipes-bsp/u-boot/files/0028-iot2050-Add-support-for-configuring-M.2-connector.patch
BaochengSu: Find for this round. However, better to add comments to this memory area so that people who do not know the details would not treat this area as redundant and delete it accidentally, this could be done in the upstreaming patch.

BaochengSu (approved):

jan-kiszka:
I'll also prepare an MR for master then, or should I wait for further testing? Ordering is unusual already anyway.

chombourger:
There is unfortunately no PV or PR bump in this changeset meaning that the linux-image-iot2050 package will have the same version and will not be upgraded in the `apt dist-upgrade` scenario

jan-kiszka:
Right, we do not address package updates so far, specifically of the kernel as it would not work under all circumstances (secure boot). That said, we can discuss where such a versioning policy would help, how is should look like and how we make sure to enforce it.

PR #366, opened 2022-10-12
Integrate new tool k3conf

AsuraZeng:
Introduction:
K3CONF is a Linux user-space standalone application designed to provide a quick'n easy way to dynamically diagnose Texas Instruments' K3 architecture based processors link:https://git.ti.com/cgit/k3conf/k3conf/about/

It's very useful tool to get and set the clock and so on. For current usage, we can use this tool to switch cpu frequency from user-space.

### DIFF
--- a/recipes-app/k3conf/files/debian/k3conf.install
+++ b/recipes-app/k3conf/files/debian/k3conf.install
@@ -1,0 +1,1 @@
+./k3conf /usr/bin
--- a/recipes-app/k3conf/k3conf_0.2.bb
+++ b/recipes-app/k3conf/k3conf_0.2.bb
@@ -1,0 +1,27 @@
+#
+# Copyright (c) Siemens AG, 2022
+#
+# Authors:
+#  Chao Zeng <chao.zeng@siemens.com>
+#
+# This file is subject to the terms and conditions of the MIT License.  See
+# COPYING.MIT file in the top-level directory.
+#
+inherit dpkg
+
+DESCRIPTION = "A Powerful Diagnostic Tool for Texas Instruments K3 based Processors"
+MAINTAINER = "chao.zeng@siemens.com"
+
+SRC_URI += "https://git.ti.com/cgit/k3conf/k3conf/snapshot/k3conf-0.2.tar.gz;protocol=https \
+            file://debian"
+SRC_URI[sha256sum] = "4e19f6fdee6d40216d786c7d9000191771fe3ff467e5b5efe659ad5acb6f610f"
+
+S = "${WORKDIR}/k3conf-${PV}"
+
+do_prepare_build[cleandirs] += "${S}/debian"
+
+do_prepare_build() {
+    cp -r ${WORKDIR}/debian ${S}/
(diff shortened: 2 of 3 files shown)


jan-kiszka:
What exactly is it providing? Not mentioned above, not even mentioned by TI.

AsuraZeng:
> 

Currently its focuses mainly on TISCI related functionality.
But there is currently no interface to read the frequency of the cpu，I'm just giving an example of what this can be used for

I think  we can use this tool to debug sysfw or to verify the clock is set correctly

jan-kiszka:
Hmm, this looks dangerous:
```
root@iot2050-debian:~/k3conf# ./k3conf set clock 202 0 1000000
|--------------------------------------------------------------------------------|
| VERSION INFO                                                                   |
|--------------------------------------------------------------------------------|
| K3CONF | (version v0.1-67-g9f4abe9 built Wed 12 Oct 2022 01:54:50 PM CEST)     |
| SoC    | AM65x SR2.0                                                           |
| SYSFW  | ABI: 3.1 (firmware version 0x0015 '21.9.1--v2021.09a (Terrific Lla)') |
|--------------------------------------------------------------------------------|

|---------------------------------------------------------------------------------------------------|
| Device ID | Clock ID | Clock Name                             | Status          | Clock Frequency |
|---------------------------------------------------------------------------------------------------|
|   202     |     0    | DEV_COMPUTE_CLUSTER_A53_0_BUS_ARM0_CLK | CLK_STATE_READY | 1000000         |
|---------------------------------------------------------------------------------------------------|
```
And then it hang. Well, running at 1 MHz is likely no fun...

Seems you can hack the core to higher frequency without hangs. Not sure yet if those have real effects, though.

jan-kiszka:
Setting a CPU clock frequency to a supported value below 1GHz, e.g. 500MHz, does have the expected impact on the core performance here. Also over-clocking seems to work, though I only left 1.001 GHz set for a second. This looks increasingly dangerous. I wonder if we should install this by default.

And I still wonder why there is this "nice" tool but no cpufreq driver?! Are we (and upstream) missing patches, or is there a good reason to not play with this during runtime?

BaochengSu:
> And I still wonder why there is this "nice" tool but no cpufreq driver?! Are we (and upstream) missing patches, or is there a good reason to not play with this during runtime?

IIRC, Ti does not provide cpufreq driver on AM65xx.

BaochengSu:
Overlocking requires some voltage tuning as well, this tool only changes the Cluster REQ, I don't think it changes the voltage settings as well. That is speaking, even if you clock the device to 1.1 Ghz, the voltage probably not working at the best setting.

Ti provides something named as AVS, for toggling the voltage setting.

jan-kiszka:
I don't want to do overclocking. I rather want to avoid that our users are playing with this tool, ruining their hardware this way, and then complaining "but you shipped it!".

I have placed a direct question with TI regarding cpufreq drivers. They recently added AM625 downstream, but no traces of AM65.

PR #369, opened 2022-10-19
Chao/refact switchserialmode

AsuraZeng:


recipes-app/switchserialmode/files/src/serial/omap_8250.c
```diff
@@ -0,0 +1,154 @@
...
+ * Copyright (c) Siemens AG, 2019-2022
+ *
+ * Authors:
+ *  Gao Nian <nian.gao@siemens.com>
+ *  Chao Zeng <chao.zeng@siemens.com>
+ *
+ * This file is subject to the terms and conditions of the MIT License.  See
+ * COPYING.MIT file in the top-level directory.
+ */
+
```
BaochengSu, line 11: The logic in this file seems common to all tty uart? If so then `tty_uart.c` sounds better to me.

AsuraZeng: for the 8250_omap this is true,we use several extra gpio to control transceiver. I have investigated we can not write a commonly logic to control external transceiver, like mode switch and termination control. 
But the 8250_exar, it can do this.

It can't be guaranteed that all controllers behave in this way, so it's safer to keep the name of the controller

BaochengSu: Is the omap8250 controlling the transceiver or actually the GPIO pins of the SoC? If later then better to decouple the transceiver toggling out of this 8250 code.

AsuraZeng: for the omap8250 we use the GPIO pins to control the transceiver.  and for the current design, the mode switch for the controller and the transceiver are separated.

recipes-app/switchserialmode/files/src/transceiver/sp339e.c
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    if (gpiod_line_request_output(line, "switchserialmode", value) < 0) {
+        perror("gpiod_line_request_output");
+    }
+
+    gpiod_line_release(line);
+    gpiod_line_close_chip(line);
+}
+
```
BaochengSu, line 36: This is a common interface, could also be used by others, better to put it in a common place.

recipes-app/switchserialmode/files/src/switchserialmode.h
```diff
@@ -0,0 +1,45 @@
...
+
+    /*For the device which support setup and hold time*/
+    void (*rs485HoldTime)(uint8_t time);
+    void (*rs485SetupTime)(uint8_t time);
+
+    /*Release*/
+    void (*release)(void);
+
+    transceiver_ops_t *transOps;
+}serial_ops_t;
```
BaochengSu, line 41: Better to have a platform.c to include the platform definitions, and put serial and transceiver at the same level, below platform, as element of device arrays.

BaochengSu: For example:

```
struct device {
    const char *name;
    serial_ops_t *ctrl_ops;
    transceiver_ops_t *xcvr_ops;
};

struct platform {
    const char *name;
    uint8_t ndev;
    struct device [];
};

struct platform platforms[] = {
    {"IOT2050-BASIC", 1, {"ttyX30", &omap_8250_ops. &sp339eOps}};
    ...
};

```

BaochengSu: This gives you the most flexibility to support more platform, and support more devices.

recipes-app/switchserialmode/files/src/transceiver/sp339e.c
```diff
@@ -0,0 +1,68 @@
...
+    {
+        sp339e_set_mode(0, 1);
+    }
+    else if(0 == strcasecmp(mode, "rs422"))
+    {
+        sp339e_set_mode(1, 1);
+    }
+}
+
+transceiver_ops_t sp339eOps = {
```
BaochengSu, line 65: Keep the same naming style, `sp339e_ops`?

recipes-app/switchserialmode/files/src/switchserialmode.h
```diff
@@ -0,0 +1,45 @@
...
+    ADVANCED_BOARD_PG1
+} boardType_e;
+
+typedef struct transceiver_ops {
+    void (*transceiver_set_termination)( uint8_t onoff);
+    void (*transceiver_switch_mode)(const uint8_t *mode);
+} transceiver_ops_t;
+
+typedef struct {
+    /*Device node*/
```
BaochengSu, line 21: Unify the coding styles.

recipes-app/switchserialmode/files/src/transceiver/sp339e.c
AsuraZeng: using gpio_helper to abstract this function

recipes-app/switchserialmode/files/src/transceiver/sp339e.c
AsuraZeng: unify the naming rules already

recipes-app/switchserialmode/files/src/switchserialmode.h
AsuraZeng: now I use board type to initial all the ops at runtime.
if we need to support more boards, I think we can mark a TBD to do in the future.

AsuraZeng:
Is there anything else that needs to be modified in this PR?

BaochengSu (approved):

PR #371, opened 2022-10-20
node-red-gpio: Refact decouple led and state control

AsuraZeng:
Separation of class LED selection and state control. The label of the LED corresponds to the silk screen of the iot2050.

There is a problem with the state switching of the old version, This refact would also fix it.


jan-kiszka:
We are upstream, so maybe go there first?

AsuraZeng:
> We are upstream, so maybe go there first?

OK, Have create a PR for node-red-node.
https://github.com/node-red/node-red-nodes/pull/957

AsuraZeng:
upstream node-red-nodes has already merged this modification.
backport this commit

jan-kiszka (approved):

BaochengSu (approved):

PR #372, opened 2022-10-21
Updates for layers, kernel, firmware and other recipes

jan-kiszka:
Big version bump, specifically regarding Isar and isar-cip-core, but also along the stable kernel and the various nodes. Needs careful testing. And reviews.

@fmoessbauer @gylstorffq


recipes-app/node-red-node-serialport/node-red-node-serialport_1.0.1.bb
```diff
@@ -14,6 +14,7 @@ DESCRIPTION = "Node-RED nodes to talk to serial ports"
 
 NPM_LOCAL_INSTALL_DIR = "/root/.node-red"
 
-do_install_append() {
-    rm -rf $(find ${D} -name prebuilds -type d)
+do_prepare_build_append() {
+    sed -i '/override_dh_install:/a\\trm -r ${PP}/image/root/.node-red/node_modules/node-red-node-serialport/node_modules/@serialport/bindings-cpp/prebuilds' \
```
fmoessbauer, line 18: This is based on the assumption that `override_dh_install` is present. Better check in advance as this otherwise might fail silently in a future version.

classes/npm.bbclass
```diff
@@ -208,50 +209,45 @@ python clean_npm() {
...
     if [ -n "${NPM_REBUILD}" ]; then
         INSTALL_FLAGS="$INSTALL_FLAGS --build-from-source --no-save"
     fi
 
-    export CHDIR INSTALL_FLAGS
-    sudo -E chroot --userspec=$(id -u):$(id -g) ${BUILDCHROOT_DIR} sh -c ' \
-        cd $CHDIR
-        npm install $INSTALL_FLAGS ${NPM_INSTALL_FLAGS} /downloads/${@get_npm_bundled_tgz(d)}
-    '
+    cat <<EOF >> ${S}/debian/rules
```
fmoessbauer, line 235: Why not simply provide that rules file (as a template) instead of generating it on the fly?

recipes-bsp/k3-rti-wdt/files/rules
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/make -f
+
+export CC=gcc
```
fmoessbauer, line 3: Never set the compiler explicitly. This is done by the dephelper.

recipes-bsp/k3-rti-wdt/files/rules
```diff
@@ -0,0 +1,9 @@
...
+
+override_dh_fixperms:
+	dh_fixperms
```
fmoessbauer, line 6: This should be called automatically. What is the reason for this override?

recipes-bsp/u-boot/u-boot-iot2050.inc
```diff
@@ -46,28 +46,38 @@ DEBIAN_BUILD_DEPENDS_append_secureboot = ", python3-pycryptodome:native, \
...
     fi
-    ln -s ../keys ${S}
+    mkdir -p ${S}/keys
+    ln -f ${WORKDIR}/keys/* ${S}/keys/
     sed -ni '/### Secure boot config/q;p' ${S}/configs/${U_BOOT_CONFIG}
     cat ${WORKDIR}/secure-boot.cfg >> ${S}/configs/${U_BOOT_CONFIG}
 }
 
+dpkg_runbuild_append_secureboot() {
+    # remove private keys from source archive
```
fmoessbauer, line 75: We need a better pattern for that. Maybe the keys can be injected into the overlay of the schroot by using a ` -chroot-setup-commands=` (currenty ISAR has no support for that).

For now, I can live with it. But should be discussed on the ISAR ML.

fmoessbauer (commented):
In general, this already looks pretty good. Just some minor comments.

recipes-app/node-red-node-serialport/node-red-node-serialport_1.0.1.bb
jan-kiszka: Both class and recipe are in the same layer. And if this rule is not patched properly, the build will fail anyway.

classes/npm.bbclass
jan-kiszka: This is a class, not some recipe include with own files folder.

recipes-bsp/k3-rti-wdt/files/rules
jan-kiszka: We are building firmware here, and that does not follow the host CPU rules.

recipes-bsp/k3-rti-wdt/files/rules
jan-kiszka: Right, that was imported blindly by taking a deb_debianize output. Will clean up.

classes/npm.bbclass
fmoessbauer: Ok, understood. Then it should be fine.

recipes-bsp/k3-rti-wdt/files/rules
fmoessbauer: Ok, fine.

recipes-bsp/optee-os/optee-os-iot2050_3.18.0.bb
```diff
@@ -34,7 +35,3 @@ python do_patch_prepend() {
     if re.search("rpmb-setup", overrides) and re.search("secureboot", overrides):
         bb.fatal("Not possible to use Secure Boot and RPMB setup for OPTEE")
 }
```
BaochengSu, line 37: This could also be refined to inside an anonymous function.

jan-kiszka: You are right - separate commit, will add it.

recipes-bsp/optee-os/optee-os-iot2050_3.18.0.bb
```diff
@@ -21,20 +21,17 @@ OPTEE_NAME = "iot2050"
```
BaochengSu, line 21: 3.19.0 released, which carries a patch that we have interest into:

https://github.com/OP-TEE/optee_os/commit/115198b44dffbe088b4aec0ff82a7d32d7d704d2

Kconfig
```diff
@@ -139,7 +127,7 @@ config SECURE_BOOT
 
 config OTP_PROVISIONING
 	bool "OTP Provisioning"
-	depends on IMAGE_BOOT_PG2 || IMAGE_BOOT_M2
+	depends on IMAGE_BOOT_PG2 && !SECURE_BOOT
```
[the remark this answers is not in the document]
jan-kiszka: Oh, confused this. Will remove this again.

recipes-bsp/optee-os/optee-os-iot2050_3.18.0.bb
jan-kiszka: OK, will update.

AsuraZeng (commented):
should we also add this explanation to the README?

Kconfig
```diff
@@ -41,21 +41,10 @@ config IMAGE_BOOT_PG1
 	  allows to recover. Otherwise you risk to BRICK THE IOT2050!
 
 config IMAGE_BOOT_PG2
-	bool "Firmware image for PG2 devices"
```
AsuraZeng, line 44: should we also add this explanation to the README?

jan-kiszka: It's in the Kconfig description, and the README points there already.

BaochengSu (approved):

PR #373, opened 2022-10-31
integrate docker support

AsuraZeng:
For now,  I want to make the software configurable and optional. At present, the preinstalled software and some of our custom software have been made configurable

It's still a draft, Please help to review whether this approach is appropriate

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -87,6 +87,17 @@ config KAS_INCLUDE_NO_CORAL
 	default "kas/opt/no-coral.yml"
 	depends on !CORAL
 
+config DOCKER
+	bool "Docker support"
+	default n
+	help
+	  Building image with docker support.
+
+config KAS_INCLUDE_DOCKER
+	string
+	default "kas/opt/docker.yml"
+	depends on DOCKER
+
 config LXDE
 	bool "LXDE graphical user interface"
 	help
--- a/kas/opt/docker.yml
+++ b/kas/opt/docker.yml
@@ -1,0 +1,16 @@
+#
+# Copyright (c) Siemens AG, 2022
+#
+# Authors:
+#  Chao Zeng <chao.zeng@siemens.com>
+#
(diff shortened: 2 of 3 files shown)


AsuraZeng:
@jan-kiszka @BaochengSu  
currently, this PR is ready for review.  pls help to review and add your comments.
thanks a lot

AsuraZeng:
@jan-kiszka  hi jan.do you any comments on this PR?

jan-kiszka:
Two things:
- We already do feature / package control in meta-iot2050. Why did you deviate from that pattern?
- Please provide some reasoning why each of those features should be controllable. It's obvious for heavy stuff like docker, but why controlling 32-bit support?
- IOT2050_DEBIAN_DEBUG_PACKAGES vs. IOT2050_BASIC_SOFTWARE_SUPPORT - does not sound like being the same thing.
- defaults differ in Kconfig vs. image recipe

jan-kiszka:
The existing pattern is to have kas options fragment and select them in the kconfig. And there are also related entries in the menu that you didn't put into the same category now.

AsuraZeng:
> 

if every item has correspond kas yml, the kas menu is just a front end of  build system.
if we'd like to control all the packages, there must many correspond yml files. 

Can kas menu be separated from kas yml ?
Selecting through the menu interface is more intuitive for some users
If we want to make software controllable, is there any elegant way?

jan-kiszka:
Yes, this does not scale very well - and it relates to the discussion of what to make configurable **at all**. If you start adding a menu item for each possible package, kconfig will not scale either.

AsuraZeng:
> 

@jan-kiszka  Has already refine this PR, pls help to review.

jan-kiszka (commented):
Looks good, I would just call the option file docker.yml - shorter.

jan-kiszka (approved):

PR #375, opened 2022-11-16
iot2050setup: Update default setup and hold time for advanced board

AsuraZeng:
the usb-serial we are using is cp2102n, when set to rs485 mode, setup time and hold time should refer the transceiver's datasheet (sp339e https://assets.maxlinear.com/web/documents/sp339e.pdf)

Discuss with FAE,Setup time should be over tDZH 1us,hold time should be over tDPLH 1us.

Currently, default value is 0xaa, this time is too long for the hold and setup time. This will affect the efficiency of communication and cause data loss.

And according to the field feedback, this default value set to 0x05 us meets most of the scenarios.


jan-kiszka:
How does this go along with switchserialmode? Shouldn't such parameter also be (or become) defaults on the command line?

AsuraZeng:
Seems induOs just use this tool to configure the serial port?
integrate this change the switchserialmode tool seems more reasonable.
I would refine this PR.

jan-kiszka:
Yes, iot2050setup is not used in Industrial OS.

I would keep the times configurable via switchserialport and possibly iot2050setup, but known-to-work default values should then indeed come already via switchserialmode.

AsuraZeng:
> 
Currently, the default value is set by switchserialmode. and in the iot2050setup also can change this value.

recipes-app/switchserialmode/files/src/serial/cp2102n.c
```diff
@@ -238,12 +240,20 @@ static void cp2102n_hardware_reset(void)
 
 static void cp2102n_set_rs485_setup_time(uint8_t setuptime)
 {
-    cp2102n_cfg_operation->cfg_rs485_setup_time(cp2102n_cfg_operation, setuptime);
+    if (0 == GET_CP2102N_RS485_SETUP_TIME(cp2102n_cfg_operation->deviceConf)){
+        cp2102n_cfg_operation->cfg_rs485_setup_time(cp2102n_cfg_operation, DEFAULT_RS485_SETUP_TIME);
+    } else {
+        cp2102n_cfg_operation->cfg_rs485_setup_time(cp2102n_cfg_operation, setuptime);
+    }
```
BaochengSu, line 247: Set the default value in `cp2102_ops_init`, after `cp2102n_read_config`. Then here no change.

Also the code here does not have the time configured if only `-m 485` is provided without setup time and hold time parameter.

BaochengSu: The correct behavior should be once the mode is switched to RS485, these two time parameter must be configured to 0x05, or any other value that specified by the command line arguments.

recipes-app/switchserialmode/files/src/serial/cp2102n.c
```diff
@@ -29,6 +29,8 @@
 #define CP2102N_RS485_HOLD_INDEX       671
 #define MAIN_RESET_LATCH_P1_INDEX      587
 #define CP2102N_GPIO2_RESET_LATH_BIT   5
+#define DEFAULT_RS485_SETUP_TIME 5
+#define DEFAULT_RS485_HOLD_TIME  5
```
BaochengSu, line 33: The default value is decided by the PHY, not the CP2102n itself, so better to refine to some compiling time provided env, such env could be defined in the bb file at this moment, and in the future could be decidable according to the board variant.

BaochengSu:
```c
#ifdef BOARD_SETUP_TIME
#define DEFAULT_RS485_SETUP_TIME    BOARD_SETUP_TIME
#else
#define DEFAULT_RS485_SETUP_TIME    (0x05)
#endif
```

same for `DEFAULT_RS485_HOLD_TIME`

recipes-app/iot2050setup/files/iot2050setup.py
```diff
@@ -538,13 +538,10 @@ def setRS485SetupHoldTime(self):
...
 
-        disSetup = '0xaa' if int(setup, 0) == 0 else setup
-        disHold = '0xaa' if int(hold, 0) == 0 else hold
-
         action, values = EntryWindow(screen=self.topmenu.gscreen,
                                      title='Set The Setup and Hold Time of RS485 Mode',
                                      text='The setup and hold time will affect the transfer stable in RS485 mode',
-                                     prompts=[('Setup (0x00 ~ 0xffff): ', disSetup), ('Hold (0x00 ~ 0xffff): ', disHold)],
+                                     prompts=[('Setup (0x00 ~ 0xffff): ', setup), ('Hold (0x00 ~ 0xffff): ', hold)],
                                      width=70,
```
BaochengSu, line 545: This makes sense only my comment regarding the `cp2102n.c` applies.

BaochengSu:
Better to follow the 52/70 rule for the commit message.

For VS Code, [rewrap](https://github.com/stkb/Rewrap) saves my life.

recipes-app/switchserialmode/files/src/serial/cp2102n.c
AsuraZeng: > The default value is decided by the PHY, not the CP2102n itself, so better to refine to some compiling time provided env, such env could be defined in the bb file at this moment, and in the future could be decidable according to the board variant.

correct, this value is related to what's your transceiver.
But I don't think this is a good way to pass the default value from 
bb file.
Imagine that one board type has two different transceivers because of material shortage.

if in that case, I think we need more consideration of the software structures.

BaochengSu: Then add back the platform abstraction.

AsuraZeng: It can't.  We need more infomaration to bind the controller and transceiver currently. But we lack this.
We'd like to provide a recommended value for most of the scenes.  It can be adjustable. So 0x5 is just our recommendation
If this is not reasonable, so give the user a recommendation that needs to be removed.

AsuraZeng:
> Better to follow the 52/70 rule for the commit message.
> 
> For VS Code, [rewrap](https://github.com/stkb/Rewrap) saves my life.

I will try this useful tool

recipes-app/switchserialmode/files/src/serial/cp2102n.c
BaochengSu: > It can't. We need more infomaration to bind the controller and transceiver currently. But we lack this.

Such as?

AsuraZeng: > > Imagine that one board type has two different transceivers because of material shortage.
> 
> From software point of view, if these two xcers are not compatible, then they are not the same `one board type`.
just a second resource,  there is no need to add one board type.
I think this modification is OK for the current

BaochengSu: I am not talking about the board type stored on the device, but rather a `logic` board type that only meaningful for the switching serial mode scenario.

BaochengSu (approved):

PR #380, opened 2022-11-18
add support for PREMIRRORS for npm registry

fmoessbauer:
This patch enhances the npm class to support fetching npm modules via a registry specified in the PREMIRRORS bb variable. This can significantly speedup the build in CN where the US mirrors of npm are often slow.

Technically, we implement just a basic subset of the full PREMIRRORS functionality, as the actual fetching of NPM moduls is not done by by the bitbake fetcher, but as part of the npm class. To keep the code as minimal as possible, we also do not do connectivity checks of the mirror. In addition, the npm:// url has to be mirrored by another npm:// url but not a tarball, etc... as the standard bitbake fetcher would support.

### DIFF
--- a/classes/npm.bbclass
+++ b/classes/npm.bbclass
@@ -91,6 +91,16 @@ def runcmd(d, cmd, dir):
             cmd, (":\n" + output) if output else ""))
     bb.note(output)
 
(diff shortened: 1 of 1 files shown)


jan-kiszka (approved):
As the non-mirror case is covered in CI, I assume to fully qualified the mirror case locally.

PR #381, opened 2022-11-18
customizations-example: Resizing a terminal

AsuraZeng:
Console window has fixed columns and rows, there is no method to pass the window columns and rows to console terminal.
So it can not adjust the window size automatically.

This change would make adjust the console terminal column and row according to the actual terminal window size.

The most important benefit is that we can see the whole output when using journalctl systemctl and so on to output the debug info if the debug information is too long.

Autosize-console origin from iot2050setup tool.

### DIFF
--- a/recipes-core/customizations-example/files/terminal_resize.sh
+++ b/recipes-core/customizations-example/files/terminal_resize.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+#
+# Copyright (c) Siemens AG, 2022
+#
+# Authors:
+#  Chao Zeng <chao.zeng@siemens.com>
+#
+# SPDX-License-Identifier: MIT
+
+terminal_resize() {
+    local IFS='[;' escape screen_size cols rows
+    # detect the size of the screen
+    echo -ne '\e7\e[r\e[999;999H\e[6n\e8'
+    read -t 5 -sd R escape screen_size || {
+        echo Unable to detect the size of the current terminal emulator. >&2
+        return 0
+    }
+
+    cols="${screen_size##*;}" rows="${screen_size%%;*}"
+    if [[ "${cols}" -gt 0 && "${rows}" -gt 0 ]];then
+        stty cols "${cols}" rows "${rows}"
+    else
+        echo Unable to change the size of the current terminal emulator. >&2
+        return 0
+    fi
+}
+
+case $(/usr/bin/tty) in
+        /dev/ttyS3)
+        terminal_resize
+        ;;
+esac
--- a/recipes-core/customizations-example/customizations-example_0.1-iot2050-debian.bb
+++ b/recipes-core/customizations-example/customizations-example_0.1-iot2050-debian.bb
@@ -26,7 +26,8 @@ SRC_URI = " \
     file://cellular-4g \
     file://eno1-default \
     file://20-assign-ethernet-names.rules \
-    file://20-create-symbolic-link-for-serial-port.rules"
+    file://20-create-symbolic-link-for-serial-port.rules \
+    file://terminal_resize.sh"
 
 do_install() {
     # add board status led service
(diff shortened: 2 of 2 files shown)


jan-kiszka:
I'm not sure if we want to run this unconditionally, also for consoles not needing this (ssh sessions eg.).

Further findings:
 - License header missing
 - Why the script indirection, why not calling the tool directly?
 - Is python the right tool choice here (speed, footprint)?
 - What happens if the script should fail for whatever reason?

AsuraZeng:
> 

yes, this is only true for serial terminal not for the ssh session.
in the script, limit to apply this change for the tty console.

you mean the `resize`?   it seems could set to the fixed column and rows, anyway we need
to get the current terminal window size.
Why I use python script  because in the iot2050setup tool has already contains this logic, just re-use that part.

I think this is good for our usage. And if this is not good to add to the `systemd service`, maybe 
at first stage, just install this tool maybe helpful.

AsuraZeng:
@jan-kiszka  Regarding terminal resize has already re-wrote.

recipes-core/customizations-example/files/terminal_resize.sh
```diff
@@ -0,0 +1,33 @@
...
+        return 1
+    fi
+}
+
+case $(/usr/bin/tty) in
+        /dev/ttyS3)
+        terminal_resize
+        ;;
+esac
+
```
jan-kiszka, line 33: Extra line at EOF.

But this looks much more compact. Let's play with it and see if we may actually install it into the default profile later.

recipes-core/customizations-example/files/terminal_resize.sh
```diff
@@ -0,0 +1,33 @@
...
+#
+# SPDX-License-Identifier: MIT
+
+terminal_resize() {
+    local IFS='[;' escape screen_size cols rows
+    # detect the size of the screen
+    echo -ne '\e7\e[r\e[999;999H\e[6n\e8'
+    read -t 5 -sd R escape screen_size || {
+        echo unsupported terminal emulator. >&2
+        return 1
```
jan-kiszka, line 16: This concerns me: What if the user has different emulator, and this path is taken. Will the login session abort?

recipes-core/customizations-example/files/terminal_resize.sh
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    cols="${screen_size##*;}" rows="${screen_size%%;*}"
+    if [[ "${cols}" -gt 0 && "${rows}" -gt 0 ]];then
+        stty cols "${cols}" rows "${rows}"
+    else
+        echo unsupported terminal emulator. >&2
+        return 1
```
jan-kiszka, line 24: Same issue here.

recipes-core/customizations-example/files/terminal_resize.sh
AsuraZeng: ok, removed the extra line.

recipes-core/customizations-example/files/terminal_resize.sh
AsuraZeng: No, just the simulation directly returns to 1 to see what happens

BaochengSu: Also `unsupported terminal emulator.` is too vague, unsupported for what? If only for resize, then this would be too wide warning, seems to me like 'my terminal emulator is not supported at all on iot2050'.

AsuraZeng: This is so clear to show the `terminal emulator` can not be supported for auto-size.
And this is error log and if there is some wrong we can also get the info from this output

BaochengSu: I mean that, just show the error message that the current in-use term emu does not support resizing.

jan-kiszka: ...and then better do not bail out with an error code to avoid any surprises. I didn't find a specification yet regarding how shells should react on non-zero exit codes, but better safe than sorry.

BaochengSu (approved):

jan-kiszka (approved):

PR #383, opened 2022-11-23
optee-os-iot2050: Disable insecurity warning

jan-kiszka:
After a discussion with TI engineers about why this was not yet done in upstream, it turned out that they would first like to decouple the integrity of the firewall configuration for the U-Boot based R5 bootloader. We are based on an own implementation here for which we do not see a similar risk. Therefore, the existing setup of the IOT2050 is already secure.


BaochengSu:
I think eventually TI would update this CFG_ in their platform conf.mk after they fix the current troubles. So this should be removed from the make command once TI finish their work.

BaochengSu (approved):

PR #387, opened 2022-11-29
fix: add missing deps to shared libraries used by mraa

fmoessbauer:
fix: add missing deps to shared libraries used by mraa
    
This patch adds the shlibs:Depends variable to the DEBIAN_DEPENS
variable. By that, automatically resolved runtime dependencies
are automatically added in their correct version.
    
This also fixes the bug that the libjson-c was missing.
    
### DIFF
--- a/recipes-app/mraa/mraa_2.2.0+git.bb
+++ b/recipes-app/mraa/mraa_2.2.0+git.bb
@@ -28,7 +28,7 @@ DEBIAN_BUILD_DEPENDS = " \
     libjson-c-dev, \
     default-jdk:native"
 
-DEBIAN_DEPENDS = "python3, nodejs"
+DEBIAN_DEPENDS = "python3, nodejs, \${shlibs:Depends}"
 
 do_prepare_build[cleandirs] += "${S}/debian"
 


jan-kiszka:
This lacks information on how to reproduce the issue. A plain `kas-container build kas-iot2050-example.yml --target mraa` is not enough, thus the package by itself is sufficiently described.

fmoessbauer:
I wanted to keep it simple to not get into the rabbits hole of not having SOVERSION and not a split into -dev and lib- packages. The problem appears when you link against libmraa.so. Then, the transitive dependency to libjsonc is missing. Normally, dpkg-shlibdeps automatically adds these dependencies. But this seems not to work for libraries without SOVERSION. In the official image, this is not an issue as libjsonc is probably anyways installed. But I recently moved a large layer over to ISAR sbuild and there, a component could not be build because the linking failed, as only mraa was added as build dep, but not libjsonc. Luckily we catched it in the build...

jan-kiszka:
OK, that explanation - or a short summary of if - definitely belongs to the commit at least as it is not obvious at all.

jan-kiszka:
FWIW: `libcryptsetup12:arm64 depends on libjson-c5 (>= 0.15).`

fmoessbauer:
I thought about my comment again and concluded that the reason is pretty strange (just made based on the observation). In fact, the reason is much simpler: We overwrite the `DEBIAN_DEPENDS` and by that also remove the `${shlibs:Depends}`. I'll check meta-iot2050 for more of these bugs and fix it in this PR.

fmoessbauer:
Actually, the default of `DEBIAN_DEPENDS` is empty in ISAR. I'm really wondering why the default does not contain shlib depends. That would not hurt in the normal case, but save us from issues like this one.

jan-kiszka:
Yeah, maybe also an upstream Isar topic, indeed.

jan-kiszka (approved):

PR #388, opened 2022-12-01
iot2050setup: Add arduino board pinout

AsuraZeng:
Our reference design is UNO3.
Although you can query the specific pin through this:
 https://cache.industry.siemens.com/dl/files/016/109779016/att_1018122/v1/iot2050_operating_instructions_en_en-US.pdf?download=true. 
But this is not very intuitive.

It shows the block diagram of our arduino board,
which is convenient for query and use.

### DIFF
--- a/recipes-app/iot2050setup/files/iot2050setup.py
+++ b/recipes-app/iot2050setup/files/iot2050setup.py
@@ -456,6 +456,43 @@ class PeripheralsMenu:
                 self.resetPinmuxOfUserConfig('ADC_' + str(n))
         self.saveConfig(self.config)
 
+    def show_arduino_pinout(self):
+        self.arduino_pinout = '''
+        +--------------------------------------------+
+        |                                     A5/SCL |  IO19
+        |                                     A4/SDA |  IO18
+        |                                       AREF |
+        |                                        GND |
+        | N/C                                 SCK/13 |  IO13
+        | IOREF                              MISO/12 |  .
+        | RST                                MOSI/11 |  .
+        | 3V3    +---+                         SS/10 |  .
+        | 5V    -| A |-                        PWM/9 |  .
+        | GND   -| R |-                        PWM/8 |  IO8
+        | GND   -| D |-                              |
+        | VIN   -| U |-                        PWM/7 |  IO7
+        |       -| I |-                        PWM/6 |  .
+  IO14  | A0    -| N |-                        PWM/5 |  .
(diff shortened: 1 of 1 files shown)


jan-kiszka (commented):
This is indeed very useful! I'm regularly searching for this layout in the web.

How about also explaining what is meant with "ICSP", this 6-pin connector?

AsuraZeng:
> This is indeed very useful! I'm regularly searching for this layout in the web.
> 
> How about also explaining what is meant with "ICSP", this 6-pin connector?

yes, that is the 6-pin header.
The "ICSP" is from the arduino pinout, and in our operations [guide](https://cache.industry.siemens.com/dl/files/016/109779016/att_1018122/v1/iot2050_operating_instructions_en_en-US.pdf?download=true.) also call this as ICSP.

We synchronize the abbreviation of this convention, I don't think we need too much description in this pinout.
how do you suggest?

jan-kiszka:
How about something like this, to improve readability?
```
            +--------------------------------------------+
            |                                     A5/SCL |  IO19
            |                                     A4/SDA |  IO18
            |                                       AREF |
            |                                        GND |
            | N/C                                 SCK/13 |  IO13
            | IOREF                              MISO/12 |  .
            | RST                                MOSI/11 |  .
            | 3V3    +---+                         SS/10 |  .
            | 5V    -| A |-                        PWM/9 |  .
            | GND   -| R |-                        PWM/8 |  IO8
            | GND   -| D |-                              |
            | Vin   -| U |-                        PWM/7 |  IO7
            |       -| I |-                        PWM/6 |  .
      IO14  | A0    -| N |-                        PWM/5 |  .
      .     | A1    -| O |-                        PWM/4 |  .
      .     | A2     +---+                    UART_RTS/3 |  .
      .     | A3                              UART_CTS/2 |  .
      .     | A4/SDA   ICSP                    UART_TX/1 |  .
      IO19  | A5/SCL +-----------------------+ UART_RX/0 |  IO0
            |        | RST  SCK/13   MISO/12 |           |
            |        | GND  MOSI/11  5V      |           |
            |        +-----------------------+ __________/
            \_________________________________/
```

BTW, what is that notch at the bottom about, what should it represent?

AsuraZeng:
> notch
Surrounding ICSP with a dotted line is good advice.
the notch is just for looks.

BaochengSu:
> BTW, what is that notch at the bottom about, what should it represent?

Almost every uno shield has such notch.

Issue #386, opened 2022-11-28
node-red: ./src/threadpool.c:329: uv__queue_done: Assertion `uv__has_active_reqs(req->loop)' failed


bergmanu:
A customer in the [forum ](https://support.industry.siemens.com/forum/ww/en/posts/assertion-uv-has-active-reqs-req-loop-failed/289029/?page=0&pageSize=10) reported that his node-red application crashes periodically when he reads four inputs in a 7.5Hz period. 
The crash reason is:
`
Assertion 'uv__has_active_reqs(req->loop)' failed
`

I reproduced that issue with a simplified flow (see attachment): Only having three digital inputs connected and provide a "blink" signal in a 7.5Hz period by DQs of a S7-1500 PLC to each of them simultaneously.
 (Mostly) As soon as the blink application starts, node-red crashes with the following statement: 

```
Jul 13 18:19:50 iot2050-debian node-red[818]: node-red: ./src/threadpool.c:329: uv__queue_done: Assertion `uv__has_active_reqs(req->loop)' failed
Jul 13 18:19:50 iot2050-debian systemd[1]: node-red.service: Main process exited, code=killed, status=6/ABRT
Jul 13 18:19:50 iot2050-debian systemd[1]: node-red.service: Failed with result 'signal'.
Jul 13 18:19:50 iot2050-debian systemd[1]: node-red.service: Consumed 29.504s CPU time.
```

Changing the period to 1Hz does not change the behavior.

UPDATE: 
Behavior only happens when min. 2 DIs are provided with the signal simultaneously. 
When providing only one DI with a signal, it works as expected. As soon as the signal is provided to a second input, the issue happens.

Hardware / Software used: 

- IOT2050 Basic (FS:02) and IOT2050 Advanced (FS:04)
- Input Module Sink/Source
- Example Image V1.2.2 and Example Image V1.3.1.1
- Node-Red Flow:  [DI_crash_simplified.txt](https://github.com/siemens/meta-iot2050/files/10102366/DI_crash_simplified.txt)

jan-kiszka:
Reproduced, also with the board generating the signals itself (I don't have a PLC around). Seems there is some race condition, hopefully in mraa (which is easier to fix) and not in libuv, the nodejs event helper library. Needs more debugging.

BTW, it seems the issue was also reported to Node-RED before, see https://discourse.nodered.org/t/assertion-uv-has-active-reqs-req-loop-failed/68564.

jan-kiszka:
[...]

PR #390, opened 2022-12-09
mraa: Add patch to fix "uv__has_active_reqs" assertions

jan-kiszka:
See the included patch for more details. Related mraa PR is https://github.com/eclipse/mraa/pull/1102.

We may not get quick feedback from upstream, so I would recommend moving forward here already.


BaochengSu:
Seems github does not link this PR to #386 here.

PR #394, opened 2022-12-29
Align bitbake and debian dependencies

fmoessbauer:
This patch aligns the bitbake and the debian dependencies to improve partial builds. This also helps to reduce the dependency graph in general, which in turn allows for more concurrency in the build.

In addition, npm auditing is disabled at install time to significantly speed up the build.

### DIFF
--- a/classes/npm.bbclass
+++ b/classes/npm.bbclass
@@ -239,7 +239,8 @@ do_install() {
 
 do_prepare_build_append() {
     INSTALL_FLAGS="--offline --only=production --no-package-lock --verbose \
-                   --arch=${NPM_ARCH} --target_arch=${NPM_ARCH}"
+                   --arch=${NPM_ARCH} --target_arch=${NPM_ARCH} \
+                   --no-audit"
 
     if [ -n "${NPM_LOCAL_INSTALL_DIR}" ]; then
         CHDIR=${PP}/image/${NPM_LOCAL_INSTALL_DIR}
--- a/recipes-core/regen-rootfs-uuid/regen-rootfs-uuid_1.0.bb
+++ b/recipes-core/regen-rootfs-uuid/regen-rootfs-uuid_1.0.bb
@@ -10,6 +10,7 @@ inherit dpkg-raw
 
 DESCRIPTION = "This service generates an individual UUID for the rootfs during first boot"
 
+RDEPENDS = "u-boot-script"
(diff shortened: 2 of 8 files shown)


recipes-core/regen-rootfs-uuid/regen-rootfs-uuid_1.0.bb
```diff
@@ -10,6 +10,7 @@ inherit dpkg-raw
 
 DESCRIPTION = "This service generates an individual UUID for the rootfs during first boot"
 
+RDEPENDS += "u-boot-script"
```
jan-kiszka, line 13: Why `+=`?

classes/npm.bbclass
```diff
@@ -239,7 +239,8 @@ do_install() {
 
 do_prepare_build_append() {
     INSTALL_FLAGS="--offline --only=production --no-package-lock --verbose \
-                   --arch=${NPM_ARCH} --target_arch=${NPM_ARCH}"
+                   --arch=${NPM_ARCH} --target_arch=${NPM_ARCH} \
+                   --no-audit"
```
jan-kiszka, line 243: Not against it, but can you quantify the gain, roughly?

jan-kiszka (commented):
Looks good, just had minor remarks.

recipes-core/regen-rootfs-uuid/regen-rootfs-uuid_1.0.bb
fmoessbauer: Habit... will change it to `=`.

classes/npm.bbclass
fmoessbauer: On a slow FS I hat up to 25% time reduction on NPM rebuilds.

jan-kiszka (approved):

fmoessbauer:
@jan-kiszka your review is addressed now. In addition, I added a commit to add the missing nodejs Debian runtime dependency to node-red. In general, this layer is VERY imprecise about the dependencies of the individual packages. While this might not be a problem for the example images, it definitely is painful when deriving images.

jan-kiszka:
Patches that are improving the situation are always welcome.

fmoessbauer:
> Patches that are improving the situation are always welcome.

Like in every OSS project :smile: . I already talked to @BaochengSu last week about coming patch series.

Issue #281, opened 2022-03-15
Access to GPIOs only with root privileges


bergmanu:
Access to the Arduino GPIOS via bash programms "mraa-gpio", "mraa-uart" and "mraa-i2c" is only working logged in as root or using sudo being logged in as user. 

Calling the programs as user results in "free(): invalid pointer" 
![image](https://user-images.githubusercontent.com/22653568/158323303-9f13a882-8822-412d-9fb2-ff4e6a9cb9b9.png)

Is this intended? Is there any other way to access GPIOs as user without using sudo?

jan-kiszka:
You can try `chown`ing the sysfs nodes that have to be accessed and can be securely forwarded.

That `free()` bug is likely a error in some cleanup path. Needs to be analyzed.

jan-kiszka:
@fr0st61te, could you eventually have a look if / what goes wrong there in mraa when it calls that wrong free()? TIA.

bergmanu:
> You can try `chown`ing the sysfs nodes that have to be accessed and can be securely forwarded.
> 
> That `free()` bug is likely a error in some cleanup path. Needs to be analyzed.

Tried to `chown` the whole _/sys/class/gpio_ directory recursively, still same output

jan-kiszka:
Just recalling: We also use direct MMIO access for configuring the pins in mraa. That involves using /dev/mem - a no-go for anything unprivileged. So, this is doomed to fail.

fr0st61te:
> @fr0st61te, could you eventually have a look if / what goes wrong there in mraa when it calls that wrong free()? TIA.

Sure, I'll take a look!

jan-kiszka:
Merged into master, happily looking for more testers now!

PR #395, opened 2022-12-30
Support combination of node-red and swupdate

fmoessbauer:
This series reworks the integration of node-red. By that, the node-red modules can be re-added to the swupdate version of the image. In addition, access rights are limited.

Please note:

- this series depends on #394 
- it was only tested without nodes that access GPIO pins (due to missing HW). This has to be done before integration.


recipes-core/customizations-example/files/board-configuration.service
```diff
@@ -1,5 +1,5 @@
 [Unit]
-Description="IOT2050 Board Configuration"
+Description=IOT2050 Board Configuration
```
jan-kiszka, line 2: Style fix or more? This is missing in your commit message.

recipes-app/node-red/node-red-package.inc
```diff
@@ -0,0 +1,15 @@
...
+#
+# This file is subject to the terms and conditions of the MIT License.  See
+# COPYING.MIT file in the top-level directory.
+#
+
+# common setting for all node-red-contrib modules
+DEBIAN_DEPENDS .= ", node-red"
+RDEPENDS += "node-red"
+# install directory of nod-red-* modules
+NPM_LOCAL_INSTALL_DIR = "/root/.node-red"
```
jan-kiszka: Maybe rather model as class? Would allow to move the common `inherit npm` over as well.

recipes-app/node-red-preinstalled-nodes/node-red-preinstalled-nodes_0.1.bb
```diff
@@ -51,6 +51,10 @@ python do_generate_package_json() {
 addtask generate_package_json before do_install
 
 do_install() {
+    # install package.json into rootfs
+    install -d ${D}/usr/lib/nodered/.node-red
+    install -m 0644 ${WORKDIR}/package.json ${D}/usr/lib/nodered/.node-red
+    # add symlink from persistent partition to file on rootfs
     install -d ${D}/${NODE_RED_HOME_DIR}/.node-red
-    install -m 0644 ${WORKDIR}/package.json ${D}/${NODE_RED_HOME_DIR}/.node-red
+    ln -sr ${D}/usr/lib/nodered/.node-red/package.json ${D}/${NODE_RED_HOME_DIR}/.node-red/
```
jan-kiszka, line 59: Why do we need to install things into a separate dir under `/usr/lib`? Can't we use the homedir directly, as before?

recipes-app/node-red/files/node-red.service.tmpl
```diff
@@ -4,7 +4,9 @@ After=syslog.target network.target
 
 [Service]
 Type=idle
-ExecStart=/usr/bin/node-red -u %h/.node-red
+User=nodered
+WorkingDirectory=${NODE_RED_HOME_DIR}
+ExecStart=/usr/bin/node-red -u ./.node-red
```
jan-kiszka, line 9: Why do we need to specify the userdir on node-red start? Doesn't `$HOME/.node-red` work anymore?

recipes-app/node-red/node-red-package.inc
```diff
@@ -12,4 +12,4 @@
 DEBIAN_DEPENDS .= ", node-red"
 RDEPENDS += "node-red"
 # install directory of nod-red-* modules
-NPM_LOCAL_INSTALL_DIR = "/root/.node-red"
+NPM_LOCAL_INSTALL_DIR = "/usr/lib"
```
jan-kiszka, line 15: Same question as above - why not using `NODE_RED_HOME_DIR` as before?

recipes-core/images/iot2050-image-swu-example.bb
```diff
@@ -28,7 +28,6 @@ WICVARS += "WDOG_TIMEOUT KERNEL_IMAGE INITRD_IMAGE DTB_FILES"
 # not compatible with SWUpdate images
 IMAGE_INSTALL_remove = "regen-rootfs-uuid"
 IMAGE_INSTALL_remove = "install-on-emmc"
-IMAGE_INSTALL_remove = "node-red-preinstalled-nodes"
```
jan-kiszka, line 31: That's not that easy, at least as long as you do not disable upgrades from the UI via the palette. Those will strangely fail because they now work against a read-only rootfs. But that would still limit the usability of Node-RED as you can no longer add custom nodes while creating flows.

The idea we had with the Node-RED preinstallation is that users can update also the pre-installed nodes during runtime - that's what they are used to. However, this is naturally in conflict with image-based swupdate. We need to resolve that conflict in a smarter way or leave the issue alone.

recipes-app/node-red/files/node-red.service.tmpl
```diff
@@ -4,7 +4,9 @@ After=syslog.target network.target
 
 [Service]
 Type=idle
-ExecStart=/usr/bin/node-red -u %h/.node-red
+User=nodered
```
jan-kiszka, line 7: This is too early, we need to resolve #281 first. We are on it.

jan-kiszka (commented):
De-privileging Node-RED is the next logical step - but only after de-privileging mraa.

Where I have bigger doubts is prematurely suggesting that Node-RED and image-based system updates can go along already.

recipes-app/node-red-preinstalled-nodes/node-red-preinstalled-nodes_0.1.bb
fmoessbauer: My understanding was that the `node-red-preinstalled-nodes` might be modified by the user at runtime. If that is not the case, we better directly install into the homedir.

jan-kiszka: "Might" or rather "might **not** be modified"? Yes, they might be modified by users, specifically by triggering an update via the palette. That may even apply now to the GPIO node is it has been upstreamed.

recipes-app/node-red/files/node-red.service.tmpl
fmoessbauer: I'm not a node-red expert. According to the docs, the default is indeed `$HOME/.node-red`. By that, we could remove the `-u ...`. Don't know which other implications this has. https://nodered.org/docs/getting-started/local

recipes-core/images/iot2050-image-swu-example.bb
fmoessbauer: Ok, got it. However, this only means we still do not include it in the example image. We still can change the install paths so that node-red can be added to a custom swupdate image. We have that use-case and updatability support of the remaining platform is more important for us than Web-UI based updates of node-red modules.

At least as the install paths are now globally changeable in a consistent way, this leaves the decision up to the integrator.

recipes-app/node-red/files/node-red.service.tmpl
fmoessbauer: With the current pattern, we can still run the systemd service as root for now and simply change the node-red user dir to `${NODE_RED_HOME_DIR}`. Will change the commit accordingly.

recipes-app/node-red/node-red-package.inc
fmoessbauer: This is all about persistency. The preinstalled modules are installed on the RO rootfs (`/usr/lib/`), while the user installed ones are installed in `/var/lib/nodered/.node-red`, which is `$HOME/.node-red`.

recipes-app/node-red-preinstalled-nodes/node-red-preinstalled-nodes_0.1.bb
fmoessbauer: With the symlink in place, this file still cannot be modified. But if required by a user it can simply be replaced by a copy and from now on the user is responsible for maintaining state.

recipes-app/node-red/node-red-package.inc
jan-kiszka: But `/var/lib` is not writable in our setup as well, is it?

recipes-core/images/iot2050-image-swu-example.bb
jan-kiszka: I'm not against the other changes, and we could also discuss making the Node-RED palette read-only. Maybe there is a way to tell Node-RED that it cannot change nodes, thus provide useful feedback to users.

In general, users should model custom swupdate images in a way that pre-installs all nodes, not just those that we provide. OTOH, we did not test if our ad-hoc node packaging mechanism really works for most Node-RED nodes. I would not be surprised if at least some fail.

recipes-app/node-red/node-red-package.inc
fmoessbauer: It is, The `/var` partition is persistent: https://github.com/siemens/meta-iot2050/blob/master/wic/iot2050-swu.wks.in#L17

jan-kiszka: Ah, forgot about that.

Then the non-modifiable nodes should rather go into the location for Node-RED itself: `/usr/lib/node_modules/node-red/node_modules/`. I do not recall how to install built-in nodes there properly, but I do recall that the dashboard used to live there in older Node-RED releases, when it was still bundled.

jan-kiszka: Wait! We install node-red-gpio like this already as it was not upstreamed at that time. Maybe just flip the destination folder in swupdate mode?

recipes-core/images/iot2050-image-swu-example.bb
sn0wcat: Before reworking this please take also a look at this issue:
https://github.com/mindsphere/node-red-contrib-mindconnect/issues/173 

The pre-installed node-red is run for some reason from '/' directory and because the default setting of the mindconnect node is to use `process.cwd() +'/.mc/'` directory for the persistence, which  is now not writeable.

Also, I am not that sure that making node-red and palette read only is such a good idea - every single bugfix (e.g. in custom nodes) would require a new image instead of simple update via palette. This is very inconvenient and the users will probably bug the mainteiners of the nodes instead of you ;)

@fmoessbauer @jan-kiszka we should probably try to get on the call this week to discuss this ;)

jan-kiszka: Happy to explain the background, specifically which setups this is affecting and which not. The more I think about it, the less I believe we should promote manual clicking on the Node-RED palette in case the system is OTA-updated - this is a classic flaw to mix both.

jan-kiszka: Oh, and setting the WorkingDirectory in https://github.com/siemens/meta-iot2050/pull/395/commits/445d77af08aadb3af3a7174fd2122ce9988200f2 should also resolve the current startup issue of mindconnect in a read-only setup (or when we finally de-privilege it).

fmoessbauer: @sn0wcat thanks for cross-referencing this.

> The pre-installed node-red is run for some reason from '/' directory and because the default setting of the mindconnect node is to use process.cwd() +'/.mc/' directory for the persistence, which is now not writeable at all, a fatal error occurs during configuration of the mindsphere connectivity.

Normally applications should store their state either in `/var/` or for user-specific applications below `/home/<user>` (or more precisely `$HOME`). Using the current working directory is highly discouraged and this should be fixed in MCL. However, I don't know other node-modules do it. I guess there is a way to query the [userDir](https://nodered.org/docs/user-guide/runtime/configuration) and all state should be stored there.

> Also, I am not that sure that making node-red and palette read only is such a good idea - every single bugfix (e.g. in custom nodes) would require a new image instead of simple update via palette. 

The whole node-red story is a nightmare from a security perspective anyways. In the swupdate case, we need to ensure to have as little state as possible as otherwise configuration drift happens and updates get unstable. If you want to go that way, the default palette needs to be read-only. However, it still is possible to add more nodes from the palette, but these then have to be updated by the user. This again adds configuration drift with all its side effects.

recipes-app/node-red/node-red-package.inc
fmoessbauer: Moved to #397

recipes-core/customizations-example/files/board-configuration.service
fmoessbauer: Moved to https://github.com/siemens/meta-iot2050/pull/397

recipes-app/node-red/node-red-package.inc
fmoessbauer: Modeled as class.

recipes-app/node-red-preinstalled-nodes/node-red-preinstalled-nodes_0.1.bb
fmoessbauer: Changed install logic. Now we install the `node-red-preinstalled-nodes` as dedicated package into the ro-rootfs on swupdate.

recipes-app/node-red/node-red-package.inc
fmoessbauer: > Wait! We install node-red-gpio like this already as it was not upstreamed at that time. Maybe just flip the destination folder in swupdate mode?

Done. It can be switched by setting `NODE_RED_IMMUTABLE = "1"`

recipes-app/node-red/files/node-red.service.tmpl
fmoessbauer: Reverted. Now we still use the root user, but make things more configurable. Also `%h/.node-red` is replaced with `-u ${NODE_RED_HOME_DIR}/.node-red` to make it independent of the users home.

fmoessbauer:
To simplify the review, I carved out the trivial dependency fixes into #397 .

classes/node-red-module.bbclass
```diff
@@ -0,0 +1,19 @@
...
+#
+# Authors:
+#  Felix Moessbauer <felix.moessbauer@siemens.com>
+#
+# This file is subject to the terms and conditions of the MIT License.  See
+# COPYING.MIT file in the top-level directory.
+#
+
+inherit npm
+inherit node-red-user
```
jan-kiszka, line 12: This is the generic class, so 'node-red-user' does not belong here. Instead, add a node-red-module-updatable which build on top of this and makes sure the module gets installed into a writable location.

kas-iot2050-swupdate.yml
```diff
@@ -13,3 +13,7 @@ header:
 build_system: isar
 
 target: iot2050-image-swu-example
+
+local_conf_header:
+  node-red-immutable: |
+    NODE_RED_IMMUTABLE = "1"
```
jan-kiszka, line 19: In the current setup, node-red-gpio remains still immutable, so this is misleading.

recipes-app/node-red/node-red_2.2.3.bb
```diff
@@ -10,13 +10,17 @@
 #
 
 inherit npm
+inherit node-red-user
```
jan-kiszka, line 13: Node-RED is not an updatable package, is it? Why now suddenly?

jan-kiszka (commented):
Does not look ready yet.

recipes-app/node-red/node-red_2.2.3.bb
fmoessbauer: `node-red-user` does not mean it is an updatable package. It just means that this package installs / prepares something in the node-red user directory. In this case, this is the systemd service that needs to execute node-red from the user directory.

jan-kiszka: The node-red-user directory is supposed to be updatable, that's its whole purpose. Otherwise, you would install it into the node-red modules dir, just like node-red-gpio.

recipes-app/node-red-preinstalled-nodes/node-red-preinstalled-nodes_0.1.bb
fmoessbauer: With the new install pattern, `node-red-preinstalled-nodes` is now a simple debian meta-package, but not a NPM package anymore.

recipes-app/node-red/node-red-package.inc
fmoessbauer: I did some tests with @jan-kiszka and we concluded that this logic is no longer required. We can simply install the modules in `/usr/lib` and updates via the palette are still working (they are handled locally in the `userDir`). For details, have a look at 2476f8318837caa23624c11a3f7cd7605fe61e1d

recipes-core/images/iot2050-image-swu-example.bb
fmoessbauer: With the latest version, this now fully works (and updates via the palette are still possible).

recipes-app/node-red/node-red_2.2.3.bb
fmoessbauer: Moved logic to inc file to avoid confusion about naming scheme

kas-iot2050-swupdate.yml
fmoessbauer: Not required anymore

classes/node-red-module.bbclass
```diff
@@ -0,0 +1,22 @@
...
+require recipes-app/node-red/node-red-home.inc
+
+# common setting for all node-red-contrib modules
+DEBIAN_DEPENDS .= ", node-red"
+RDEPENDS += "node-red"
+
+# We need to explicitly set this directory to vendor the dependencies instead
+# of using NPMs "global" pattern that shares the dependencies.
+# This is required, because we do not model all transitive dependencies as
+# as individual debian packages. Modelling all all transitive dependencies as
```
jan-kiszka, line 20: duplicate "all"

recipes-app/node-red-node-serialport/node-red-node-serialport_1.0.1.bb
```diff
@@ -8,13 +8,12 @@
...
+inherit node-red-module
 inherit npm
 
 DESCRIPTION = "Node-RED nodes to talk to serial ports"
 
-NPM_LOCAL_INSTALL_DIR = "/root/.node-red"
-
 do_prepare_build_append() {
-    sed -i '/override_dh_install:/a\\trm -r ${PP}/image/root/.node-red/node_modules/node-red-node-serialport/node_modules/@serialport/bindings-cpp/prebuilds' \
+    sed -i '/override_dh_install:/a\\trm -r ${PP}/image/${@ d.getVar("NPM_LOCAL_INSTALL_DIR") or "/usr/lib" }/node_modules/node-red-node-serialport/node_modules/@serialport/bindings-cpp/prebuilds' \
```
jan-kiszka, line 17: We don't have a fallback for undefined NPM_LOCAL_INSTALL_DIR elsewhere, why here?

jan-kiszka (commented):
Just two small remarks remaining, looking good in general now.

recipes-app/node-red-node-serialport/node-red-node-serialport_1.0.1.bb
fmoessbauer: Ah yes... This is not required anymore. I'll remove it in the next version.

jan-kiszka (approved):

PR #398, opened 2023-01-04
linux-iot2050: dts: fix the gpio configuration of m.2

AsuraZeng:
For the nodes main_gpio0 and main_gpio1,
when we add the property pinctrl-0, it would override the previous property not to add to the previous node. So it would lack some configurations.

Add the property back to solve this issue.

### DIFF
--- a/recipes-kernel/linux/files/patches-5.10/0120-arm64-dts-ti-iot2050-Add-support-for-M.2-variant.patch
+++ b/recipes-kernel/linux/files/patches-5.10/0120-arm64-dts-ti-iot2050-Add-support-for-M.2-variant.patch
@@ -1,4 +1,4 @@
-From 2d9ecc4326b94c22dc4e71b359b4e0528e768aa7 Mon Sep 17 00:00:00 2001
+From 4d54ccf8bae2b39fc469912bdfef0eb5bda3f143 Mon Sep 17 00:00:00 2001
 From: chao zeng <chao.zeng@siemens.com>
 Date: Tue, 11 Jan 2022 11:28:24 +0800
 Subject: [PATCH 120/120] arm64: dts: ti: iot2050: Add support for M.2 variant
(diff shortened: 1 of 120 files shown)


recipes-kernel/linux/files/patches-5.10/0121-dts-fix-the-gpio-configuration.patch
```diff
@@ -0,0 +1,41 @@
+From 5a96b3a123e350b170c0eae82c84efdf7d0d3d1c Mon Sep 17 00:00:00 2001
+From: chao zeng <chao.zeng@siemens.com>
+Date: Tue, 3 Jan 2023 16:58:54 +0800
+Subject: [PATCH 121/121] dts: fix the gpio configuration
+
+For the nodes main_gpio0 and main_gpio1,
+when we add the property pinctrl-0, it would override the
+previous property not to add to the previous node.
+So it would lack some configurations.
```
jan-kiszka, line 9: What exactly would be broken?

Changes should be folded into the original commit adding the m2 dts file.

AsuraZeng: for the serial port(cp2102n), when switching the serial mode, it should be reset right now. so we use `switchserialmode` to switch the mode, it would not take effect right away.

jan-kiszka (commented):
There is a hard-to-review delta in recipes-kernel/linux/files/patches-5.10/0048-arm64-dts-ti-iot2050-Prepare-for-adding-2nd-generati.patch. Could you check your local git settings? I have these here:

```
[diff]
        renames = copies
        renameLimit = 10000

[difftree]
        renames = copies
```

AsuraZeng:
> There is a hard-to-review delta in recipes-kernel/linux/files/patches-5.10/0048-arm64-dts-ti-iot2050-Prepare-for-adding-2nd-generati.patch. Could you check your local git settings? I have these here:
> 
> ```
> [diff]
>         renames = copies
>         renameLimit = 10000
> 
> [difftree]
>         renames = copies
> ```

thanks, Have updated the patches.

jan-kiszka (approved):
Looks good. Will update the pending upstream patch as well - still waiting for a working upstream baseline there...

recipes-kernel/linux/files/patches-5.10/0120-arm64-dts-ti-iot2050-Add-support-for-M.2-variant.patch
```diff
@@ -107,7 +107,7 @@ index 000000000000..cb0f3a872920
 +
 +&main_gpio0 {
 +	pinctrl-names = "default";
-+	pinctrl-0 = <&main_m2_pcie_mux_control>;
++	pinctrl-0 = <&main_m2_pcie_mux_control &arduino_io_d4_to_d9_pins_default>;
```
jan-kiszka, line 110: Just noticed: Better follow the indention pattern below, now with two entries here.

AsuraZeng: I have considered it before. I have gone through dts of the kernel, and for most of the two entries' scenes, they wrote in one line.  if there have more entries, use multiple lines.

PR #405, opened 2023-01-12
Add missing RT deps for patch-u-boot-env

fmoessbauer:
cc @gylstorffq

### DIFF
--- a/recipes-bsp/patch-u-boot-env/patch-u-boot-env_0.1.bb
+++ b/recipes-bsp/patch-u-boot-env/patch-u-boot-env_0.1.bb
@@ -10,7 +10,8 @@ inherit dpkg-raw
 
 DESCRIPTION = "prepare u-boot environment for swupdate"
 
-DEBIAN_DEPENDS = "libubootenv-tool"
+RDEPENDS += "u-boot-${MACHINE}-config"
+DEBIAN_DEPENDS = "libubootenv-tool, u-boot-${MACHINE}-config"
 
 SRC_URI += "file://postinst \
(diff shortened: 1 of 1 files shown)


fmoessbauer:
@jan-kiszka please review

recipes-bsp/u-boot/u-boot-iot2050.inc
```diff
@@ -46,6 +46,10 @@ TEMPLATE_VARS += "FIRMWARE_SECURE_VER"
 
 U_BOOT_CONFIG_PACKAGE = "1"
 
+# break dependency chain by adding support for runtime provides / depends
+# TODO: once this is globally enabled in ISAR, remove this line
+RPROVIDES += "${PROVIDES}"
+
```
jan-kiszka, line 52: IIUC, we should wait for Isar to pick up the generic solution.

fmoessbauer: We now have the support in ISAR. Once we have the update here in meta-iot2050 as well, I will update this.
A similar change also needs to be done in `meta-coral` to reduce the number of components that are re-build on changes.

jan-kiszka:
You may already want to look into rebasing over #420.

fmoessbauer:
> You may already want to look into rebasing over #420.

Just tested that locally and can confirm it works. For Github I would like to wait till #420 is integrated to avoid including a lot of changes in the diff which might distract from the actual fix when reviewing.

jan-kiszka (approved):

PR #410, opened 2023-01-18
make example image package selections reusable

fmoessbauer:
This patch carves out the package selections from the example image. The selections provide great value for downstream users that do not want to base their image on the example image, but still want to avoid re-creating these selections to e.g. make wifi work.

### DIFF
--- a/recipes-core/images/iot2050-package-selections.inc
+++ b/recipes-core/images/iot2050-package-selections.inc
@@ -1,0 +1,103 @@
+#
+# Copyright (c) Siemens AG, 2023
+#
+# Authors:
+#  Su Baocheng <baocheng.su@siemens.com>
+#
+# This file is subject to the terms and conditions of the MIT License.  See
+# COPYING.MIT file in the top-level directory.
+#
+
+# debug tools
+IOT2050_DEBIAN_DEBUG_PACKAGES = " \
+    busybox \
+    bash-completion \
+    less \
+    vim \
+    psmisc \
+    bsdmainutils \
+    nano \
+    ifupdown \
+    iputils-ping \
+    ssh \
+    pciutils \
+    usbutils \
+    ethtool \
+    rt-tests \
+    stress-ng \
+    build-essential \
+    python3 \
+    python-is-python3 \
+    gawk \
+    curl \
+    wget \
+    ca-certificates \
+    resolvconf \
+    python3-newt \
+    gdb \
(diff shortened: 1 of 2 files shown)


fmoessbauer:
@BaochengSu : I'll keep you as the author of the package-selections file, as my only contribution was to copy the content.

jan-kiszka:
I vaguely recall a proposal by @AsuraZeng to have a more fine-grained feature selection, also via Kconfig. Don't find this anymore, but this here should be aligned while we are refactoring.

AsuraZeng:
> I vaguely recall a proposal by @AsuraZeng to have a more fine-grained feature selection, also via Kconfig. Don't find this anymore, but this here should be aligned while we are refactoring.


yes, my proposal at the time was to exclude the necessary software, divide other software according to function blocks and those could be selected through kconfig.

fmoessbauer:
Independent of the feature-selection concept, I still believe that this patch is valuable. Especially for downstream users it is much easier to use the provided variables directly. Also things like inspection and removal from these vars is possible, which is hard to achieve with a pure feature-selection based approach.

In short, we don't want to use the example image, but still do want to use pre-selected packages for the wifi.

jan-kiszka:
If there is no overlap or immediate conflict with the feature selection (as I said, I didn't find the changes anymore), then I'm fine with this step.

recipes-core/images/iot2050-package-selections.inc
```diff
@@ -0,0 +1,103 @@
...
+    libstdc++6:armhf \
+    "
+
+# docker support
+IOT2050_DEBIAN_DOCKER_PACKAGES = " \
+    docker.io \
+    docker-compose \
+    "
+# coral support
+IOT2050_CORAL_PACKAGES = " \
```
AsuraZeng, line 99: for coral support is from meta-coral. others are from debian repo. Can optimize the granularity of division and distinguish the source?

fmoessbauer: @AsuraZeng : Yes, that's true. But in the end the package selection is also curated by the meta-iot2050 project. By that, IMHO it makes more sense to name the variable `IOT2050_CORAL_PACKAGES`.

fmoessbauer:
@AsuraZeng any news on this topic? Is there something missing from my side?

AsuraZeng (approved):

PR #412, opened 2023-01-18
Fixes for iot2050-fimware-update

jan-kiszka:


recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
```diff
@@ -663,7 +666,7 @@ def main(argv):
         if args.force:
             # When forced, ignore all the checking and call the underlying updater directly
             updater = FirmwareUpdate(args.firmware)
-        if args.rollback:
+        elif args.rollback:
```
AsuraZeng, line 669: Actually, rollback and force update are mutexes.
this would not break force update.

It's better to modify like this

AsuraZeng: others LGTM.

AsuraZeng (approved):

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update
jan-kiszka: Without this fix, force mode won't work because it also tries the managed mode.

PR #421, opened 2023-03-01
iot2050setup: Fix reading back console size values and simplify code

jan-kiszka:
The current code was not compatible with consoles delivering the result of escape sequence in fragments or even byte-wise. Adjust things for that and also simplify the logic in way that we no longer need to switch to non-blocking mode.


AsuraZeng (approved):
Looks Fine.

PR #424, opened 2023-03-14
Update meta-coral to fix CI

BaochengSu:
update bazel-bootstrap to v4.2.2. The old jc-4.1.0 branch is removed from salsa repo.

### DIFF
--- a/kas/iot2050.yml
+++ b/kas/iot2050.yml
@@ -33,7 +33,7 @@ repos:
 
   meta-coral:
     url: https://github.com/siemens/meta-coral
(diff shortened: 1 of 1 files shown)


BaochengSu:
@fmoessbauer Need your review, thanks!

jan-kiszka:
Let's do #420 instead.

fmoessbauer:
> Let's do #420 instead.

This one is a trivial fix, available now. MR #420 might need more time to be integrated. I would also be fine with first integrating this and later on #420.

AsuraZeng (approved):
for the CI fix, this is fine.
Maybe we need to pick up the latest commit of the meta-coral at a certain time

PR #426, opened 2023-03-22
Secure Boot revision: Use RPMB as the key storage

BaochengSu:
This brings RPMB based key storage for secure boot, backed by StMM running in OPTee OS


jan-kiszka:
Inviting more to the party: @stormc @gylstorffq

recipes-app/iot2050-efivarfs-helper/files/postinst
```diff
@@ -0,0 +1,12 @@
...
+#
+# Authors:
+#  Su Baocheng <baocheng.su@siemens.com>
+#
+# This file is subject to the terms and conditions of the MIT License.  See
+# COPYING.MIT file in the top-level directory.
+#
+
+# mount efivarfs as read and write
+echo "efivarfs /sys/firmware/efi/efivars efivarfs rw,nosuid,nodev,noexec  0 0" >> /etc/fstab
```
jan-kiszka, line 12: Not sure where to leave this comment best: I rather think that the kernel still believes the EFI vars it sees are read-only, thus reports the efivarfs as read-only and mounts it as such by default. Did you instrument the related kernel path already (https://github.com/siemens/linux/blob/652b61c67ace146e76d570b34bea45559cbe6c0c/fs/efivarfs/super.c#L207)?

BaochengSu: Refer to:

```c
int efivar_supports_writes(void)
{
	return __efivars && __efivars->ops->set_variable;
}
```

I've checked neither the `__efivars` nor the `__efivars->ops->set_variable` is null after `tee_stmm_efi` is loaded.

Also if `__efivars->ops->set_variable` is null, the variable writing would be impossible.

However, since the `tee_stmm_efi` module is built as `M` and is loaded pretty lately, it could be possible that systemd asked for the writable flag before this module is loaded and cached the returned r/o flag inside, and never asked again after `tee_stmm_efi` is loaded.

jan-kiszka: Then maybe try to build that driver into the kernel next?

BaochengSu: tee_stmm_efi depends on tee-supplicant.service to do the job.  see the commit message from the `tee_stmm_efi` kernel patch:

```
This feature is implemented as a kernel module.
StMM PTA has TA_FLAG_DEVICE_ENUM_SUPP flag when registered to OP-TEE
so that this tee_stmm_efi module is probed after tee-supplicant starts,
since "SetVariable" EFI Runtime Variable Service requires to
interact with tee-supplicant.
```

Not sure if built-in could solve it, since anyway this module need be probed AFTER tee-supplicant starts, at that time the systemd mount setup is already done. However I will give it a try.

jan-kiszka: systemd dependencies between the service that starts tee-supplicant and the one that mounts efivarfs? Otherwise, report the problem now upstream.

BaochengSu:
@jan-kiszka @AsuraZeng @stormc @gylstorffq, any new comments?

recipes-app/optee-client/files/control.tmpl
```diff
@@ -0,0 +1,51 @@
+Source: optee-client
```
stormc, line 1: Hm, have you reported the issues here to [upstream Debian package](https://packages.debian.org/bookworm/optee-client-dev)? Seems that with this fixed there's no need to maintain this here, right?

recipes-app/iot2050-efivarfs-helper/files/iot2050-efivarfs-helper.tmpl
```diff
@@ -0,0 +1,180 @@
...
+
+def do_prepare():
+    message_prepare_fail = "Fail to prepare the efivarfs for manipulating. \
+You may need to check the environment."
+
+    try:
+        mod_required = 'tee_stmm_efi'
+        print("Checking for the kernel module {}... ".format(mod_required),
+              end='')
+        cmd = 'lsmod'
```
stormc, line 74: Hm, can't you just check that `/sys/modules/tee_stmm_efi` is an existing directory? Parsing is always a bit whacky and relies on output format stability.

recipes-app/iot2050-efivarfs-helper/files/iot2050-efivarfs-helper.tmpl
```diff
@@ -0,0 +1,180 @@
...
+        print("Done.")
+    except subprocess.CalledProcessError as e:
+        dump_subprocess_error(e)
+        exit_on_failure(message_prepare_fail, code=e.returncode)
+
+    # efitools provides convenient commands to manipulate the UEFI authenticated
+    # variables.
+    try:
+        print("Checking for the efitools pre-installation... ", end='')
+        cmd = 'which efi-readvar'
```
stormc, line 128: This may be better done with the included batteries: [shutil.which](https://docs.python.org/3.12/library/shutil.html#shutil.which)

recipes-app/iot2050-efivarfs-helper/files/iot2050-efivarfs-helper.tmpl
```diff
@@ -0,0 +1,180 @@
...
+
+        print("Done.")
+    except subprocess.CalledProcessError as e:
+        print("ERROR: service is inactive!")
+        exit_on_failure(message_prepare_fail)
+
+    # efivarfs must be mounted with read & write option.
+    try:
+        print("Checking for efivarfs mounting... ", end='')
+        cmd = 'findmnt {}'.format(EFIVARFS_MOUNT_POINT)
```
stormc, line 103: Maybe better done with the included batteries: [os.path.ismount](https://docs.python.org/3/library/os.path.html#os.path.ismount)

recipes-app/iot2050-efivarfs-helper/files/iot2050-efivarfs-helper.tmpl
```diff
@@ -0,0 +1,180 @@
...
+        result = subprocess.run(
+            cmd, shell=True, capture_output=True, check=True, text=True)
+        efivarfs_mount_error = 'not mounted.'
+        for line in result.stdout.split('\n'):
+            if line.startswith(EFIVARFS_MOUNT_POINT):
+                mount_options = line.strip().split(' ')[-1].split(',')
```
stormc, line 109: ... This can then be done with something along adapting the following lines:
```python
for item in (line.split() for line in open("/proc/mounts")
             if line.split()[1].startswith('/')):
    print(item)
```

recipes-app/iot2050-efivarfs-helper/files/iot2050-efivarfs-helper.tmpl
```diff
@@ -0,0 +1,180 @@
...
+    # By default the UEFI authenticated variable files in efivarfs have the
+    # immutable flag. This prevents from modifying it.
+    try:
+        change_efivarfs_immutable_flag(immutable=False)
+    except subprocess.CalledProcessError as e:
+        dump_subprocess_error(e)
+        exit_on_failure(message_prepare_fail, code=e.returncode)
+
+    print()
+    print('OK:\tNow you can use efitools to manipulate the UEFI authenticated variables')
```
stormc, line 147: Hm, why not extend this tool to call efitools with the appropriate flags so that +i flag gets removed, efi variable is modified, and then +i flag gets set again?

stormc:
LGTM in general. Is there an ETA on when the Kernel patches will be available from upstream Kernels? Then, we have to remove them here again...

jan-kiszka:
Upstream is more relevant right now from QA and from issue reporting (that r/o mounting problem...) perspective. There is unfortunately still a rather long road towards getting all our patches via a recent kernel. Even current LTS 6.1 is still missing icssg-eth.

recipes-app/optee-client/files/control.tmpl
BaochengSu: Do you mean the 0002.patch? I've upstreamed this patch to optee project , see https://github.com/OP-TEE/optee_client/commit/dd2d39b49975d2ada7870fe2b7f5a84d0d3860dc.

The target would be to use the upstream debian package provided by bookworm when debian 12 is released. However, currently there are still some todos, for example currently the debian upstream version use the RPMB emulation but not the real RPMB device, (see the RPMB_EMU=0 in the file `rules`), this might need introduce changes into linaro upstream for making it configurable via parameters. So at least for now, the debian upstream version could not be used directly.

recipes-app/iot2050-efivarfs-helper/files/iot2050-efivarfs-helper.tmpl
BaochengSu: Indeed, checking the file path existence would be a more easy and stable way.

recipes-app/iot2050-efivarfs-helper/files/iot2050-efivarfs-helper.tmpl
BaochengSu: Exactly.

recipes-app/iot2050-efivarfs-helper/files/iot2050-efivarfs-helper.tmpl
BaochengSu: Indeed.

AsuraZeng (approved):

PR #429, opened 2023-04-06
Update SEBoot to D01.03.01.02

BaochengSu:
This brings the flash read retries to work around occasionally reading error.


jan-kiszka:
What do we do with this?

BaochengSu:
This will be obsoleted by the upcoming new SEBoot version which supports the new hardware variants. Let's keep it open for now, and it will be mentioned and closed in the new PR.

BaochengSu:
Obsoleted by #509

PR #430, opened 2023-04-10
u-boot-iot2050: Make m.2 power control also avaiable to minipcie

BaochengSu:
Issue is that sometimes, some miniPCIE card hangs, due to firmware/software issue. Such hang could not be recovered by software reset, since it is not power on reset.

However, the power supply design on M.2 variant, that control the 3.3V power pin through a GPIO is proven to work against such scenario. With this GPIO, during boot, a POR will be triggered to fully recover the miniPCIE or M.2 module.

So port this design also to the miniPCIE variants.

The target branch is stable/V01.03


BaochengSu:
This MR is targeting only stable V1.3 branch, hence the kernel side change is not included. Instead, it is included into #431

BaochengSu:
Test passed on the new hardware sample.

huaqianli:
Looks good to me.

huaqianli (approved):

PR #431, opened 2023-04-10
Make m.2 power control also avaiable to minipcie

BaochengSu:
Replay #430 to master.

However the new FS number is not decided, that's why this PR is marked as draft, I will update it once the new FS number is fixed.

### DIFF
--- a/recipes-bsp/u-boot/files/0020-iot2050-Refactor-the-m.2-and-minipcie-power-pin.patch
+++ b/recipes-bsp/u-boot/files/0020-iot2050-Refactor-the-m.2-and-minipcie-power-pin.patch
@@ -1,0 +1,126 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Baocheng Su <baocheng.su@siemens.com>
+Date: Mon, 10 Apr 2023 13:25:47 +0800
+Subject: [PATCH] iot2050: Refactor the m.2 and minipcie power pin
+
+Make the m.2 power control pin also available on miniPCIE variants.
+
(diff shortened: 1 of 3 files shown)


recipes-bsp/u-boot/files/0019-iot2050-Refactor-the-m.2-and-minipcie-power-pin.patch
```diff
@@ -0,0 +1,109 @@
+From ae1bea40390ccd7682cbd017026510e7a610891f Mon Sep 17 00:00:00 2001
+From: Baocheng Su <baocheng.su@siemens.com>
+Date: Mon, 10 Apr 2023 13:25:47 +0800
+Subject: [PATCH] iot2050: Refactor the m.2 and minipcie power pin
+
+Make the m.2 power control pin also available on miniPCIE variants.
+
+This can fix some miniPCIE card hang issue, by forcing a power on reset
+during boot.
```
jan-kiszka, line 9: But only on new hardware variants, right?

BaochengSu: Yep. However the new FS number is not decided, that's why this PR is marked as draft, I want to update that information once the new FS number is fixed.

BaochengSu:
Kernel patch is ready now, actually it is the same as u-boot.

jan-kiszka (approved):
Could go in as is - or you already split up the U-Boot patch into two parts (1. DTS sync against the kernel, 2. board.c changes) as will be needed for upstream.

Issue #432, opened 2023-04-10
Use dedicated base-schroot for NPM package builds


fmoessbauer:
Since the introduction of sbuild in ISAR, all `do_dpkg_build` tasks run in their on isolated environment. The base rootfs of the builds is shared, but the build itself uses an overlay so that multiple builds can work isolated and concurrently.

For NPM modules, a lot of packages need to be installed prior to the build. These packages are needed for all NPM builds, but are only installed into the overlay of the schroot. By that, a ton of logic is executed in each build (which also needs to be emulated due to missing cross support). This is very inefficient and disk-bound.

Instead, it would be better to create a dedicated schroot with all base packages for the NPM builds. Then, the build itself only needs to install the package specific dependencies and runs much quicker.

As this is becoming a pattern, we should also consider a small ISAR patch to make it easier to create derived schroots from either the host or the target base schroot.

jan-kiszka:
Sure - already have patches?

fmoessbauer:
> Sure - already have patches?

Not yet. I just had a brief look at the `sbuild-chroot.inc` file and it looks quite straight forward to implement. I can give it a try and send it to the ISAR ML.

PR #433, opened 2023-04-13
feat: use dedicated schroot to speedup npm builds

fmoessbauer:
Let's wait till this has been integrated upstream.

closes: #432

### DIFF
--- a/classes/npm.bbclass
+++ b/classes/npm.bbclass
@@ -18,6 +18,7 @@ NPM_INSTALL_FLAGS ?= ""
 NPM_REBUILD ?= "1"
 
 ISAR_CROSS_COMPILE = "0"
+SBUILD_FLAVOR = "npm"
 
 SRC_URI = "npm://registry.npmjs.org;name=${NPMPN};version=${PV} \
     ${NPM_SHRINKWRAP}"
--- a/recipes-devtools/sbuild-chroot/sbuild-chroot-target-npm.bb
+++ b/recipes-devtools/sbuild-chroot/sbuild-chroot-target-npm.bb
@@ -1,0 +1,16 @@
+#
+# Copyright (c) Siemens AG, 2023
+#
+# Authors:
+#  Felix Moessbauer <felix.moessbauer@siemens.com>
+#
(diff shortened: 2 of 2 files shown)


BaochengSu:
@fmoessbauer How is the upstreaming going on?

BTW, now we are using `git format-patch --no-numbered --no-signature --zero-commit` for managing patch files :smirk:
 (Though it is not formal documented...)

jan-kiszka:
Just commented on the upstream series today: https://groups.google.com/g/isar-users/c/7c_r4LOfors/m/FpG93JhfAQAJ

fmoessbauer:
Let's just wait till this is merged upstream. Then I'll rebase this MR and ping you again.

jan-kiszka:
Feature was merged upstream. Will you rebase?

fmoessbauer:
> Feature was merged upstream. Will you rebase?

Rebased and updated to debian bookworm.

classes/npm.bbclass
```diff
@@ -18,6 +18,7 @@ NPM_INSTALL_FLAGS ?= ""
 NPM_REBUILD ?= "1"
 
 ISAR_CROSS_COMPILE = "0"
+SBUILD_FLAVOR = "${@ 'npm' if d.getVar('NPM_REBUILD') == '1' else '' }"
```
jan-kiszka, line 21: We always need npm, also when not rebuilding. We need python3 and libnode108 on top when rebuilding, see below.

fmoessbauer: Ok, but I want to avoid creating another schroot for the non-rebuild case. Shall we always run the build in the npm schroot?

jan-kiszka: Just build a schroot that contains everything we need to installing and also rebuilding npm. That can be used for all npm packages, even if they don't have anything to rebuild.

jan-kiszka:
This needs an update.

classes/npm.bbclass
fmoessbauer: Fixed by making in unconditional.

jan-kiszka (approved):

PR #434, opened 2023-04-17
add service to acquire hw information for swupdate

fmoessbauer:
This patch adds a one-shot systemd service that reads the hardware revision part from /proc/device-tree/compatible and writes that to /etc/hwrevision. By that, the hardware-compatibility node in the swupdate description file can be used to limit the boards an update can be applied to. This helps to avoid that images are applied to unsupported boards (e.g. due to lack of RAM, no M2, etc...).


jan-kiszka:
But this service is neither used nor otherwise described in meta-iot2050 right now, correct? I'd like to see a demo use case here as well so that user get an idea what it can be good for.

fmoessbauer:
The service is integrated into the swupdate image, so it is used. However, the data that is written to `/etc/hwrevision` is not being used yet. There are multiple issues in providing an example:

1. The corresponding entries in the sw-description file just recently were added to cip-core (they are in next currently, but not master)
2. What is considered compatible highly depends on the use-case
3. In case enable the `hardware-compatibility` node in the sw-description file, the update can no longer be applied via swupdate, as the corresponding data in `/etc/hwrevision` is not yet there.

That's why I would like to only integrate this service, but not enforce compatibility by now.

fmoessbauer:
> That leaves us with documenting this option at minimum.

Tell me where and I'll add it.

gylstorffq:
Also Hardware compatibility check is currently disabled when using SWUpdate 2021.11 from isar-cip-core. See https://gitlab.com/cip-project/cip-core/isar-cip-core/-/blob/master/recipes-core/swupdate/swupdate_2021.11-1%2Bdebian-gbp.bb#L32

jan-kiszka:
What's the status?

chombourger:
should this PR be updated? or closed?

gylstorffq:
It can be closed in my opinion as the value of `/etc/hwrevision` is set by https://github.com/siemens/meta-iot2050/blob/master/recipes-core/swupdate-config/swupdate-config_%25.bbappend

PR #435, opened 2023-04-25
iot2050-firmware-update: Refactor, bugfix and add some arguments

huaqianli:
Add `-q|--quiet` and `-d|--backup-dir`  arguments, refactor the exit code and fix some bugs.

### DIFF
--- a/recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.tmpl
+++ b/recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.tmpl
@@ -114,6 +114,9 @@ import io
 from enum import Enum
 from io import StringIO
 
+BOLD = "\033[1m"
+ENDC = "\033[0m"
+
 class UpgradeError(Exception):
     def __init__(self, ErrorInfo):
         super().__init__(self)
@@ -700,13 +703,13 @@ def main(argv):
         return ErrorCode.INVALID_ARG.value
 
     update_input = the_input(args.quiet, 'y',
-        "\nWarning: Update may render device unbootable. Continue (y/N)? ")
+        "\n" + BOLD + "Warning: Update may render device unbootable." + ENDC + " Continue (y/N)? ")
     if not update_input == "y":
         return ErrorCode.CANCELED.value
 
     if args.force or args.reset:
         erase_env_input = the_input(args.quiet, 'y',
-                "\nWarning: All U-Boot environment variables will be reset to factory settings. Continue (y/N)? ")
+                "\n" + BOLD + "Warning: All U-Boot environment variables will be reset to factory settings." + ENDC + " Continue (y/N)? ")
         if not erase_env_input == "y":
             return ErrorCode.CANCELED.value
     try:
@@ -739,12 +742,12 @@ def main(argv):
             for env in envlist:
                 print(env)
             preserved_env_input = the_input(args.quiet, 'y',
-                "\nWarning: Preserve the above env variables. Confirm(Y/n)? ")
+                "\n" + BOLD + "Warning: Preserve the above env variables." + ENDC + " Continue (y/N)? ")
             if not preserved_env_input == "n":
                 updater.update_uboot_env(envlist)
 
         reboot_input = the_input(args.quiet, 'n',
-            "\nWarning: Completed. Please reboot the device. Reboot now(y/N)? ")
+            "\n" + BOLD + "Warning: Completed. Please reboot the device." + ENDC + " Reboot now (y/N)? ")
         if not reboot_input == "y":
             return ErrorCode.SUCCESS.value
         os.system('reboot')
@@ -761,7 +764,7 @@ Hit the Enter Key to Exit:
         if not "No firmware image could be updated on this board" == str(e):
             # Rollback the firmware when the upgrade fails
             try:
-                print('Warning: Upgrade failed, rollback started')
+                print(BOLD + 'Warning: Upgrade failed, rollback started' + ENDC + '\n')
                 updater = RollbackFirmware(args.firmware, args.backup_dir)
                 if ErrorCode.ROLLBACK_FAILED.value == updater.rollback_firmware():
                     return ErrorCode.ROLLBACK_FAILED.value


recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.tmpl
```diff
@@ -121,12 +122,28 @@ class UpgradeError(Exception):
...
 
+#
+# | Error code  | Description                              |
+# |-------------|------------------------------------------|
+# | 0           | Success                                  |
+# | 1           | Failed to parse the arguments            |
+# | 2           | User canceled                            |
+# | 3           | Failed to backup firmware                |
+# | 4           | Failed to rollback the original firmware |
+# | 5           | Internal error                           |
```
jan-kiszka, line 133: The code is a good comment already IHMO. Better comment inline if some error code name is not verbose enough.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.tmpl
```diff
@@ -647,53 +668,71 @@ def main(argv):
...
+            preserved_env_input = the_input(args.quiet,
                 "\nWarning: Preserve the above env variables. Confirm(Y/n)? ")
             if not preserved_env_input == "n":
                 updater.update_uboot_env(envlist)
+
+        reboot_input = the_input(args.quiet,
+            "\nWarning: Completed. Please reboot the device. Reboot now(y/N)? ")
+        if not reboot_input == "y":
+            return ErrorCode.SUCCESS.value
+        os.system('reboot')
```
colpane, line 735: If I'm not incorrect, this will reboot the device after FW update is finished. Could we please disable this for quiet mode? It's better to not reboot the device automatically without user's recognition.

colpane (changes requested):

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.tmpl
huaqianli: Indeed, done.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.tmpl
```diff
@@ -660,6 +659,10 @@ def main(argv):
     parser.add_argument('-n', '--no-backup',
                         help='Do not generate a firmware backup',
                         action='store_true')
+    parser.add_argument('-d', '--backup-dir',
```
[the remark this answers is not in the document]
huaqianli: Already modified as `--backup-dir`.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.tmpl
```diff
@@ -647,53 +668,71 @@ def main(argv):
...
+            updater = FirmwareUpdate(args.firmware, args.bkup_dir)
         elif args.rollback:
-            updater = RollbackFirmware(args.firmware)
+            updater = RollbackFirmware(args.firmware, args.bkup_dir)
         else:
-            updater = ManagedFirmwareUpdate(args.firmware)
+            updater = ManagedFirmwareUpdate(args.firmware, args.bkup_dir)
 
         if args.rollback:
             rollback_confirm = input("\nRollback the firmware to the version before the upgrade (y/N)?")
```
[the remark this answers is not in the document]
huaqianli: I have corrected it, I missed this one when I did the `rebase`.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.tmpl
huaqianli: Yes, this script will reboot the device automatically with `quiet` mode. Hi @jan-kiszka @BaochengSu @AsuraZeng  any comments here?

AsuraZeng: The `quiet mode` is that everything is selected by default, reboot is also the selected behavior.  So from my perspective, reboot with the quiet mode is ok.

BaochengSu: Quiet mode is mainly used in script, in that case, normally user also want to scripting-ly control the system behavior, so sounds to me, a non-reboot should be the default behavior under quiet mode, so that user can control the system reboot on their demands.

Also, before the reboot, user may want to collect the upgrading logs for further analysis, and also may want to use USER_LED to indicates the upgrading result, since in this case there is no way (FTDI, DP) to indicates the result other than LED.

So a default no-reboot sounds more reasonable to me.

But user needs to be aware of the risks of not immediately reboot, if some u-boot env manipulation happens before reboot, the firmware may corrupt.

Another way is to keep the reboot as the default quiet action, but also add a `no-reboot` parameter, but this would also introduce to handling the special case of combining the `no-reboot` with non-quiet mode upgrading, sounds not a elegant way.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.tmpl
```diff
@@ -121,6 +122,13 @@ class UpgradeError(Exception):
     def __str__(self):
         return self.erroinfo
 
+class ErrorCode(Enum):
+    SUCCESS = 0
+    INVALIDARG  = 1
+    CANCELED  = 2
+    BACKUPFAILED  = 3
+    ROLLBACKFAILED = 4
+    INTERNALERR = 5
```
BaochengSu, line 131: Better to use `_` to make it more clear, such as `INVALID_ARG`, `ROLLBACK_FAILED`

BaochengSu:
Otherwise LGTM.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.tmpl
```diff
@@ -670,6 +669,10 @@ def main(argv):
     parser.add_argument('-n', '--no-backup',
                         help='Do not generate a firmware backup',
                         action='store_true')
+    parser.add_argument('-d', '--bkup-dir',
```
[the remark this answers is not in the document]
karadaganil: +1

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.tmpl
karadaganil: > If I'm not incorrect, this will reboot the device after FW update is finished. Could we please disable this for quiet mode? It's better to not reboot the device automatically without user's recognition.

+1

huaqianli: @colpane @BaochengSu @AsuraZeng @karadaganil  Is there any more comments here?

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.tmpl
```diff
@@ -710,12 +737,12 @@ def main(argv):
...
                 print(env)
-            preserved_env_input = input(
+            preserved_env_input = the_input(args.quiet, 'y',
                 "\nWarning: Preserve the above env variables. Confirm(Y/n)? ")
             if not preserved_env_input == "n":
                 updater.update_uboot_env(envlist)
 
-        reboot_input = input(
+        reboot_input = the_input(args.quiet, 'n',
             "\nWarning: Completed. Please reboot the device. Reboot now(y/N)? ")
```
BaochengSu: Refine the commit message, now not `all` answers yes.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.tmpl
```diff
@@ -663,19 +684,24 @@ def main(argv):
                         help='Specify the firmware backup directory',
                         nargs=1,
                         default=os.getenv('HOME'))
+    parser.add_argument('-q', '--quiet',
+                        help='Update firmware quietly, assume "yes" as answer to \
+                        all promptas except reboot is "no" and run non-interactively',
```
BaochengSu, line 692: Tell  the user that this would not auto reboot the device after upgrading.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.tmpl
huaqianli: `except reboot is 'n'` is telling this mesage.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.tmpl
huaqianli: except reboot is 'n' is telling this message. You were suggesting to write it more detailed?

BaochengSu: s/promptas/prompts/

BaochengSu: > `except reboot is 'n'` is telling this message. You were suggesting to write it more detailed?

I think it's better to describe the behavior from user perspective, instead of from the tool perspective, so I would write something like `Update firmware quietly with the old firmware auto backed up and rollback in case of failure, and without automatic rebooting.`

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.tmpl
karadaganil: Hi @huaqianli,

Here is my understanding;
   if  quiet mode is enabled, all questions(excluded reboot) are "Yes".

From my point of view, "rollback_confirm" should be "No" since "-r|--rollback" is provided by the tool + platform builder cannot know the actual operation state since rollback_failed error code is just sent once no valid backup file found*. rollback seems perform like fimware update. So, the error code is not raised by update_firmware(included flash_operate()) method.


* https://github.com/huaqianli/meta-iot2050/blob/master/recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.tmpl#L601

karadaganil: I strongly suggest using of other cli arguments to separate rollback and reboot performing configs from "quiet" mode.

--auto-reboot      to specify firmware auto reboot configuration 
--auto-rollback   to specify firmware auto rollback configuration

colpane: @huaqianli Please, comment on @karadaganil's comment. It is needed for us.

BaochengSu: What about we don't add `--auto-reboot` now and let the quiet mode not reboot the device?

My understanding is that auto reboot does not bring too much value to the quiet mode. Even if people really use auto-reboot, then a simple combination of `iot2050-fw-update --options && reboot` will do the trick.

BaochengSu: Regarding the rollback in quiet mode, what about to add a new return code, for example, named with `Rollback_Happened`, with non-zero return code, to indicates updating failed and rollback happened?

Or do you rather want to use the `-q` with no auto rollback and do the rollback in your own script / manually depends on your use case?

I am 50 50.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.tmpl
```diff
@@ -702,17 +758,20 @@ def main(argv):
...
+            print(BOLD + 'Warning: Upgrade failed, rollback started' + ENDC + '\n')
+            ret = updater.rollback_firmware()
+            if ErrorCode.ROLLBACK_FAILED.value == ret:
+                return ret
         except UpgradeError:
             print("Rollback failed")
 
-        input(intput_reminder)
-
+        the_input(False, intput_reminder)
```
colpane, line 772: Why do not we use args.quiet instead of False? Also, what is INTERNAL_ERR? Is it an error during flashing process? I think we should use a more descriptive name.

colpane (changes requested):

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.tmpl
```diff
@@ -647,53 +683,73 @@ def main(argv):
...
         if args.rollback:
-            rollback_confirm = input("\nRollback the firmware to the version before the upgrade (y/N)?")
+            rollback_confirm = the_input(args.quiet, 'y',
+                    "\nRollback the firmware to the version before the upgrade (y/N)?")
             if not rollback_confirm == "y":
-                sys.exit(1)
-            updater.rollback_firmware()
-            sys.exit(1)
+                return ErrorCode.CANCELED.value
+            ret = updater.rollback_firmware()
```
colpane, line 724: What if an UpgradeError is thrown during update_firmware method? The code will jump to line 753 and try the rollback again? What will be the return code at the end, INTERNAL_ERR?

colpane (changes requested):

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.tmpl
```diff
@@ -702,17 +758,20 @@ def main(argv):
...
+            if ErrorCode.ROLLBACK_FAILED.value == ret:
+                return ret
         except UpgradeError:
             print("Rollback failed")
 
-        input(intput_reminder)
-
+        the_input(False, intput_reminder)
+        return ErrorCode.INTERNAL_ERR.value
```
colpane, line 773: We need to separate rollback or main binary flashing errors. Please recheck the flow of code for upgrading process for both normal and rollback parts.

colpane (changes requested):

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.tmpl
```diff
@@ -121,12 +125,18 @@ class UpgradeError(Exception):
     def __str__(self):
         return self.erroinfo
 
+class ErrorCode(Enum):
+    SUCCESS = 0
+    INVALID_ARG  = 1
+    CANCELED  = 2
+    BACKUP_FAILED  = 3
+    ROLLBACK_FAILED = 4
+    INTERNAL_ERR = 5
```
colpane, line 134: Please rename it for better description.

colpane (changes requested):

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.tmpl
colpane: If we provide False, then the code will still wait for an input in quiet mode too.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.tmpl
colpane: Also, any problem during the flashing process should return the fail return code.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.tmpl
colpane: I think, maybe we should have two different exception. UpgradeError for normal process and UpgradeRollbackError for rollback process :)

jan-kiszka:
I was looking at https://support.industry.siemens.com/forum/ww/en/posts/update-firmware-v1-3-1-1/299468/?page=0&pageSize=10, noticing at least 3 improvables:
- no clear reason given why the update is not applicable (most critical issue)
- "Upgrade Failed! Please do not reboot!!" - as no update was applied, apparently, this warning is misleading
- We should check compatibility first and, if given, then ask for the ok to apply the update, not the other way around like now

Are these issues already in scope for the merge request here?

huaqianli:
> * no clear reason given why the update is not applicable (most critical issue)
> * "Upgrade Failed! Please do not reboot!!" - as no update was applied, apparently, this warning is misleading
> * We should check compatibility first and, if given, then ask for the ok to apply the update, not the other way around like now
@jan-kiszka  There is some improvement here, some of them still act as before.

BaochengSu:
> > * no clear reason given why the update is not applicable (most critical issue)
> > * "Upgrade Failed! Please do not reboot!!" - as no update was applied, apparently, this warning is misleading
> > * We should check compatibility first and, if given, then ask for the ok to apply the update, not the other way around like now
> 
> @jan-kiszka There is some improvement here, some of them still act as before.

Why not fix them in this MR as well? In general, no big change required.

huaqianli:
> > > * no clear reason given why the update is not applicable (most critical issue)
> > > * "Upgrade Failed! Please do not reboot!!" - as no update was applied, apparently, this warning is misleading
> > > * We should check compatibility first and, if given, then ask for the ok to apply the update, not the other way around like now
> > 
> > 
> > @jan-kiszka There is some improvement here, some of them still act as before.
> 
> Why not fix them in this MR as well? In general, no big change required.

From my perspective, it is not a small effort especially go with the current PR, If we try to fix all issues in this PR, it will be potentially unstopped. Actually, there is already many unexpected changes to be added in this PR. 

If we do want to improve it to the next level, I'd prefer a new PR to do that.

huaqianli:
**A record here:**
As discussed offline, Colpan was okay with the improvement based on his comment, he mainly cares about how the tool behavior when an error or exception occurs. Especially the return value should be as expected.

Let's go with an example:
The tool will return FLASHING_ERR(5) if it fails to upgrade and rollback successfully.

PR #444, opened 2023-05-23
iot2050-example-image: Fix wrongly installed openssl armhf

BaochengSu:
The openssl (target version, arm64) should be installed into the example image, not the openssl-compat (armhf version).

Meanwhile integrate the libssl1.1-compat for armhf compatible architecture.

Test result:

```shell
root@iot2050-debian:~# apt list --installed | grep ssl

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

libssl-dev/now 1.1.1n-0+deb11u4+iot2050 arm64 [installed,local]
libssl1.1/now 1.1.1n-0+deb11u4+iot2050 arm64 [installed,local]
libssl1.1/now 1.1.1n-0+deb11u4+iot2050 armhf [installed,local]
openssl/now 1.1.1n-0+deb11u4+iot2050 arm64 [installed,local]
```

### DIFF
--- a/recipes-core/images/iot2050-image-example.bb
+++ b/recipes-core/images/iot2050-image-example.bb
@@ -28,7 +28,8 @@ IMAGE_PREINSTALL += "${@ ' \
     ' if d.getVar('IOT2050_DOCKER_SUPPORT') == '1' else ''}"
 
 IMAGE_INSTALL += " \
-    openssl-compat \
+    libssl1.1-compat \
+    openssl \
     expand-on-first-boot \
     sshd-regen-keys \
     regen-rootfs-uuid \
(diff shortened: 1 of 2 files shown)


jan-kiszka:
This is very likely broken as now nothing is requesting the building of the compat packages. We should rather add a
```
PROVIDES += "libssl1.1"
```
to the openssl recipe and then request
```
IMAGE_INSTALL += "libssl-1.1-compat openssl"
```
Make sure to validate all of this via clean rebuilds.

BaochengSu:
> This is very likely broken as now nothing is requesting the building of the compat packages. We should rather add a
> 
> ```
> PROVIDES += "libssl1.1"
> ```
> 
> to the openssl recipe and then request
> 
> ```
> IMAGE_INSTALL += "libssl-1.1-compat openssl"
> ```
> 
> Make sure to validate all of this via clean rebuilds.

Yep, this is indeed a more safer way to avoid possible future brokens.

Refined to adapt this way.

jan-kiszka (approved):

Issue #427, opened 2023-03-28
Kernel OOPS while using rmmod for ethernet driver


sgthammer:
Hello,

For an image with swupdate support, I tried to remove ethernet driver module using _rmmod_. "**rmmod iccsg_prueth**" command is used for this aim and the _rmmod_ throws segmentation fault. Also, kernel throws OOPS as internal error. Kernel log file is attached.

What could be the reason causes this issue?
[oops.log](https://github.com/siemens/meta-iot2050/files/11090619/oops.log)

nhahnetsrot:
Pls. try the attached patch, needs to be copied to meta-iot2050/recipes-kernel/linux/files/patches-5.10/ and then the image needs to be rebuild (e.g.: sudo kas-container clean; kas-container build )

[0211-fix-rmmod.patch](https://github.com/siemens/meta-iot2050/files/11551349/0211-fix-rmmod.patch)

jan-kiszka:
Will be fixed by #491

PR #445, opened 2023-05-24
Fix kernel driver crash of icssg_prueth driver upon module removal

nhahnetsrot:
Fixes the crash reported in https://github.com/siemens/meta-iot2050/issues/427 . 

Allows now the removal / reload of the driver, network devices, links etc. are restored and working.

### DIFF
--- a/recipes-kernel/linux/files/patches-5.10/0211-fix-rmmod.patch
+++ b/recipes-kernel/linux/files/patches-5.10/0211-fix-rmmod.patch
@@ -1,0 +1,30 @@
+--- linux-cip-5.10.162-cip24/drivers/net/ethernet/ti/icssg_prueth.c.orig	2023-05-08 16:22:22.805004920 +0200
++++ linux-cip-5.10.162-cip24/drivers/net/ethernet/ti/icssg_prueth.c	2023-05-24 06:39:39.072977487 +0200
+@@ -2082,7 +2082,13 @@ static int emac_ndo_stop(struct net_devi
+ 	netif_tx_stop_all_queues(ndev);
+ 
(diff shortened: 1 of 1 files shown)


jan-kiszka:
Thanks for looking into that! Still, it looks like there are still some todos remaining:
 - resolve the open questions on the fixing patch itself - we really need to understand the issue when trying to fix it
 - write the kernel patch like the patch adding it: subject, commit message, signed-off

Keep in mind that, mid-term, we would like to get rid of the patch again by sharing our findings with TI and making them adopt them, directly or at least logically. Also for them, we need to explain why we did those changes based on the code, not just the observed stability improvement.

nhahnetsrot:
> Thanks for looking into that! Still, it looks like there are still some todos remaining:
> 
> * resolve the open questions on the fixing patch itself - we really need to understand the issue when trying to fix it
> * write the kernel patch like the patch adding it: subject, commit message, signed-off
> 
> Keep in mind that, mid-term, we would like to get rid of the patch again by sharing our findings with TI and making them adopt them, directly or at least logically. Also for them, we need to explain why we did those changes based on the code, not just the observed stability improvement.

TI already knows about this bug ( i emailed with Danish Anwar <a0501179@ti.com>, he ist responsible for the driver at TI side). He stated that a "cleaned up" version of the driver does not has this issue and will be going into the official kernel "soon". Then this patch is no longer required.
(he did not define "soon" ...)

jan-kiszka:
Will we get an update? Or will this bug have to wait for replacing the patches with their upstream versions?

jan-kiszka:
Obsoleted by #491

PR #446, opened 2023-05-31
iot2050-firmware-update: Add making FWU package option

huaqianli:
Firmware update package is good to generate within building time, so add a image type to make the FWU package.

### DIFF
--- a/recipes-core/images/iot2050-image-example.bb
+++ b/recipes-core/images/iot2050-image-example.bb
@@ -28,7 +28,8 @@ IMAGE_PREINSTALL += "${@ ' \
     ' if d.getVar('IOT2050_DOCKER_SUPPORT') == '1' else ''}"
 
 IMAGE_INSTALL += " \
-    openssl-compat \
+    libssl1.1-compat \
+    openssl \
     expand-on-first-boot \
     sshd-regen-keys \
     regen-rootfs-uuid \
--- a/recipes-security/openssl/openssl_latest.bb
+++ b/recipes-security/openssl/openssl_latest.bb
@@ -16,6 +16,8 @@ SRC_URI = " \
     "
 CHANGELOG_V="<orig-version>+iot2050"
 
+PROVIDES += "libssl1.1"
+
 DEB_BUILD_OPTIONS += "nocheck"
 
 do_prepare_build() {


kas-iot2050-fwu-package.yml
```diff
@@ -0,0 +1,18 @@
...
+#
+
+header:
+  version: 10
+  includes:
+    - kas/iot2050.yml
+
+build_system: isar
+
+target: firmware-package-tool
```
jan-kiszka, line 18: "tool"? Or rather "firmware-update-package"?

recipes-devtools/firmware-package-tool/firmware-package-tool.bb
```diff
@@ -0,0 +1,24 @@
...
+
+inherit dpkg-raw
+
+do_deploy_deb[nostamp] = "1"
+
+do_deploy_deb() {
+    # Generate the firmware update package
+    sh ${WORKDIR}/iot2050-fwu-update-json.sh ${WORKDIR} ${DEPLOY_DIR_IMAGE} $(${ISAR_RELEASE_CMD})
+}
+do_deploy_deb[dirs] = "${DEPLOY_DIR_IMAGE}"
```
jan-kiszka, line 24: This is missing dependencies: If you only build this package, the firmware artifacts won't be there.

recipes-devtools/firmware-package-tool/files/update.conf.json
```diff
@@ -0,0 +1,34 @@
...
+        "SIMATIC IOT2050-BASIC",
+        "SIMATIC IOT2050 Basic",
+        "SIMATIC IOT2050-ADVANCED",
+        "SIMATIC IOT2050 Advanced"
+      ]
+    },
+    {
+      "description": "IOT2050 PG2 Bootloader Release V01.01.01",
+      "name": "iot2050-pg2-image-boot.bin",
+      "version": "V01.01.01",
```
jan-kiszka, line 17: Version must be generated, likely derived from `git describe` (see `ISAR_RELEASE_CMD`).

recipes-devtools/firmware-package-tool/firmware-package-tool.bb
```diff
@@ -0,0 +1,24 @@
...
+# This file is subject to the terms and conditions of the MIT License.  See
+# COPYING.MIT file in the top-level directory.
+#
+DESCRIPTION = "Generate The Firmware Update Package"
+MAINTAINER = "huaqian.li@siemens.com"
+
+SRC_URI = "file://update.conf.json \
+           file://iot2050-fwu-update-json.sh"
+
+inherit dpkg-raw
```
jan-kiszka, line 16: You are not building a deb package, are you?

recipes-devtools/firmware-package-tool/firmware-package-tool.bb
```diff
@@ -0,0 +1,24 @@
...
+
+do_deploy_deb[nostamp] = "1"
```
jan-kiszka, line 18: Why "nostamp"?

Kconfig
```diff
@@ -50,6 +51,13 @@ config IMAGE_BOOT_PG2
 	  that it fits AND you have an external flash programmer at hand that
 	  allows to recover. Otherwise you risk to BRICK THE IOT2050!
 
+config IMAGE_FWU_PKG
+	bool "Firmware update package for all devices"
+	help
+	  This is the official firmware update package in IOT2050-FWU-PKG-Vx.x.x.tar.xz
+	  format, which includes firmware bin, and update criteria, and builtin env,
+	  and it is for all IOT2050 devices.
```
jan-kiszka, line 59: I still have on my to-do list to refactor the firmware build again. Upstream TI folks suggested to generate both artifacts with a single binman (U-Boot) run. Then we could just append the update package generation to the firmware build and don't have to introduce a separate config option.

jan-kiszka:
The general idea is valuable! We just need to sort out the implementation details to make this a smooth thing we can also use during the release.

kas-iot2050-fwu-package.yml
huaqianli: ACK.

recipes-devtools/firmware-package-tool/files/iot2050-fwu-update-json.sh
```diff
@@ -0,0 +1,53 @@
...
+{
+    sed -i '/"version": ".*"/s/"V.*"/"'$2'"/g' $1
+    sed -i '/"description": ".*"/s/V.*["$]/'$2\"'/g' $1
+}
+
+generate_fwu_tarball()
+{
+    echo "Generating the firmware tarball..."
+
+    if [ ! -e $2/iot2050-pg1-image-boot.bin ]; then
```
AsuraZeng, line 22: Anyway we should build the pg1 and pg2 uboot previously. 
this is not a full automatic build flow.
Could it be done through our pipeline?

recipes-devtools/firmware-package-tool/files/update.conf.json
huaqianli: This is raw json, which provides formatting. The final version value will be replaced by ISAR_RELEASE_CMD. Please check [iot2050-fwu-update-json.sh ](https://github.com/siemens/meta-iot2050/pull/446/commits/95c6a370ae3e0e65bba2bfc8985f334230e19696#diff-0c9ec4ecf7b23e3159a1bac613fc1112d4338a917ec3698a6d00d71944214b50).

recipes-devtools/firmware-package-tool/firmware-package-tool.bb
huaqianli: Yes, I didn't find a more elegant way. If you know, please suggest me, thanks.

recipes-devtools/firmware-package-tool/firmware-package-tool.bb
huaqianli: I was thinking that it is good to generate this tarball each building.

recipes-devtools/firmware-package-tool/files/iot2050-fwu-update-json.sh
huaqianli: Agreed, to build the pg1 and pg2 previously is better. 

Here is a workaround after internal discussion. Since @jan-kiszka  is also not agreed with this solution, I guess we need to discuss more.

Kconfig
huaqianli: Exactly, from my perspective, the solution you mentioned will be the best way.

recipes-devtools/firmware-package-tool/files/iot2050-fwu-update-json.sh
```diff
@@ -0,0 +1,53 @@
...
+    cp $2/iot2050-pg1-image-boot.bin $2/.tarball
+    cp $2/iot2050-pg2-image-boot.bin $2/.tarball
+    cp $2/u-boot-initial-env $2/.tarball
+
+    cd $2/.tarball 
+    tar -cJvf $2/IOT2050-FWU-PKG-$3.tar.xz *
+    cd - && rm -rf $2/.tarball
+}
+
+generate_fwu_tarball $*
```
BaochengSu, line 53: The json validation should be performed before tar.

Python script would be more suitable, since it has built-in json and tar support.

recipes-devtools/firmware-package-tool/files/update.conf.json
```diff
@@ -0,0 +1,34 @@
...
+    {
+      "type": "Example Image",
+      "key": "BUILD_ID",
+      "min_version": "V01.01.01"
+    }
+  ],
+  "suggest_preserved_uboot_env": [
+    "boot_targets"
+  ]
+}
```
BaochengSu, line 34: Long-term speaking, this should be replaced by the CLI interface.

This json is more for an internal implementation details/choice, not an interface.

But I am ok with it for the 1st version of this packaging tool.

Kconfig
huaqianli: @jan-kiszka I will leave this PR here until you finish your work, then I will see what I can do.

recipes-devtools/firmware-package-tool/firmware-package-tool.bb
jan-kiszka: Another reason to fix the u-boot build first - good to know. In any case, this workaround is not nice at all.

recipes-devtools/firmware-package-tool/firmware-package-tool.bb
jan-kiszka: If you are not building a deb package, don't inherit a class that tries to do that. But as I said, this path is not the best one anyway.

recipes-devtools/firmware-package-tool/firmware-package-tool.bb
jan-kiszka: ...which is a bug.

PR #449, opened 2023-06-03
Automate the RPMB key pairing

BaochengSu:
Before some manual steps were required under u-boot console to trigger
the RPMB key provisioning, this was not so friendly for both the users
and the factory during manufactoring.

With this change, manual trigger is never required. Now by booting the
special firmware, the RPMB key is auto provisioned. And the result could
be checked both from frimware booting log and from linux `mmc` command.

Although, with a well equiped kernel image - with CONFIG_TEE_STMM_EFI or
CONFIG_TCG_FTPM_TEE enabled, it is also possible to auto provision the
RPMB key by a simple booting of that OS image, this brings extra
dependency on the OS image. It's better for the bootloader to finish
this task by its own, which benifits for some operation envioronment,
such as the manufactoring.

### DIFF
--- a/recipes-bsp/u-boot/README.md
+++ b/recipes-bsp/u-boot/README.md
@@ -95,9 +95,33 @@ key, together with a special U-Boot binary for kicking off the key programming.
 > To protected the programmed RPMB key, it is required to flash a signed image,
 > program secure boot keys and enable secure boot.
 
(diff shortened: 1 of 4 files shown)


jan-kiszka (approved):
Look good to me. I was about to test it on real device but then managed to accidentally program the key otherwise...

PR #452, opened 2023-06-10
Rework non secure firmware

jan-kiszka:
Non-secure UEFI booting was broken, and we should also avoid the TA splats by not putting StandaloneMM and fTPM into the non-secure firmware.


recipes-bsp/optee-os/optee-os-iot2050_3.21.0.inc
```diff
@@ -21,8 +21,12 @@ OPTEE_PLATFORM = "k3-am65x"
...
-    CFG_CONSOLE_UART=1 CFG_RPMB_FS=y CFG_RPMB_FS_DEV_ID=1 CFG_CORE_DYN_SHM=y \
-    CFG_IN_TREE_EARLY_TAS=avb/023f8f1a-292a-432b-8fc4-de8471358067 \
+    CFG_CONSOLE_UART=1 CFG_CORE_DYN_SHM=y \
     CFG_WARN_INSECURE=n \
     CFG_REE_FS=n CFG_CORE_HEAP_SIZE=524288 \
     "
+
+OPTEE_EXTRA_BUILDARGS:append:secureboot = " \
+    CFG_RPMB_FS=y CFG_RPMB_FS_DEV_ID=1 \
+    CFG_IN_TREE_EARLY_TAS=avb/023f8f1a-292a-432b-8fc4-de8471358067 \
```
BaochengSu, line 31: IIRC, avb is not related to secureboot, but rpmb pairing.

The rpmb pairing command 'optee_rpmb' is taking advantage of the avb persistent variable which depends on the secure storage of optee - in our case, RPMB.

recipes-bsp/u-boot/files/0017-spl-fit-Report-fdt-error-for-loading-u-boot.patch
```diff
@@ -17,7 +17,7 @@ Signed-off-by: Baocheng Su <baocheng.su@siemens.com>
  1 file changed, 6 insertions(+), 2 deletions(-)
 
 diff --git a/common/spl/spl_fit.c b/common/spl/spl_fit.c
-index a35be52965..00404935cb 100644
+index a35be529656..00404935cb2 100644
```
BaochengSu, line 20: I think maybe it's better to drop these meaningful changes.

BaochengSu (commented):
Otherwise LGTM.

recipes-bsp/optee-os/optee-os-iot2050_3.21.0.inc
jan-kiszka: Right, we need only to permit some RPMB write from u-boot command line. I'll switch that piece to `rpmb-setup`. Or we write an EFI variable from U-Boot instead.

recipes-bsp/u-boot/files/0017-spl-fit-Report-fdt-error-for-loading-u-boot.patch
jan-kiszka: Means we need to pin to `--abbrev=10` - or switch to `--full-index` so that we no longer have to agree on the length.

BaochengSu: I am ok to both options, as long as we have an unified command set to generate the patch set, to make sure people do not step on each other.

I think we can put the command in somewhere for later reference, what about the wiki page, or the README.md?

BaochengSu (approved):

PR #454, opened 2023-06-16
Fixes and cleanups for SWUpdate image

jan-kiszka:
@gylstorffq


recipes-core/customizations-swupdate/files/swupdate.cfg
```diff
@@ -11,4 +11,5 @@
 globals :
 {
     bootloader = "ebg";
+    mtd-blacklist = "0 1 2 3 4 5 6 7";
```
BaochengSu, line 14: Other than the u-boot env partitions, what else benefits could we get from the mtd partition? If not, I prefer to only keep the mtd partitions for u-boot envs, that could save us from some trouble if the flash layout is changed in the future.

BaochengSu:
LGTM

recipes-core/customizations-swupdate/files/swupdate.cfg
jan-kiszka: We don't perform any firmware updates on the mtd partitions, only work against the mass storage in this integration. So there is no value in probing it. We can change that if we ever find a use case.

Issue #453, opened 2023-06-15
Porting to Debian 12 (bookworm)


jan-kiszka:
Baseline: https://github.com/siemens/meta-iot2050/commits/jan/bookworm

Todos:
 - rebase over master
 - drop openssl patch
 - make npm packages build (see f015adfd8f36817edef5998b5763f2c7ee8fcbb4)

@nhahnetsrot @BaochengSu

jan-kiszka:
First two steps done locally. And the third step now passes for node-red and node-red-gpio - without changing anything. Strange.

jan-kiszka:
Merged by now.

PR #456, opened 2023-06-19
Move to Debian 12

jan-kiszka:
To be discussed when to actually do that, but this shows that it is already possible without apparent problems. We probably had issues related to bookworm pre-release state in #453.

This will drop Google Coral support as discussed internally already.


jan-kiszka:
By now in the mode of cleaning up more details. All still look very good, and I would propose to move forward with this.

recipes-bsp/trusted-firmware-a/trusted-firmware-a-iot2050_2.9.bb
```diff
@@ -11,7 +11,7 @@
 require recipes-bsp/trusted-firmware-a/trusted-firmware-a-custom.inc
 
 SRC_URI += "https://git.trustedfirmware.org/TF-A/trusted-firmware-a.git/snapshot/trusted-firmware-a-${PV}.tar.gz"
-SRC_URI[sha256sum] = "53422dc649153838e03820330ba17cb10afe3e330ecde0db11e4d5f1361a33e6"
+SRC_URI[sha256sum] = "76a66a1de0c01aeb83dfc7b72b51173fe62c6e51d6fca17cc562393117bed08b"
```
BaochengSu, line 14: I wonder the detailed compiling errors? And seems the bookworm version is 2.8? https://packages.debian.org/bookworm/arm-trusted-firmware

jan-kiszka: https://git.trustedfirmware.org/TF-A/trusted-firmware-a.git/commit/?id=1f49db5f25cdd4e43825c9bcc0575070b80f628c
Maybe 2.8 is not bailing out anymore, only giving a warning (that no one saw). But we are self-building, so latest is better anyway.

BaochengSu: Sure, we already have the optee 3.21.0 > stock 3.19.0.

BaochengSu (approved):

PR #458, opened 2023-06-20
ci cancel in progress for same branch

BaochengSu:
This could save some resources, also makes the queue shorter,
environmentally friendly :)


jan-kiszka (approved):
:green_heart:

PR #459, opened 2023-06-25
linux-iot2050: Update to v5.10.184-cip36[-rt14]

jan-kiszka:
2 upstreamed patches less.

Tested as part of the demo preparation.


BaochengSu (approved):
LGTM, let's wait for the action before merging.

PR #467, opened 2023-07-13
Updates for our node-red nodes

jan-kiszka:
Draft as this still requires real device testing which is hard with the low bandwidth of my train.


jan-kiszka:
Tested now, seems to work fine.

BaochengSu (approved):

jan-kiszka:
Merge?

PR #455, opened 2023-06-18
Rework disk encryption

BaochengSu:
Use the latest upstreaming scheme, now all the changed bits are targeting isar.

~Also add the optee examples for iot2050, however, I am not sure whether or not to integrate it into the example image. Any comments?~ Dropped the optee-examples, to ease the integration.

PR #469, opened 2023-07-17
Switch npm from buildchroot to schroot

WiseLord:
This simplifies the code, allows to use cleaner prepare/build environment for npm packages and drops dependency on buildchroot that is going to be removed from isar soon.


kas/iot2050.yml
```diff
@@ -27,12 +27,9 @@ repos:
 
   isar:
     url: https://github.com/ilbers/isar
-    refspec: f95f69d3b83cc28976ef81eeefb5ec5dd17e544f
+    refspec: 9f6b5d92b974158d770a89459962764273995730
     layers:
       meta:
-    patches:
-      optee-os-tadevkit:
-        path: isar-patches/0001-optee-os-Add-package-optee-os-tadevkit.patch
```
jan-kiszka, line 35: I would prefer to get that done before this series, and that will not be that simple to move secure boot to the isar scheme.

recipes-devtools/sbuild-chroot/sbuild-chroot-target-npm.bb
```diff
@@ -0,0 +1,6 @@
+DESCRIPTION = "Isar sbuild/schroot filesystem for target (npm variant)"
+
+require recipes-devtools/sbuild-chroot/sbuild-chroot-target.bb
+
+SBUILD_FLAVOR = "npm"
+SBUILD_CHROOT_PREINSTALL_EXTRA ?= "npm dpkg-dev"
```
jan-kiszka, line 6: As discussed in #433, this will need more. I wonder why you didn't check that MR.

classes/npm.bbclass
```diff
@@ -211,7 +167,7 @@ python fetch_npm() {
         json_objs = {'dependencies': { npmpn: '' }}
         json.dump(json_objs, outfile, indent=2)
 
-    runcmd(d, "npm ci --global-style --ignore-scripts --verbose", "fetch-tmp")
+    runcmd(d, "npm ci --global-style --ignore-scripts --verbose")
```
jan-kiszka, line 170: Why dropping "fetch-tmp"?

recipes-devtools/sbuild-chroot/sbuild-chroot-target-npm.bb
WiseLord: I didn't see this pull-request. Will look later.

WiseLord: Looked a bit at #433. It seems to be related just to switching to derived chroot (with "npm" preintalled), while the patchset I propose is mostly about complete removing of buildchroot from npm.bbclass.

jan-kiszka: Sure, but that does not mean we should ignore what we discussed there.

kas/iot2050.yml
jan-kiszka: IOW, #455 needs to come first.

jan-kiszka:
Buildroot in isar upstream is gone - are you planning to refresh this, or should we take over, @WiseLord?

classes/npm.bbclass
jan-kiszka: I suppose that is because we are running this in a schroot, no longer in WORKDIR.

jan-kiszka:
OK, I'm taking over.

jan-kiszka:
Merged via #493

PR #472, opened 2023-08-01
Fix for RPMB key writing

BaochengSu:
Currently the RPMB key writing does not work:

1st issue was u-boot SPL complaining:

| alloc space exhausted
| Could not get FIT buffer of 1057180 bytes
|         check CONFIG_SYS_SPL_MALLOC_SIZE

This is fixed by expanding the heap size for malloc_simple.

Then comes to this issue:

| IOT2050> setenv -e -nv -bs -rt pairing 1
| I/TC: Reserved shared memory is enabled
| I/TC: Dynamic shared memory is enabled
| I/TC: Normal World virtualization support is disabled | I/TC: Asynchronous notifications are disabled
| E/LD:  init_elf:486 sys_open_ta_bin(ed32d533-99e6-4209-9cc0-2d72cdd998a7) | E/TC:? 0 ldelf_init_with_ldelf:131 ldelf failed with res: 0xffff0009 | Unable to open OP-TEE session (err=-5)
| mm_communicate failed!
| Error: Cannot initialize UEFI sub-system, r = 3

This is due to the StMM is not integrated for the rpmb-setup override.

### DIFF
--- a/recipes-bsp/optee-os/optee-os-iot2050_3.21.0.bb
+++ b/recipes-bsp/optee-os/optee-os-iot2050_3.21.0.bb
@@ -11,10 +11,10 @@
 require recipes-bsp/optee-os/optee-os-custom.inc
 require optee-os-iot2050_${PV}.inc
 
-# StMM integration
-DEPENDS:append:secureboot = " edk2-standalonemm-rpmb"
-DEBIAN_BUILD_DEPENDS:append:secureboot = ", edk2-standalonemm-rpmb"
-OPTEE_EXTRA_BUILDARGS:append:secureboot = " \
+# StMM integration, required by UEFI auth variable management
+DEPENDS:append:uefi-stmm = " edk2-standalonemm-rpmb"
+DEBIAN_BUILD_DEPENDS:append:uefi-stmm = ", edk2-standalonemm-rpmb"
+OPTEE_EXTRA_BUILDARGS:append:uefi-stmm = " \
     CFG_STMM_PATH=/usr/lib/edk2/BL32_AP_MM.fd \
     "
 
--- a/kas/opt/rpmb-setup.yml
+++ b/kas/opt/rpmb-setup.yml
@@ -13,4 +13,5 @@ header:
 
 local_conf_header:
   optee-os-iot2050_override: |
+    OVERRIDES .= ":uefi-stmm"
     OVERRIDES .= ":rpmb-setup"
--- a/kas/opt/secure-boot.yml
+++ b/kas/opt/secure-boot.yml
@@ -10,6 +10,7 @@ header:
 
 local_conf_header:
   secureboot_override: |
+    OVERRIDES .= ":uefi-stmm"
     OVERRIDES .= ":secureboot"
   secure-boot-signer: |
     IMAGER_BUILD_DEPS += "ebg-secure-boot-signer"
--- a/recipes-bsp/u-boot/files/0020-configs-Increase-malloc-size-after-relocation.patch
+++ b/recipes-bsp/u-boot/files/0020-configs-Increase-malloc-size-after-relocation.patch
@@ -1,0 +1,29 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Neha Malcom Francis <n-francis@ti.com>
+Date: Wed, 11 Jan 2023 18:11:23 +0530
+Subject: [PATCH] configs: Increase malloc size after relocation
+
+Current default size of 0x100000 is not capable of getting the FIT
+buffer during boot when transitioning to using binman generated boot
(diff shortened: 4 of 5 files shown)


recipes-bsp/u-boot/files/0020-iot2050-Increase-the-heap-size-for-malloc_simple.patch
```diff
@@ -0,0 +1,44 @@
...
+--- a/configs/iot2050_pg2_defconfig
++++ b/configs/iot2050_pg2_defconfig
+@@ -44,6 +44,7 @@ CONFIG_SPL_SYS_MALLOC_SIMPLE=y
+ CONFIG_SPL_STACK_R=y
+ CONFIG_SYS_SPL_MALLOC=y
+ CONFIG_SYS_SPL_MALLOC_SIZE=0x800000
++CONFIG_SPL_STACK_R_MALLOC_SIMPLE_LEN=0x200000
+ CONFIG_SYS_MMCSD_RAW_MODE_U_BOOT_USE_SECTOR=y
+ CONFIG_SYS_MMCSD_RAW_MODE_U_BOOT_SECTOR=0x1400
+ CONFIG_SPL_DM_MAILBOX=y
```
jan-kiszka, line 44: Please make sure to submit that upstream as well (will only affect a single config file now).

recipes-bsp/optee-os/optee-os-iot2050_3.21.0.bb
```diff
@@ -28,7 +28,11 @@ OPTEE_EXTRA_BUILDARGS:append:secureboot = " \
     "
 
 # RPMB key pairing
-OPTEE_EXTRA_BUILDARGS:append:rpmb-setup = " CFG_RPMB_WRITE_KEY=y"
+DEPENDS:append:rpmb-setup = " edk2-standalonemm-rpmb"
+DEBIAN_BUILD_DEPENDS:append:rpmb-setup = ", edk2-standalonemm-rpmb"
+OPTEE_EXTRA_BUILDARGS:append:rpmb-setup = " CFG_RPMB_WRITE_KEY=y \
+    CFG_STMM_PATH=/usr/lib/edk2/BL32_AP_MM.fd \
```
jan-kiszka, line 34: Extract the commong uefi config bits and inject them via a variable?

recipes-bsp/u-boot/files/0020-iot2050-Increase-the-heap-size-for-malloc_simple.patch
BaochengSu: Seems TI was bite by the same issue, and they submitted a patch to expand the size by *4.  Use their upstream version instead.

recipes-bsp/optee-os/optee-os-iot2050_3.21.0.bb
BaochengSu: Added a new "uefi" override, and put the StMM into this override. Both rpmb-setup.yml and secureboot.yaml declare this new override.

recipes-bsp/optee-os/optee-os-iot2050_3.21.0.bb
```diff
@@ -11,10 +11,10 @@
...
 require optee-os-iot2050_${PV}.inc
 
-# StMM integration
-DEPENDS:append:secureboot = " edk2-standalonemm-rpmb"
-DEBIAN_BUILD_DEPENDS:append:secureboot = ", edk2-standalonemm-rpmb"
-OPTEE_EXTRA_BUILDARGS:append:secureboot = " \
+# StMM integration, required by UEFI auth variable management
+DEPENDS:append:uefi = " edk2-standalonemm-rpmb"
+DEBIAN_BUILD_DEPENDS:append:uefi = ", edk2-standalonemm-rpmb"
+OPTEE_EXTRA_BUILDARGS:append:uefi = " \
```
jan-kiszka, line 17: `uefi` is a bit to generic for this. We may eventually switch to UEFI-only booting, but even then this override is only valid in secure UEFI mode. How about `uefi-stmm` or so?

BaochengSu: > but even then this override is only valid in secure UEFI mode

Couldn't we reuse this StMM for non-secure UEFI as well?

jan-kiszka: How? The RPMB would not be protected and on some boards not even available?

BaochengSu: I was thinking of letting only advanced variants use it, but that needs separate building for basic and advanced. So no, StMM is a no go for generic UEFI.

Let's go with `uefi-stmm`.

jan-kiszka (approved):

PR #474, opened 2023-08-10
iot2050:watchdog: Add support for WDIOF_CARDRESET

huaqianli:
This patch adds the WDIOF_CARDRESET support for the TI AM65x platform
watchdog, to know if the board reboot is due to a watchdog reset.

This is done via reserved memory(RAM), which indicates if specific
info saved in k3-rti-wdt, triggering the watchdog reset in last boot.

### DIFF
--- a/recipes-bsp/u-boot/files/0022-Add-a-reserved-memory-for-watchdog.patch
+++ b/recipes-bsp/u-boot/files/0022-Add-a-reserved-memory-for-watchdog.patch
@@ -1,0 +1,43 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Li Hua Qian <huaqian.li@siemens.com>
+Date: Wed, 7 Jun 2023 11:29:59 +0800
+Subject: [PATCH] Add a reserved memory for watchdog
+
+To have the WDIOF_CARDRESET support for the TI AM65x platform watchdog,
+this patch reserves the memory, which indicates if the current boot due
+to a watchdog reset.
+
+Signed-off-by: Li Hua Qian <huaqian.li@siemens.com>
+---
+ arch/arm/dts/k3-am65-iot2050-common.dtsi | 11 +++++++++++
+ 1 file changed, 11 insertions(+)
+
+diff --git a/arch/arm/dts/k3-am65-iot2050-common.dtsi b/arch/arm/dts/k3-am65-iot2050-common.dtsi
+index e60006be..acc41335 100644
+--- a/arch/arm/dts/k3-am65-iot2050-common.dtsi
++++ b/arch/arm/dts/k3-am65-iot2050-common.dtsi
+@@ -64,6 +64,12 @@
+ 			alignment = <0x1000>;
+ 			no-map;
+ 		};
++
++		/* To reserve the power-on(PON) reason for watchdog reset */
++		wdt_reset_memory_region: wdt-memory@a2200000 {
++			reg = <0x00 0xa2200000 0x00 0x00001000>;
++			no-map;
++		};
+ 	};
+ 
+ 	leds {
+@@ -731,6 +737,11 @@
+ 	mboxes = <&mailbox0_cluster1 &mbox_mcu_r5fss0_core1>;
+ };
+ 
++&mcu_rti1 {
++	memory-region = <&wdt_reset_memory_region>;
++
++};
++
+ &icssg0_mdio {
+ 	status = "disabled";
+ };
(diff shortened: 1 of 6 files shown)


jan-kiszka:
Just to note this clearly: This will also need the k3-wdt firmware patch - or an update of upstream once the needed changes are accepted there.

huaqianli:
> 
Yes, I didn't add it because I don't know which commit id will be used. Once the k3-wdt firmware is merged, I will update it.

recipes-bsp/k3-rti-wdt/k3-rti-wdt_1.1.bb
```diff
@@ -13,7 +13,7 @@ inherit dpkg
 SRC_URI = " \
     git://github.com/siemens/k3-rti-wdt.git;protocol=https;branch=master \
     file://rules"
-SRCREV = "33a6680184996074ca9731710161520edf334725"
+SRCREV = "806a7597d7853962d3eaac6340067829af0306c0"
```
jan-kiszka, line 16: Strictly spoken, we are no longer "1.1" with this bump.

Is everything working as it should? Then I could also tag 1.2 for k3-rti-wdt.

recipes-bsp/u-boot/files/0021-Add-a-reserved-memory-for-watchdog.patch
```diff
@@ -0,0 +1,30 @@
...
++++ b/arch/arm/dts/k3-am65-iot2050-common.dtsi
+@@ -64,6 +64,11 @@
+ 			alignment = <0x1000>;
+ 			no-map;
+ 		};
++
++		wdt_restet_memory_region: wdt-memeory@a2200000 {
++			reg = <0x00 0xa2200000 0x00 0x00001000>;
++			no-map;
++		};
```
jan-kiszka, line 27: This patch should keep `k3-am65-iot2050-common.dtsi` in sync between the kernel and U-Boot.

recipes-bsp/k3-rti-wdt/k3-rti-wdt_1.1.bb
huaqianli: Yes, I tested uboot and kernel boot watchdog, watchdog reset reason can get correctly.

recipes-bsp/u-boot/files/0021-Add-a-reserved-memory-for-watchdog.patch
huaqianli: Done.

recipes-bsp/k3-rti-wdt/k3-rti-wdt_1.1.bb
jan-kiszka: Ok, v1.2 tagged and pushed.

recipes-bsp/u-boot/files/0022-Add-a-reserved-memory-for-watchdog.patch
```diff
@@ -0,0 +1,31 @@
...
+@@ -64,6 +64,12 @@
+ 			alignment = <0x1000>;
+ 			no-map;
+ 		};
++
++		/* To reserve the power-on(PON) reason for watchdog reset */
++		wdt_reset_memory_region: wdt-memory@a2200000 {
++			reg = <0x00 0xa2200000 0x00 0x00001000>;
++			no-map;
++		};
```
jan-kiszka, line 28: Still missing the second hunk of the kernel patch.

huaqianli: Even if the second hunk is unnecessary,  do we have to add it in the uboot patch?

The second hunk is used by kernel driver to get the reserved address.

jan-kiszka: The U-Boot dts files are by now (in upstream) copies of the kernel files. So, yes, we want this, even if things won't be sync'ed anymore for old 2022.10, the patch may still apply to upcoming 2023.10.

jan-kiszka (approved):

jan-kiszka:
Quick question: How do I obtain the status from the kernel driver easiest?

huaqianli:
> Quick question: How do I obtain the status from the kernel driver easiest?

From my perspective, waiting for an email from community is the easiest way. You and @BaochengSu  have more experience on upstreaming, maybe you know better.

jan-kiszka:
Misunderstand: I meant how to read back the reset status on the device if this patch here is applied?

huaqianli:
> Misunderstanding: I meant how to read back the reset status on the device if this patch here is applied?

Here is an example:
```c
void foo(void)
{
    if ((fd = open("/dev/watchdog", O_RDWR)) < 0) {
        printf("Failed to open watchdog\n");
        return;
    }

    if (ioctl(fd, WDIOC_GETBOOTSTATUS, &status) < 0) {
        printf("Failed to get boot status!");
    }

    if (WDIOF_CARDRESET & status) {
        // WDIOF_CARDRESET is detected, do what you want to do
    }
}
````

jan-kiszka:
Ah, found it myself: `cat /sys/class/watchdog/watchdog0/bootstatus`

PR #476, opened 2023-08-17
Layer updates

jan-kiszka:
Brings new isar-cip with EFI Boot Guard update and signed SWUs as well as fresh Isar and kas.


jan-kiszka:
Hold on, there might be a problem with mounting efivarfs. Examining...

jan-kiszka:
> Hold on, there might be a problem with mounting efivarfs. Examining...

False alarm, that was caused by the RT kernel which disables EFI runtime services.

But I found and fixed some other issue. Now things should be ok.

BaochengSu (approved):

PR #488, opened 2023-09-27
u-boot-iot2050: Lift to 2023.10-rc4

jan-kiszka:
Most of our patches are merged now, just one DT change and a small fix for the signing script are remaining.

In this latest upstream version, we folded PG1 and PG2 U-Boot configs into a single one again, generating both artifacts in the same run. Consequently, we can simplify the build as there is now only one recipe as well.

Tests are pending.


jan-kiszka:
Now tested a bit better. Two more patches were needed, both sent upstream already.

jan-kiszka:
2023.10 has been released some days ago. Update to that final version will follow (one patch was merged in time).

Kconfig
```diff
@@ -128,7 +119,7 @@ config SECURE_BOOT
 
 config SECURE_BOOT
 	bool "Secure boot"
-	depends on IMAGE_BOOT_PG2
+	depends on IMAGE_BOOT
```
BaochengSu, line 122: Maybe we need to note somewhere that the security bits are not applicable to PG1 devices

recipes-bsp/u-boot/files/secure-boot.cfg
```diff
@@ -1,5 +1,5 @@
 ### Secure boot config
-# CONFIG_DISTRO_DEFAULTS is not set
+CONFIG_NETDEVICES=y
```
BaochengSu, line 2: Why this one is required?

Kconfig
jan-kiszka: Let me check if this fits into the Kconfig help.

recipes-bsp/u-boot/files/secure-boot.cfg
jan-kiszka: We need that config enabled so that the MAC addresses are forwarded to the DTB. But maybe it stays on, now that DISTRO_DEFAULTS are left in place. Will check again.

jan-kiszka: We get this without it:
```
  LD      u-boot
aarch64-linux-gnu-ld.bfd: board/siemens/iot2050/board.o: in function `set_board_info_env':
/data/u-boot/build-iot2050/../board/siemens/iot2050/board.c:213: undefined reference to `eth_env_set_enetaddr_by_index'
```
Reason remains the same: We use this to transfer the MAC address from the EEPROM to the DTB.

BaochengSu (approved):

PR #490, opened 2023-10-03
Drop obsolete bullseye-backports / fix CI / fix arm64 native build / fix Node-RED disabling

jan-kiszka:
We are on bookworm now.


jan-kiszka:
...and this also fixes a CI regression.

jan-kiszka:
...and now it also fixes a native build issue.

BaochengSu (approved):

PR #493, opened 2023-10-12
Rework npm to drop buildchroot dependency

jan-kiszka:
Based on patch by @WiseLord.


BaochengSu (approved):

jan-kiszka:
Looks like the isar or isar-cip-core update brought some regressions, swupdate image is broken :frowning_face:

PR #497, opened 2023-10-18
iot2050-firmware-update: Refactor and add map3 firmware update

huaqianli:
iot2050-firmware-update: Refactor and add map3 firmware update

### DIFF
--- a/recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.tmpl
+++ b/recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.tmpl
@@ -24,32 +24,28 @@ Example of update.conf.json:
 {
     "firmware": [
         {
-            "description": "[optional] bla bla bla",
-            "name": "pg1-basic-vx.y.z.bin",
-            "version": "[optional] whatever",
+            "description": "IOT2050 PG1 Bootloader Release V01.01.01",
+            "name": "iot2050-pg1-image-boot.bin",
+            "version": "V01.01.01",
+            "type": "uboot",
             "target_boards": [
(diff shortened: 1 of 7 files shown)


recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.tmpl
jan-kiszka, line 1: NACK, I'm using that all the time to update the freshly build binaries directly.

jan-kiszka:
FWIW: copy `flash-pgX.bin` / `iot2050-pgX-image-boot.bin` to target, run `iot2050-firmware-update -fn that.bin` on the target is a developer use case for the flashing tool that we should preserve. It already asks you often enough to confirm what you do.

jan-kiszka:
No commit message for first patch, and a major diff for that one - not really reviewable. What are the goals of this rewrite? Can those be achieved in smaller steps?

huaqianli:
> > No commit message for first patch, and a major diff for that one - not really reviewable. What are the goals of this rewrite? Can those be achieved in smaller steps?
> 
> At the beginning, we planed to add the map3 firmware into this tarball and expected to upgrade firmware as a whole, Which is why I removed the "--force" option.
> 
> When I added the map3 firmware, I found that we had to refactor the code for better maintainability and scalability. So I refactored it. There are no new features to be introduced, just refactoring.

Yesterday, we had some new findings, after a meeting we decided to upgrade map3 firmware with another way and keep the refactoring code without map3 part.

jan-kiszka:
Whenever you touch code, explain the reason. And the more you touch it, the longer the reasoning should be. "For better maintainability" is a bit too short here.

PR #504, opened 2023-11-06
linux-iot2050, mraa: Update kernel patches to latest upstream queue

jan-kiszka:
This primarily updates our own DT patches after their reviews and checks. That change also needs small adjustments of the MRAA path for debugfs-based pinmuxing due to node renamings.

While updating, also include more of the DT cleanups that are in latest upstream so that we are closer to the official DT.


jan-kiszka:
I was planning to include some more DT patches on top because I've noticed that upstream turns off more devices that we do so far. But that requires more effort, @diogo-ivo will look into that, and should not longer delay these changes here.

BaochengSu (approved):

PR #511, opened 2023-12-14
New applications added for IOT2050-SM

BaochengSu:
This is the last part of the IOT2050-SM series, targeting new added applications, including:
- EIO manager, which is a service set to manage the extended IO subsystems
- Conf webui, which is a web frontend for configuring the extended IO subsystem.
- Event Record, introduces some IOT2050 related events.


jan-kiszka:
Why all theses removal patches in this series? Why not simply starting small and only adding the incomplete things in a later series?

BaochengSu:
Refined the story line for a easier reviewing.

recipes-app/iot2050-eio-manager/iot2050-eio-manager_0.1.bb
```diff
@@ -0,0 +1,47 @@
...
+
+    install -v -m 755 ${WORKDIR}/iot2050_eio_global.py ${D}/usr/lib/iot2050/eio/
+    install -v -m 755 ${WORKDIR}/iot2050-eio-service.py ${D}/usr/lib/iot2050/eio/
+
+    install -v -d ${D}/lib/systemd/system/
+    install -v -m 644 ${WORKDIR}/iot2050-eiod.service ${D}/lib/systemd/system/
+
+    install -v -d ${D}/usr/bin/
+    ln -sf ../lib/iot2050/eio/iot2050-eio-service.py ${D}/usr/bin/iot2050-eio-service
+}
```
jan-kiszka, line 114: Wouldn't pip packaging be nice for this tool?

recipes-app/iot2050-eio-manager/iot2050-eio-manager_0.1.bb
```diff
@@ -40,8 +41,12 @@ do_install() {
     install -v -m 755 ${WORKDIR}/iot2050-eio-service.py ${D}/usr/lib/iot2050/eio/
 
     install -v -d ${D}/lib/systemd/system/
+    install -v -m 644 ${WORKDIR}/iot2050-eiofsd.service ${D}/lib/systemd/system/
     install -v -m 644 ${WORKDIR}/iot2050-eiod.service ${D}/lib/systemd/system/
 
     install -v -d ${D}/usr/bin/
+    install -v -m 755 ${WORKDIR}/bin/iot2050-eiofsd ${D}/usr/bin/
```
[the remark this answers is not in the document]
BaochengSu: The eiofsd is an internal Siemens project, and due to some reason, it will not be open sourced.

Indeed this will increase the repo size, just like we already have for SEBoot binaries... But I don't think it's easy to find some place eleswhere to hold them...

recipes-app/iot2050-eio-manager/iot2050-eio-manager_0.1.bb
BaochengSu: Any examples?

recipes-app/iot2050-eio-manager/iot2050-eio-manager_0.1.bb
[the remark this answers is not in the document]
jan-kiszka: We should clarify licenses - also for the SEBoot binaries as it becomes clear to me only now. Or do we have a statement for the latter already somewhere? Keep in mind that `COPYING.MIT`, the default license, obviously cannot apply.

recipes-app/iot2050-eio-manager/iot2050-eio-manager_0.1.bb
jan-kiszka: https://packaging.python.org/en/latest/tutorials/packaging-projects/?

jan-kiszka: Just an idea, any given that the debian package later on gains even a binary blob, it may not be the best fit.

recipes-app/iot2050-event-record/README.md
```diff
@@ -0,0 +1,75 @@
...
+    return response.event
+```
+
+And, please find the api definition in `gRPC/EventInterface/iot2050-event.proto`.
+
+## Regenerate the gRPC python modules if proto file changes:
+
+```sh
+python3 -m grpc_tools.protoc -I. --python_out=. --pyi_out=. --grpc_python_out=. gRPC/EventInterface/iot2050-event.proto
+```
```
jan-kiszka, line 75: How is this thing stressed in-tree? This is still unclear to me. It looks like it is only the service, not consumer at hand.

recipes-app/iot2050-event-record/files/iot2050-event-wdt.py
```diff
@@ -0,0 +1,94 @@
...
+    with grpc.insecure_channel(iot2050_event_api_server) as channel:
+        stub = EventRecordStub(channel)
+        response = stub.Read(ReadRequest(event_type=event_type))
+    return response.event
+
+def feeding_the_watchdog(fd):
+    while True:
+       ret = os.write(fd, b'watchdog')
+       print("Feeding the watchdog ...")
+       time.sleep(30)
```
jan-kiszka, line 74: Watchdog is driven by systemd, at least in the swupdate image. This looks fragile at best.

jan-kiszka: Even worse: Our watchdog requires feeding at specific times, not too early. I would refrain from doing this here. The whole example looks questionable TBH.

recipes-app/iot2050-eio-manager/iot2050-eio-manager_0.1.bb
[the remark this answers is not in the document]
BaochengSu: Any idea where to add license clarification of these binaries?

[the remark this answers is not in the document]
jan-kiszka: Next to the binary or in a top-level file (`COPYING.BINARY` or `COPYING.SIEMENS`?) that lists the paths along with the license text that applies to those files.

recipes-app/iot2050-event-record/files/iot2050-event-wdt.py
BaochengSu: Agree.

We should remove this watchdog from the event record service. Instead, this could be put into some wiki or readme.

@huaqianli, any comments?

recipes-app/iot2050-eio-manager/iot2050-eio-manager_0.1.bb
BaochengSu: I plan to make this EIO subsystem a docker image, could be in a separate layer, in the future, so let's keep it as it is now.

recipes-app/iot2050-conf-webui/README.md
```diff
@@ -0,0 +1,36 @@
+# IOT2050 Conf WEBUI
+
+This is the IOT2050 configuration webUI, it is used to setup the device and it's
+extension modules.
```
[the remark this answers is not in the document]
BaochengSu: Yep, will highlight it also here.

recipes-app/iot2050-event-record/files/iot2050-event-wdt.py
huaqianli: Agree. Due to the fact that the watchdog is non-stoppable, getting a watchdog event always requires opening the watchdog and feeding the watchdog. 

I removed this example and putted it into a `README(WATCHDOG.md)` and let the users to decide how to use it.

recipes-app/iot2050-event-record/README.md
```diff
@@ -0,0 +1,71 @@
...
+If enabling logging of sensor events is expected, please create a systemd
+drop-in for `iot2050-event-record.service`, as follows:
+
+```sh
+cp /usr/lib/iot2050/event/iot2050-event-record.conf /etc/systemd/system/iot2050-event-record.service.d/
+```
+
+# Watchdog events example
+
+If watchdog events recording is expected, please refer to [WATCHDOG.md](./WATCHDOG.md).
```
BaochengSu, line 24: It's not `example` anymore

BaochengSu: Do we have more than 1 event for watchdog?

recipes-app/iot2050-event-record/WATCHDOG.md
```diff
@@ -0,0 +1,127 @@
+Due to the fact that the watchdog is non-stoppable, getting a watchdog event always requires opening the watchdog and feeding the watchdog. So `event-record` doesn't contain the watchdog initially.
```
BaochengSu, line 1: Line too long

BaochengSu: , the watchdog event could not be included into the `iot2050-event-record` service.

recipes-app/iot2050-event-record/WATCHDOG.md
```diff
@@ -0,0 +1,127 @@
...
+
+This README file explains how to get the watchdog reset status and how to inject it into `event-record`.
```
BaochengSu, line 3: inject the event into the xxx service.

recipes-app/iot2050-event-record/WATCHDOG.md
```diff
@@ -0,0 +1,127 @@
...
+
+# How to get the watchdog reset status?
+
+Here is an example to get watchdog reset status, and print it. Let's name it `wdt_example.py`.
```
BaochengSu, line 7: The `wdt_example.py ` below shows how to get the watchdog reset status.

recipes-app/iot2050-event-record/README.md
huaqianli: ACK, and refined.

recipes-app/iot2050-event-record/README.md
huaqianli: ACK and done.

recipes-app/iot2050-event-record/WATCHDOG.md
huaqianli: ACK and done.

recipes-app/iot2050-event-record/WATCHDOG.md
huaqianli: ACK and done.

recipes-app/iot2050-event-record/WATCHDOG.md
huaqianli: ACK and done.

COPYING.SIEMENS
```diff
@@ -0,0 +1,28 @@
...
+
+3. No reverse engineering, decompilation, or disassembly is permitted.
+
+TO THE MAXIMUM EXTENT PERMITTED BY APPLICABLE LAW, THIS SOFTWARE IS PROVIDED
+"AS IS" WITHOUT WARRANTY OF ANY KIND, INCLUDING, WITHOUT LIMITATION, ANY EXPRESS
+OR IMPLIED WARRANTIES OF MERCHANTABILITY, ACCURACY, FITNESS OR SUFFICIENCY FOR A
+PARTICULAR PURPOSE, SATISFACTORY QUALITY, CORRESPONDENCE WITH DESCRIPTION, QUIET
+ENJOYMENT OR NON-INFRINGEMENT OF THIRD PARTY INTELLECTUAL PROPERTY RIGHTS.
+SIEMENS AG, ITS AFFILIATES AND THEIR SUPPLIERS DISCLAIM ANY WARRANTY THAT THE
+DELIVERABLES WILL OPERATE WITHOUT INTERRUPTION OR BE ERROR-FREE.
```
jan-kiszka, line 28: Just to be clear: This is not yet an official text!

recipes-core/images/iot2050-image-example.bb
```diff
@@ -41,6 +41,9 @@ IMAGE_INSTALL += " \
     libteec1 \
     optee-client-dev \
     tee-supplicant \
+    iot2050-eio-manager \
+    iot2050-conf-webui \
+    iot2050-event-record \
```
jan-kiszka, line 46: This should become a config option, maybe default-y but not unconditionally included - may too heavy.

BaochengSu: separated from the main config, and added Kconfig entry.

Kconfig
```diff
@@ -106,6 +106,19 @@ config KAS_INCLUDE_LXDE
 	default "kas/opt/lxde.yml"
 	depends on LXDE
 
+config EIO
+	bool "EIO subsystem support"
+	help
```
jan-kiszka, line 111: Not 'default y'? But we will enable it by default in the next example image release, I assume.

BaochengSu: Yes, we will enable it in our example image release building (internal CI), which will be uploaded into SIOS.

jan-kiszka: Then I would suggest to turn it on in the selection here by default so that users can easily reproduce what they also get prebuilt, just like for the Node-RED things.

BaochengSu: But that will leave the user a building error that the EIO binaries do not exist...

BaochengSu: Unless some dummy files are used for building, and print a warning to user that dummy files are used, and non interested users can just ignore that one.

jan-kiszka: Hmpf, just realized that as well. So this whole thing is also untested in CI? I suppose we can't easily fetch the missing files from their new homes, can we?

BaochengSu: > Hmpf, just realized that as well. So this whole thing is also untested in CI? I suppose we can't easily fetch the missing files from their new homes, can we?

We will copy the files into some folder at the prepare stage in CI.

No it's not possible to auto fetch the files, thanks to SIOS's policy that login before download.

recipes-app/iot2050-event-record/WATCHDOG.md
```diff
@@ -0,0 +1,81 @@
...
+    "wdt": "{} watchdog reset is detected",
+    "no-wdt": "{} watchdog reset isn't detected"
+}
+
+def feeding_the_watchdog(fd):
+    while True:
+       ret = os.write(fd, b'watchdog')
+       print("Feeding the watchdog ...")
+       # Let's say the watchdog timeout is more than 30 s
+       time.sleep(30)
```
jan-kiszka, line 57: What if systemd is feeding the watchdog? Do we still need this? Can we actually still use all this?

Kconfig
jan-kiszka: I'm not seeing any changes to this CI (downstream is irrelevant for QA).

recipes-app/iot2050-event-record/WATCHDOG.md
huaqianli: Rewrote the example to get boot status via `/sys/class/watchdog/watchdog0/bootstatus`.

Kconfig
fmoessbauer: Can't this be accessed via an API key? We could store that API key in a CI secret. It's ridiculous that the automation is blocked because of this non-sense.

BaochengSu: I've checked on the SIOS website, and no API key or token could be found... Maybe I am missing something...

BaochengSu: Change to stub the EIO binaries in the CI. And now CI will build EIO for all of the example image variants.

.github/workflows/main.yml
```diff
@@ -51,8 +61,13 @@ jobs:
...
       - name: Free Disk Space
         uses: ./.github/workflows/free-disk-space
+      - name: Mock EIO binary stubs
+        run: |
+          touch ./recipes-app/iot2050-eio-manager/files/bin/iot2050-eiofsd
+          touch ./recipes-app/iot2050-eio-manager/files/bin/map3-fw.bin
+          touch ./recipes-app/iot2050-eio-manager/files/bin/firmware-version
       - name: Build image
-        run: ./kas-container build kas-iot2050-swupdate.yml:kas/opt/secure-boot.yml
+        run: ./kas-container build kas-iot2050-swupdate.yml:kas/opt/secure-boot.yml:kas/opt/eio.yml
```
jan-kiszka, line 70: I think we don't need to enable EIO in all three cases because it is not directly affected by the variants. Maybe just append it to the secure boot variant?

Actually, but that is an optimization on top, I could imagine folding the RT option into the secure boot target as well to save resources.

recipes-app/iot2050-conf-webui/README.md
```diff
@@ -0,0 +1,47 @@
...
+Supported browser: Firefox, Chrome, Edge, etc. IE is not supported.
+
+## Development 
+
+This application is using Next.JS + Material UI.
+
+There are two `npm-shrinkwrap.json` files:
+
+- `npm-shrinkwrap.json`
+- `npm-shrinkwrap.json.nodev`
```
jan-kiszka, line 24: "nodev" - or rather "nodep"?

jan-kiszka: Ah, dev dependencies, now I'm getting the naming.

jan-kiszka (approved):
I'm fine, remaining comment is not a blocker.

.github/workflows/main.yml
BaochengSu: Agree and changed.

PR #521, opened 2024-01-22
linux: pci: Fix PCI probing

diogo-ivo:
The Keystone PCI controller had a race condition hidden that made its
probing fail. Add a patch provided by TI (currently mainlined) that
fixes this race condition and returns PCI back to a funcional state.
Tested with a Hailo-8 NPU card.

While at it add a second patch suggested by TI that fixes "a 45 second
delay when no Endpoint Device is connected to the PCIe connector".

### DIFF
--- a/recipes-kernel/linux/files/patches-6.1/0095-PCI-keystone-Fix-race-condition-when-initializing-PH.patch
+++ b/recipes-kernel/linux/files/patches-6.1/0095-PCI-keystone-Fix-race-condition-when-initializing-PH.patch
@@ -1,0 +1,54 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Siddharth Vadapalli <s-vadapalli@ti.com>
+Date: Wed, 27 Sep 2023 09:48:45 +0530
+Subject: [PATCH] PCI: keystone: Fix race condition when initializing PHYs
+
+The PCI driver invokes the PHY APIs using the ks_pcie_enable_phy()
+function. The PHY in this case is the Serdes. It is possible that the
+PCI instance is configured for two lane operation across two different
+Serdes instances, using one lane of each Serdes.
+
+In such a configuration, if the reference clock for one Serdes is
+provided by the other Serdes, it results in a race condition. After the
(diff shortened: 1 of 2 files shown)


jan-kiszka:
And do they fix the issues we've seen? This remains unclear to me when just reading this PR.

If so, it would we good to also request them to go into stable.

diogo-ivo:
Yes, this fixes the probing and now the Hailo device is seen when running lspci

diogo-ivo:
Regarding pushing them to stable I'll send an e-mail tomorrow

recipes-kernel/linux/files/patches-6.1/0086-PCI-keystone-Fix-race-condition-when-initializing-PH.patch
```diff
@@ -0,0 +1,61 @@
+From 36836820f571872d6856c17bdc1d0e4f8418915d Mon Sep 17 00:00:00 2001
```
BaochengSu, line 1: The patch file should be generated from:

```shell
git format-patch --abbrev=12 --no-numbered --zero-commit --no-signature ...
```

see: https://github.com/siemens/meta-iot2050/blob/master/CONTRIBUTING.md#patch-files-generation

Meanwhile I will test the SSD on M.2 variant.

diogo-ivo: Ok, I'll prepare the patches like that. Please let me know the results of your testing when you can so I can reply to TI

BaochengSu: Now with the patch my PM971 256G SSD is working.

BaochengSu (changes requested):
Please rebase the patches - change the number, sorry that #520 merged firstly 😉

BaochengSu:
> Please rebase the patches - change the number, sorry that #520 merged firstly 😉

Never mind, I plan to merge the #511 first, then I will do the rebase for this PR.

BaochengSu (approved):

PR #530, opened 2024-03-06
Isar and isar-cip-core updates

jan-kiszka:
Probably post-release material, but downstream users are already in need for this.

@gylstorffq


gylstorffq:
Thanks i tested it downstream and looks good to me.

BaochengSu (approved):

PR #531, opened 2024-03-07
refactor: use change-dir to split rootfs parts

asrinuva:
This patch refactors the WKS script w.r.t how the rootfs is distributed across multiple partitions. Previously we used rootfs-dir with a full rootfs paths plus the sub directory. However, this is not ideal, as the relation to the rootfs folder is lost and also the new rootfs dir is technically not a rootfs. While this aparently did not create any issues in meta-iot2050, we still believe it is semantically incorrect.

This patch switches that over to the WIC option --change-directory that is provided exactly for this use-case and only deploys a subdirectory without loosing the rootfs relation with the permissions pseudo dir.

### DIFF
--- a/wic/iot2050-swu-secure.wks.in
+++ b/wic/iot2050-swu-secure.wks.in
@@ -14,7 +14,7 @@ part --source rawcopy --sourceparams "file=${IMAGE_FULLNAME}.verity" --align 102
 part --source empty --align 1024 --fixed-size 4G --uuid "${ABROOTFS_PART_UUID_B}"
 
 # persistent partitions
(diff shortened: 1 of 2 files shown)


asrinuva:
@jan-kiszka Please have a look.

jan-kiszka (approved):
Valid cleanup. Not critical for the release but also not risky.

asrinuva:
Can you please merge this changes if no objection.

jan-kiszka:
I explained before that there is an upcoming release and that this is a cleanup, not a fix. I suspect it will only be merged afterwards.

asrinuva:
> I explained before that there is an upcoming release and that this is a cleanup, not a fix. I suspect it will only be merged afterwards.

Okay, thanks.

PR #535, opened 2024-03-28
ti-pruss-firmware: Set DEBIAN_COMPAT to overcome do_dpkg_build fail

ananthakr23:
With latest ISAR in our layers we observed do_dpkg_build was giving following errors:
  - objcopy: Unable to recognise the format of the input file `debian/ti-pruss-firmware/usr/lib/debug/.dwz/aarch64-linux-gnu/ti-pruss-firmware.debug'
  - strip: Unable to recognise the format of the input file `debian/ti-pruss-firmware/lib/firmware/ti-pruss/am65x-sr2-rtu0-prueth-fw.elf' 
 This occurs with the change in default DEBIAN_COMPAT version. Build passes if we use DEBIAN_COMPAT = "10" instead of default debhelper-compat = 13


jan-kiszka:
Just wait for #530 (or carry that locally for now).

Issue #29, opened 2020-07-06
GPIO accessible via gpio group


simonabler:
can you put the arduino interface DI, DO, user leds , ..  in a gpio group. 

i managed to put the led via the udev rules into the gpio group but failed with the DOs of the arduino shield.

cat >/etc/udev/rules.d/98-comLed.rules  << EOF1
SUBSYSTEM=="gpio*", PROGRAM="/bin/sh -c 'chown -R root:gpio /sys/class/leds && chmod -R 770 /sys/class/leds && chown -R root:gpio /sys/devices/platform/leds/leds && chmod -R 770 /sys/devices/platform/leds/leds'"
EOF1 

can you set this per default?

jan-kiszka:
What is the requirement behind this request? Is it about access control management?

The example image is a reference for creating own customization on top. Adding own users, managing their permissions would be one of the possible customizations. If a pattern is generic enough, we are happy to consider including it also in the example image. The best approach to propose that is by submitting a pull request with the desired changes. If that is too tricky, we need at least some more details on the desired use case.

simonabler:
if you using the example image. you are forced to use the root user for node-red.
otherwise you cannot use the gpios.

So in my view it is generic to put the leds, gpios and user buttons into a special "gpio" group.

Therefore someone can put own customization on top, adding new user, and give them easily access to the hardware via the gpio group.

jan-kiszka:
Ok, so the actual requirement is "run Node Red as non-root user". Permitting non-root access to GPIOs is a building block for that, but it is not reasonable to only do that without a use case, thus also a test case.

jan-kiszka:
I just learned that there are limits so far in de-privileging anything that needs mraa to access IOs - and that included node-red. The reason is the way the pin mux controller is accessed from mraa: as directly mapped MMIO via /dev/mem, and that is naturally a highly privileged access. Currently, we are missing an alternative. And nothing would be won by simply adding /dev/mem to such a GPIO group as well. Any user in that group would still have effective root rights.

simonabler:
ok, i understand. That makes sense.

Thank you for your researching.

jan-kiszka:
We can reopen this topic when the dependency on /dev/mem is resolved. But we have no good idea for that yet, though it's clearly desirable.

mwesenjak:
[...]

PR #541, opened 2024-04-18
Fix MRAA non-root access

benbrenson:
This MR fixes some issues related the MRAA API access, as non-root user.
The following things have been changed/fixed:
1. Add gpio chardev support for MRAA (switch from sysfs to chardev)
2. Add udev rules that adjust permissions and assign required groups to all interfaces
3. The PWM pins require a more special handling and need to be pre-allocated during boot. This avoids
    timing problems when exporting those pins the first time via MRAA (see issue #29). 
4. Create an example (non-root) user that has been assigned to all required groups.
5. Fix MRAA pinmux handling for pin 20

### DIFF
--- a/kas-iot2050-example.yml
+++ b/kas-iot2050-example.yml
@@ -22,3 +22,12 @@ local_conf_header:
     USERS += "root"
     USER_root[password] ??= "root"
     USER_root[flags] ??= "clear-text-password force-passwd-change"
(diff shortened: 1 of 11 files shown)


jan-kiszka (approved):
Looks good, worth a try - hope it still cleanly rebases over master.

Issue #440, opened 2023-05-09
Reboot problem


rFond:
Hi,

When IOT works for a very long period (60 day) and I launch a reboot with cmd **sudo reboot**. 

IOT can't reboot, with software, I force to do an hard reset electrical.
`=======================
Board: IOT2050-ADVANCED
Serial: xx
MLFB: 6ES7647-0BA00-1YA2
UUID: 8069DFE5E3D14C05B9B8D6C4CAC1CEF0
A5E: A5E452229880AB08
MAC[0]: xx
MAC[1]: xx
SKU: SE
Loading PK... ok
PK count: 00
PK version: 00
SV: 00-00
Security ID 0x5125e53f-0x842e215d
Security policy: soft
Loading image atf...
Loading image tee...
Loading image spl...
Loading image k3-am65-iot2050-spl.dtb...
NOTICE:  BL31: v2.5(release):
NOTICE:  BL31: Built : 14:41:24, May 17 2021
I/TC:
I/TC: OP-TEE version: 3.12.0 (gcc version 10.2.1 20210110 (Debian 10.2.1-6)) #1 Wed Jan 20 17:48:48 UTC 2021 aarch64
I/TC: Primary CPU initializing
I/TC: Primary CPU switching to normal world boot

U-Boot SPL 2021.04-V01.02.01.03-0-g7e29ca7 (Nov 04 2021 - 07:46:39 +0000)
SYSFW ABI: 3.1 (firmware rev 0x0015 '21.5.0--v2021.05 (Terrific Llam')
Trying to boot from SPI
QSPI: QSPI is still busy after poll for 10000 times.
SF: Calibration failed (read)
cadence_spi spi@47040000: Cannot set speed (err=-5)
SPI probe failed.
SPL: failed to boot from all boot devices
### ERROR ### Please RESET the board ###
`

Thanks for your help.

rFond:
After a cold reset (with button ou electrical), error no gone again.


After.
======================
Board: IOT2050-ADVANCED
Serial: NNEM7528
MLFB: 6ES7647-0BA00-1YA2
UUID: 8069DFE5E3D14C05B9B8D6C4CAC1CEF0
A5E: A5E452229880AB08
MAC[0]: xx
MAC[1]: xx
SKU: SE
Loading PK... ok
PK count: 00
PK version: 00
SV: 00-00
Security ID 0x5125e53f-0x842e215d
Security policy: soft
Loading image atf...
Loading image tee...
[...]

PR #570, opened 2024-12-16
[v2]u-boot: Fix the SPI and eMMC hang issue

BaochengSu:
Another fix for the infamous #440 , should be the right one.

Refer to 1 and 2, the root cause is the wrong data type of the timer value.

Thanks ronald.wahl@legrand.com for the patches.

1. https://lists.denx.de/pipermail/u-boot/2024-December/574817.html
2. https://lists.denx.de/pipermail/u-boot/2024-December/574818.html

v1 -> v2:

The SD patch is removed, we doubt the SD card is another different issue and should be treated differently.

v1:
The SD patch is kept, however most probably the SD card issue is a different one.


recipes-bsp/u-boot/u-boot-iot2050_2023.10.bb
```diff
@@ -23,9 +23,10 @@ SRC_URI += " \
     file://0009-dts-iot2050-Sync-kernel-dts-to-u-boot.patch \
     file://0010-dts-iot2050-Support-new-IOT2050-SM-variant.patch \
     file://0011-arm-dts-iot2050-Disable-lock-step-mode-for-all-iot20.patch \
-    file://0012-qspi-iot2050-Add-temporary-workaround-for-the-QSPI-i.patch \
-    file://0013-driver-iot2050-Add-a-temporary-workaround-for-the-eMMC.patch \
-    file://0014-driver-iot2050-Add-a-temporary-workaround-for-the-SD.patch \
+    file://0012-spi-cadence-quadspi-Fix-error-message-on-stuck-busy-.patch \
+    file://0013-spi-cadence-quadspi-fix-potential-malfunction-after-.patch \
+    file://0014-mmc-Fix-potential-timer-value-truncation.patch \
+    file://0015-driver-iot2050-Add-a-temporary-workaround-for-the-SD.patch \
```
jan-kiszka, line 29: Do we still need patch 15, given patch 14?

BaochengSu: Actually during my speeded-up reproducing (by using a `u16 start`), I have not observed the SD card issue, this echoes the code exactly, no wrong sized timer variables are used in the SD driver part.

So I think the SD issue is a different one, so I keep the patch 15.

BaochengSu: Wait, I actually tested the SD on an advanced device, IIRC, the SD card issue is reported on a basic device, let me try it on the basic device then let's see.

BaochengSu: Just tested on both PG1 & 2 basic devices, SD card booting does not have such issue pattern. So better to keep the patch 15.

jan-kiszka: - the mmc fails when also SPI failed
- patch 15 also deals with timeouts
- the patched code also used an "int" for holding the timeout
- patch 14 changes some uint to and ulong in the same code file to address an overflow issue

Isn't all that screaming for a closer look? I suspect `sd_send_op_cond` is just missing the same fix as `__mmc_switch` and `mmc_send_op_cond` receive via patch 14. Please do so, upstream first.

BaochengSu: The` int timeout` in `sd_send_op_cond` does not bind to any `get_timer`.

jan-kiszka: This is correct, but it does not change the key point: Either, patch 15 is no obsolete because the root cause was found and fixed, or we should look for any remaining timer overflows that caused that issues. It is more than likely that the same root cause applies to "our" mmc issue as well.

BaochengSu: That's also what I thought, however, after a deep dive into the SD card calling stack (not only, I've checked almost every get_timer calling in the u-boot -- indeed we are in lucky, there are few u32 cases, and most of them are command which does not matter in our issue...), I found there is no other place to overflow the timer, besides the reported place, so this SD card issue is either a false alarm or a different issue, in this case, better to keep the patch 15, since that patch won't cause anything harmful even if it does not fix the issue.

jan-kiszka: Tune get_time to return an overflowing value until the mmc issue reoccurs and debug the root cause / validate that it is fixed by patch 14. Just keeping the hack of patch 15 is not really the way forward anymore, I would say.

BaochengSu: Test with 

```diff
diff --git a/arch/arm/cpu/armv8/generic_timer.c b/arch/arm/cpu/armv8/generic_timer.c
index 8f83372cbca4..cef3b03e7335 100644
--- a/arch/arm/cpu/armv8/generic_timer.c
+++ b/arch/arm/cpu/armv8/generic_timer.c
@@ -85,7 +85,10 @@ unsigned long notrace timer_read_counter(void)
 	isb();
 	asm volatile("mrs %0, cntpct_el0" : "=r" (cntpct));
 
-	return cntpct;
+	/* Add 0xFFFFFFFF * get_tbclk()
+	 * [get_tbclk() == 200_000_000]
+	 */
+	return cntpct + (unsigned long)0xBEBC1FFF4143E00;
 }
 #endif
```

Time is indeed u32 overflowed:

```
IOT2050> gettime
Timer val: 52216177527 (0xC_2853_A377)
Seconds : 52216177
Remainder : 527
```

With this change:

- without the fixes, the SPI issue is reproduced, after patching the SPI, the eMMC issue is reproduced.
- with both fixes, no issue found for SD card booting on both BASIC & ADVANCED PG2:

```
U-Boot 2023.10-00015-g77b0be9f7ee2-dirty (Dec 23 2024 - 09:36:43 +0800)

Model: SIMATIC IOT2050 Basic PG2
DRAM:  1 GiB
Core:  175 devices, 32 uclasses, devicetree: separate
WDT:   Not starting watchdog@40610000
MMC:   mmc@4fa0000: 0
Loading Environment from SPIFlash... SF: Detected w25q128 with page size 256 Bytes, erase size 64 KiB, total 16 MiB
OK
In:    serial@2810000
Out:   serial@2810000
Err:   serial@2810000
Net:   No ethernet found.
Hit SPACE to stop autoboot in 2 seconds...
switch to partitions #0, OK
mmc0 is current device
Scanning mmc 0:1...
Found U-Boot script /boot/boot.scr
532 bytes read in 30 ms (16.6 KiB/s)
## Executing script at 80000000
Loading /usr/lib/linux-image-6.1.102-cip26/ti/k3-am6528-iot2050-basic-pg2.dtb...
57889 bytes read in 41 ms (1.3 MiB/s)
Loading /boot/vmlinux-6.1.102-cip26...
19728896 bytes read in 819 ms (23 MiB/s)
No EFI system partition
No EFI system partition
Failed to persist EFI variables
Booting /\boot\vmlinux-6.1.102-cip26
EFI stub: Booting Linux Kernel...
EFI stub: Using DTB from configuration table
EFI stub: Exiting boot services...
[    0.000000] Booting Linux on physical CPU 0x0000000000 [0x410fd034]
...
```

- With only the SPI fix, the SD card can still boot without issue -- the eMMC bug does not impact the common path.

So still sounds to me either a false positive or a different issue.

jan-kiszka: Just to make sure I got that correctly: The enforced 32-bit overrun of timer_read_counter helps to trigger the eMMC bug immediately, it's just that patch 14 does not help with it. But then you now have the means to debug the eMMC issue, right?

BaochengSu: >  The enforced 32-bit overrun of timer_read_counter helps to trigger the eMMC bug immediately

Yes

> it's just that patch 14 does not help with it

No, patch 14 works perfectly for eMMC issue, otherwise I would remove it from the patch list.

The point now is all about patch 15 or the "SD card issue".

BaochengSu: The two patches mentioned in my reply are 13 & 14.

jan-kiszka: Ok, I was confusing the eMMC issue with the SD card one. The latter was never observed so far if I read 088f2a9adec5f0da68c20846e81cf26fd4059949 correctly. In that light, patch 15 looks even more dubious to me.

BaochengSu: > Ok, I was confusing the eMMC issue with the SD card one. The latter was never observed so far if I read [088f2a9](https://github.com/siemens/meta-iot2050/commit/088f2a9adec5f0da68c20846e81cf26fd4059949) correctly. In that light, patch 15 looks even more dubious to me.

Just have had a talk with Hua Qian, the mentioned commit message is kind of out-dated: there was one observation of SD card not booting with the SPI flash workaround in our testing.

However, now I doubt on the helpfulness of patch 15 as well, I think that observation should be another issue, maybe some incidental SD card non-recognition. If so such kind of issue should be easily solved by a watchdog reboot if watchdog_timeout_ms is configured as non-zero.

BaochengSu: Patch 15 is removed.

jan-kiszka (approved):

PR #576, opened 2025-01-17
u-boot-iot2050: Match kernel dts for bootefi issue

huaqianli:
The bootefi command leads to the kernel FDT being overwritten by the U-Boot FDT. Although the UKI's FDT ultimately takes precedence, this intermediate overwrite can cause the device to fail to apply the overlay.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,7 +9,7 @@ concurrency:
 jobs:
   debian-example-image:
     name: Debian example image
-    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


huaqianli:
> Be reminded that this will remain a temporary patch due to https://source.denx.de/u-boot/u-boot/-/commit/53413181655d22416784c94a0a3b0efe2c421e04.

Ok, will  refine it when that window opens.

BaochengSu (approved):

PR #581, opened 2025-02-06
linux-iot2050: Add the missing kbuildtarget

BaochengSu:
The kbuildtarget is not set in the iot2050-debian.conf, this is causing both the kernels (cip & upstream) are be parsed and generated in the build process, which causes the parsing stage is very slow when connecting to github is not good.

### DIFF
--- a/conf/distro/iot2050-debian.conf
+++ b/conf/distro/iot2050-debian.conf
@@ -26,6 +26,8 @@ PREFERRED_VERSION_linux-iot2050-rt-kbuildtarget ?= "${PREFERRED_VERSION_linux-io
 
 KERNEL_NAME ?= "cip"
 
+BBMASK:append = " linux-iot2050_.+-upstream\.bb"
+
 PREFERRED_VERSION_customizations ?= "0.1-iot2050-debian"
 PREFERRED_VERSION_customizations-debug ?= "0.1-iot2050-debian"
 
--- a/conf/distro/iot2050-debian.conf
+++ b/conf/distro/iot2050-debian.conf
@@ -19,6 +19,11 @@ HOSTNAME ??= "iot2050-debian"
 PREFERRED_VERSION_linux-iot2050 ?= "6.1.%"
 PREFERRED_VERSION_linux-iot2050-rt ?= "6.1.%"
 
+PREFERRED_VERSION_linux-iot2050-native ?= "${PREFERRED_VERSION_linux-iot2050}"
+PREFERRED_VERSION_linux-iot2050-kbuildtarget ?= "${PREFERRED_VERSION_linux-iot2050}"
+PREFERRED_VERSION_linux-iot2050-rt-native ?= "${PREFERRED_VERSION_linux-iot2050-rt}"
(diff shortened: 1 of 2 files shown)


huaqianli (approved):

conf/distro/iot2050-debian.conf
```diff
@@ -18,6 +18,8 @@ HOSTNAME ??= "iot2050-debian"
 
 PREFERRED_VERSION_linux-iot2050 ?= "6.1.%"
 PREFERRED_VERSION_linux-iot2050-rt ?= "6.1.%"
+PREFERRED_VERSION_linux-iot2050-kbuildtarget ?= "6.1.%"
+PREFERRED_VERSION_linux-iot2050-rt-kbuildtarget ?= "6.1.%"
```
jan-kiszka, line 22: To keep things a bit more maintainable, I would suggest this pattern:
https://gitlab.com/cip-project/cip-core/isar-cip-core/-/blob/master/conf/distro/cip-core-common.inc

BaochengSu: Acked and revised. BTW, sorry for the confusion, the kbuildtarget missing only confuse isar while choosing the correct kbuildtarget version, the upstream recipe is parsed regardless. This is because the "AUTOREV" in the upstream recipe. I append a new commit to address it, by masking the upstream recipe.

conf/distro/iot2050-debian.conf
```diff
@@ -17,10 +17,14 @@ DISTRO_NAME = "IOT2050 Debian System"
 HOSTNAME ??= "iot2050-debian"
 
 PREFERRED_VERSION_linux-iot2050 ?= "6.1.%"
+PREFERRED_VERSION_linux-iot2050-kbuildtarget ?= "${PREFERRED_VERSION_linux-iot2050}"
 PREFERRED_VERSION_linux-iot2050-rt ?= "6.1.%"
+PREFERRED_VERSION_linux-iot2050-rt-kbuildtarget ?= "${PREFERRED_VERSION_linux-iot2050-rt}"
```
jan-kiszka, line 22: Doesn't the missing PREFERRED_VERSION...-native bit you?

BaochengSu: Should we append the -native to the linux-iot2050? How it looks like when it bites us? At least the recipe parsing has no problem and kernel/rt-kernel is building

jan-kiszka: You will see multiple provider warnings when bitbake identifies them during parsing.

BaochengSu: Didn't see any warnings... I guess due to isar version? Let me try

jan-kiszka: -kbuildtarget and -native were introduced by the same isar changes. It takes two or more providers for those targets to make bitbake complain. I cannot explain, though, why 6.x-upstream should trigger the warning for kbuildtarget but not for native.

BTW, iot2050-linux-rt should only have one provider at this stage.

So much for the theory...

PR #582, opened 2025-02-07
Userland PM16D17

BaochengSu:
Upstream is hard for this chip driver, we decide to use userland driver instead.


recipes-app/iot2050-event-record/files/iot2050-event-record.py
```diff
@@ -15,6 +15,7 @@
 from datetime import datetime
 from systemd import journal
 from enum import Enum
+from pystemd.dbuslib import DBus, DbusMessage
```
huaqianli, line 18: Make the `event-record` an identical user story?

huaqianli: There is a leakage due to the lack of kernel ABI or userspace API compatibility, which seems to be an unmanageable limitation.

Otherwise, looks good to me.

huaqianli: Regarding our upcoming release, we may need to hold off on merging this PR to avoid involving additional third-party libraries for OSS clearance.

jan-kiszka:
Quite a lot of changes... Given all this effort, it might have been worth to make at least the complications with Everlight transparent to upstream and ask for further guidance.

BaochengSu:
> Quite a lot of changes... Given all this effort, it might have been worth to make at least the complications with Everlight transparent to upstream and ask for further guidance.

Sure, I will append our new findings to the kernel mailing list.

huaqianli (approved):

PR #584, opened 2025-02-24
fix(firmware-update): explicitly set DPKG_ARCH to "any"

chombourger:
Isar commit 0816ae6e97d15712f6b8eb873311464263f5df59 changed the default architecture for dpkg-raw from "any" to "all". Having a task that assumes the generated package file ends with "_arm64.deb" (not great btw), set DPKG_ARCH to "any".


chombourger:
@BaochengSu kindly review

BaochengSu (approved):

PR #585, opened 2025-02-25
[RFC]: patch-u-boot-env: optimize patch-u-boot-env service execution

rakeshk7097:
Update patch-u-boot-env.service to set the U-Boot environment (e.g., watchdog_timeout_ms=60000) earlier and more reliably during boot.With the special case when watchdog_timeout_ms value set as '0' patch-u-boot-env.service ExecStart is not executing the 'fw_setenv watchdog_timeout_ms 60000' in this case patch-u-boot-env.service is getting failed in setting the watchdog_timeout_ms as 60000. Running the service after 'local-fs.target' and 'systemd-udev-settle.service' is letting the block device intialize ready to run the fw_setenv command.

### DIFF
--- a/recipes-bsp/patch-u-boot-env/files/patch-u-boot-env.service
+++ b/recipes-bsp/patch-u-boot-env/files/patch-u-boot-env.service
@@ -8,12 +8,14 @@
 
 [Unit]
 Description=update u-boot environment for swupdate
+After=multi-user.target
 Conflicts=shutdown.target
 Before=swupdate.service shutdown.target
+DefaultDependencies=no
 
 [Service]
 Type=oneshot
-ExecStart=sh -c '/usr/share/u-boot-env/patch-u-boot-env.sh watchdog_timeout_ms /usr/share/u-boot-env/patch-u-boot-env.config; fw_setenv watchdog_timeout_ms 60000'
+ExecStart=/usr/share/u-boot-env/patch-u-boot-env.sh watchdog_timeout_ms /usr/share/u-boot-env/patch-u-boot-env.config
 ExecStartPost=-/bin/systemctl disable patch-u-boot-env.service
 
 [Install]
--- a/recipes-bsp/patch-u-boot-env/files/patch-u-boot-env.sh
+++ b/recipes-bsp/patch-u-boot-env/files/patch-u-boot-env.sh
@@ -7,9 +7,14 @@
 #
 # SPDX-License-Identifier: MIT
 
-if [ -z "$(fw_printenv -n "$1")" ]; then
-    if ! fw_setenv -s "$2"; then
-        echo "$0: could not patch u-boot firmware"
+if ! fw_setenv -s "$2"; then
+        echo "$0: Failed to apply config file $2"
         exit 1
-    fi
+fi
+
+watchdog_timer="60000"
+current_value=$(fw_printenv -n "$1" 2>/dev/null || echo "")
+if [ "$current_value" != "$watchdog_timer" ]; then
+        echo "$0: Failed to set $1 to $expected_value (got $current_value)"
+        exit 1
 fi


jan-kiszka:
We are hunting some sporadic setup issue around this as well, and your patch looks promising. However, we didn't get any error messages in case fw_setenv failed to set the watchdog. Did you see any errors? Or was the watchdog_timeout_ms just not updated, silently?

colpane:
This is the same (sporadic) issue I have reported to both of you. We are still trying to get what fw_setenv returns at least (in our case it is silent). Maybe Rakesh did observe something.

rakeshk7097:
Yes @jan-kiszka , its sporadic, but we also didn't observer any error while running the service, sometimes we are seeing that service got started and finished also but this `fw_setenv watchdog_timeout_ms 60000` in service file seems executed silently.
I tried with putting as sleep of 3 sec after ExecStart as ExecStartPost and this is also working fine, there I suspected fw_setenv might needs device node to populate and ready which `systemd-udev-settle.service` fulfill. 

Do we need to set the timer watchdog_timeout_ms value hardcoded as 60sec even if its 0 or will it be better if we can make it as build time configurable for user to set its value as per their need. 

I will push the updated patch for making the watchdog timer as configurable.

jan-kiszka:
> I tried with putting as sleep of 3 sec after ExecStart as ExecStartPost and this is also working fine

Did you actually mean "Post", or rather "Pre"? That would make even less sense?

We are still searching for the real dependency. Already compared the ordering of services in failed case with a working on? -> `journalctl`

BTW, were you able to reproduce the issue also with this layer alone, or was there always another layer (with potentially extra services) involved?

rakeshk7097:
HI @jan-kiszka , thanks for your input,

Tested multiple times with meta-iot2050 image alone, and i can observed that watchdog_timeout_ms is not getting updated every time, 3 out of 5 its getting updated and rest its not able to set the timer value, so seems its quite sporadic issue.  

Seems like  in every boot the hardware initialization, or system load, causing the device to be ready in some boots but not others.
fw_setenv might exit with 0 (success) even if it fails to write to the environment (e.g., due to device not ready or permissions). Systemd doesn’t see this as a failure unless the command explicitly returns a non-zero exit code.

If the device is ready, the write succeeds; if not, it fails silently, but the service still reports success.

Do we really need to hard code this watchdog timer as 60 secs
`ExecStart=sh -c '/usr/share/u-boot-env/patch-u-boot-env.sh watchdog_timeout_ms /usr/share/u-boot-env/patch-u-boot-env.config; fw_setenv watchdog_timeout_ms 60000'`

Putting fw_setenv at the end after ; ,` fw_setenv watchdog_timeout_ms 60000` it will set the timer value to 60 secs irrespective of what patch-u-boot-env.sh script set the timer value.

Testing steps
1) Enable the patch-u-boot-env.service to let it run on every boot.
2) Set the timer to 0

rakeshk7097:
Hi @jan-kiszka ,

In my current changes` After=multi-user.target ` is really helpful which makes the patch-u-boot-env service to run at the end of all the all-essential non-graphical services.

Along with this now the service is giving the correct exit status, FAIL in case if fw_setenv silently unable to set the timer.
Giving the correct exit status of service was also required which was missing in the previous implementation.

Tested these changes 10 times and every times it gave the expected output (timer set to 60 sec) .

sbobade:
> Hi @jan-kiszka ,
> 
> In my current changes`After=multi-user.target` is really helpful which makes the patch-u-boot-env service to run at the end of all the all-essential non-graphical services.
> 
> Along with this now the service is giving the correct exit status, FAIL in case if fw_setenv silently unable to set the timer. Giving the correct exit status of service was also required which was missing in the previous implementation.
> 
> Tested these changes 10 times and every times it gave the expected output (timer set to 60 sec) .

@jan-kiszka please review and provide feedback, we are willing to finalize the patch for upcoming release thanks

recipes-bsp/patch-u-boot-env/files/patch-u-boot-env.service
```diff
@@ -8,12 +8,14 @@
...
+After=multi-user.target
 Conflicts=shutdown.target
 Before=swupdate.service shutdown.target
+DefaultDependencies=no
 
 [Service]
 Type=oneshot
-ExecStart=sh -c '/usr/share/u-boot-env/patch-u-boot-env.sh watchdog_timeout_ms /usr/share/u-boot-env/patch-u-boot-env.config; fw_setenv watchdog_timeout_ms 60000'
+ExecStart=/usr/share/u-boot-env/patch-u-boot-env.sh watchdog_timeout_ms /usr/share/u-boot-env/patch-u-boot-env.config
 ExecStartPost=-/bin/systemctl disable patch-u-boot-env.service
```
colpane, line 20: Should we add Restart=on-failure? What we want is to set watchdog timeout correctly at the end.

rakeshk7097: So, in that case we can better disable the service in the `patch-u-boot-env.sh` if the watchdog timer set is successful.
and if it failed then we can use below in service file

```
# Restart settings
Restart=on-failure
RestartSec=5s
```

@jan-kiszka, could you please give your suggestion on this.

Regards,
Rakesh

jan-kiszka:
I did my own debugging today. Following the theory that we have a dependency issue on $something, I tried the inverse and moved the service rather early (brute force method):
```
Before=swap.target system-modprobe.slice systemd-journald.service systemd-modules-load.service systemd-udev-trigger.service tee-supplicant.service shutdown.target
DefaultDependencies=no
```
And that works here, all the time I tested so far.

@rakeshk7097, can you confirm the same? Because it that "resolves" the issue as well, we are rather looking for a concurrency with some unknown service that is interfering with fw_setenv while it runs.

jan-kiszka:
@colpane provided some dump of the envs after a failed write today. And this is our finding now:
- first env is valid, contains timeout=0 and was marked as "obsolete" by setenv
- second env contains timeout=60000, is marked as active - but its CRC is invalid

Here is why:
```
00000000  09 44 8d fd 01 61 64 64  72 5f 66 69 74 3d 30 78  |.D...addr_fit=0x|
00000010  39 30 30 30 30 30 30 30  00 61 72 63 68 3d 61 72  |90000000.arch=ar|
...
000012e0  00 76 65 6e 64 6f 72 3d  73 69 65 6d 65 6e 73 00  |.vendor=siemens.|
000012f0  77 61 74 63 68 64 6f 67  5f 74 69 6d 65 6f 75 74  |watchdog_timeout|
00001300  5f 6d 73 3d 36 30 30 30  30 00 00 00 00 00 00 00  |_ms=60000.......|
00001310  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
*
00008af0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
00008b00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
*
00020000
```
If I flip those ff to 0 as well, the checksum is fine again. We have some issue with writing the flash, and that may even be the same as we saw when writing complete firmware images. Further examinations are ongoing.

rakeshk7097:
> I did my own debugging today. Following the theory that we have a dependency issue on $something, I tried the inverse and moved the service rather early (brute force method):
> 
> ```
> Before=swap.target system-modprobe.slice systemd-journald.service systemd-modules-load.service systemd-udev-trigger.service tee-supplicant.service shutdown.target
> DefaultDependencies=no
> ```
> 
> And that works here, all the time I tested so far.
> 
> @rakeshk7097, can you confirm the same? Because it that "resolves" the issue as well, we are rather looking for a concurrency with some unknown service that is interfering with fw_setenv while it runs.

Hi @jan-kiszka ,

I have tested your changes and observed 5 successes out of 7 tests, with 2 failures where the system did not update the timer to 60 seconds.

Again, the issue appears to be sporadic.

We may need to improve the success rate and ensure that, in case of failure, the true exit status is correctly reported—returning 'failed' if the timer update is unsuccessful and 'success' if it is updated correctly.

rakeshk7097:
> I did my own debugging today. Following the theory that we have a dependency issue on $something, I tried the inverse and moved the service rather early (brute force method):
> 
> ```
> Before=swap.target system-modprobe.slice systemd-journald.service systemd-modules-load.service systemd-udev-trigger.service tee-supplicant.service shutdown.target
> DefaultDependencies=no
> ```
> 
> And that works here, all the time I tested so far.
> 
> @rakeshk7097, can you confirm the same? Because it that "resolves" the issue as well, we are rather looking for a concurrency with some unknown service that is interfering with fw_setenv while it runs.

Hi @jan-kiszka 

I have tested your changes and observed 3 successes out of 5 tests, with 2 failures where the system did not update the timer to 60 seconds.

Again, the issue appears to be sporadic.

We may need to improve the success rate and ensure that, in case of failure, the true exit status is correctly reported—returning 'failed' if the timer update is unsuccessful and 'success' if it is updated correctly.

You have correctly identified the root cause. It seems that there is an issue with writing to the flash. At times, the MMC block device may not be ready to accept the fw_setenv command

BaochengSu:
`00008b00` is the start of a page, and it's not the boundary of any sector, so seems to me not an erase problem, but rather a program issue.

And according to the code of libubootenv, the flash is erased at 32KB, the `00008b00` is not the boundary of 32KB, so I don't think erase issue can occur in the middle.

so possible reason:

- Data was overwritten during write, this can be checked by a printf after the write in `mtdwrite` of libubootenv.
- driver sent wrong data to flash. This can be checked on the wire with a logic analyzer.
- Timing issue. There could be some more delays between PAGE commands for this flash in the driver. Also this can be checked from logic analyzer.

jan-kiszka:
I think it is rather unlikely that libubootenv passed inconsistent data to the driver. It is single-threaded and calculates the checksum right before passing the buffer to the driver.

BaochengSu:
Today I've made a script to reproduce the issue, turned out the reproduce rate is *HIGH* enough, I've only run twice, the 1st time was the 7th, and the second time is the 17th, so a very rough reproduce rate: 1/12.

my script: (need install `xxd` & `libarchive-zip-perl`)

```sh
#!/bin/sh

iteration=0

while true; do
    iteration=$((iteration + 1))

    rm -rf ./*.bin
    # Print current time
    echo "Current time: $(date '+%Y-%m-%d %H:%M:%S') iteration: $iteration"

    # Run fw_setenv command with the current iteration count
    echo "Setting test_env to $iteration"
    fw_setenv test_env "$iteration"

    for mtd in mtd3 mtd4; do
        # Read /dev/$mtd to a bin file
        dd if=/dev/$mtd of=$mtd.bin 2>/dev/null
        dd if=$mtd.bin of=$mtd.data.bin skip=5 iflag=skip_bytes 2>/dev/null

        # Check CRC32 of $mtd.bin
        crc32_calculated=$(crc32 $mtd.data.bin)
        crc32_stored=$(dd if=$mtd.bin bs=1 count=4 2>/dev/null | xxd -p | \
            tr -d '\n' | sed 's/../& /g' | awk '{print $4$3$2$1}')
        echo "CRC32 calculated for $mtd: $crc32_calculated"
        echo "CRC32 stored for $mtd: $crc32_stored"
        if [ "$crc32_calculated" != "$crc32_stored" ]; then
            echo "Error: CRC32 mismatch in $mtd.bin"
            exit 1
        fi
    done

    echo ""
    # Pause for a short duration before the next loop iteration
    sleep 1
done
```

rakeshk7097:
> @rakeshk7097 , does this happen only on specific devices or this is common, I don't think you have any logic analyzer, I think a reproducible setup can help this debug.

Hi @BaochengSu,

I have cross-verified the issue on another device and did not encounter any problems while setting the U-Boot environment. The iteration 7/7 was successful.

I will use your script to verify the issue on the same device and update you.

BaochengSu:
> > @rakeshk7097 , does this happen only on specific devices or this is common, I don't think you have any logic analyzer, I think a reproducible setup can help this debug.
> 
> Hi @BaochengSu,
> 
> I don't have the logic analyzer now, I have cross-verified the issue on another device and did not encounter any problems while setting the U-Boot environment. The iteration 7/7 was successful.
> 
> Thank you for sharing the script, I will use your script to verify the issue on the same device and update you.

Make sure to align the kernel version

BaochengSu:
Reproduced without fw_setenv: (need apt install mtd-utils)

```
#!/bin/sh

# need apt install mtd-utils

iteration=0

while true; do
    iteration=$((iteration + 1))

    rm -rf ./*.bin
    rm -rf ./*.txt

    echo "Current time: $(date '+%Y-%m-%d %H:%M:%S') iteration: $iteration"
    dd if=/dev/urandom of=random.bin bs=128k count=1 2>/dev/null

    for mtd in mtd3 mtd4; do
        mtd_debug erase /dev/$mtd 0x0 65536
        mtd_debug erase /dev/$mtd 0x10000 65536
        mtd_debug write /dev/$mtd 0x0 131072 ./random.bin
    done

    for mtd in mtd3 mtd4; do
        dd if=/dev/$mtd of=$mtd.bin bs=128k 2>/dev/null
        if ! cmp -s $mtd.bin random.bin; then
            echo "Error: Mismatch between $mtd.bin and random.bin"
            hexdump $mtd.bin -C > dump.flash.txt
            hexdump random.bin -C > dump.random.txt
            diff dump.flash.txt dump.random.txt
            exit 1
        else
            echo "Success: $mtd.bin matches random.bin"
        fi
    done

    echo ""
    # Pause for a short duration before the next loop iteration
    sleep 1
done
```

BaochengSu:
Tried to write in 256-byte chunks, does not work. I guess something wrong inside the ospi controller driver. Need more investigation. Maybe write a similar script in u-boot to check if the u-boot driver has the same problem? Or next week I can write a dummy code on r5 code for programming the flash region.

BaochengSu:
Seems one page was split into two PAGE(0x02) programming commands for unknown reason, which should be acceptable. However, the second one was sent without a WRITE_ENABLE (0x06) command, which is required by the flash, and I think this is the root cause.

Normally for a whole page (256 bytes), note the 0x06 before the 0x02:

image

image


Wrong case, the page were split into two PAGE commands 0x02, and the second is without a WRITE_ENABLE (0x06):

image

image

Resulted in:

```
root@iot2050-debian:~/123# diff dump.flash.txt dump.random.txt
5821,5822c5821,5824
< 00016bc0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
< *
---
> 00016bc0  2d f0 45 b3 57 8f b8 fa  bf b4 70 10 4b c0 d3 1a  |-.E.W.....p.K...|
> 00016bd0  35 7f 06 c2 dc 9d ff 4c  88 b4 42 ad ac 02 27 3d  |5......L..B...'=|
> 00016be0  2c 71 01 3e f4 bf 38 f7  72 ff 96 9b 76 1d 56 c9  |,q.>..8.r...v.V.|
> 00016bf0  7e e7 b0 d6 6e d3 df 21  d1 4f c0 fb 9c d1 6e 04  |~...n..!.O....n.|
```

I guess the flash driver is splitting the PAGE commands? Need dive into the code now.

BaochengSu:
After a deep diving into the driver code and the OSPI controller, terns out that this OSPI controller provides a `direct mode`, in this mode, you are accessing the flash memory just like accessing a normal memory, and the SPI commands are generated by the hardware controller. Then the command data is sent to the spi controller TX FIFO for sending.

So a reasonable guessing is that normally the command generating should be faster than the TX FIFO consuming, and suddenly for some unknown reason the generating is slower and the TX FIFO is exhausted, in such case the command generating shall end the currently partial command to fill the TX FIFO. This makes the PAGE command to be truncated.

This is only my guessing but if I slow down the SPI clock to 4.2MHz, and programming 250 times without error. So I guess this shall be a reasonable guessing.

BaochengSu:
I will check the testing result tomorrow, if it looks promising then at least we will have a workaround. (well, the flash programming may be slowed down but I don't think it should be big deal?)

BaochengSu:
Slowing down the flash SPI clock does reduce the reproduction rate to 1/1000, however it does not fix the issue.

This issue does not appear in the mainline kernel, so there should be already patches available.

jan-kiszka:
You also want to check the very latest 6.1-cip - if there was a recognized fix in mainline, that should have been backported by now. If there were refactortings that happen to change this as well, those may indeed be limited to mainline.

BaochengSu:
> You also want to check the very latest 6.1-cip - if there was a recognized fix in mainline, that should have been backported by now. If there were refactortings that happen to change this as well, those may indeed be limited to mainline.

Already checked, no fix... Seems to me either a backporting missing, or some kind of overlooked side-effect of something else...

Currently I highly suspect the DMA part, since I've noticed somewhere the DMA had issue and had been fixed. And if the data was feed to the command generating via DMA, it could be slowed down by some DMA issue in theory.

Anyway it's time to bisect the mainline.

BaochengSu:
Turns out the current head of mainline disabled the DAC and use INDAC but this will be soon reverted by https://patchwork.kernel.org/project/spi-devel-general/patch/20250328173500.3687483-1-miquel.raynal@bootlin.com/.

rakeshk7097:
Closing this ticket as a better workaround has been merged with the commit below.

https://github.com/siemens/meta-iot2050/commit/605c620787d7683680f85b7d0d6ebfebd144fc28

PR #587, opened 2025-03-04
switchserialmode: suffix package with -iot2050

chombourger:
Several Siemens products have a switchserialmode package. The version included in this repository is really designed for the IOT2050: note the hardware dependency in the name of the package by prefixing it with "iot2050-". The package uses a Provides clause for backward compatibility (both for Isar builds and for apt).


chombourger:
@BaochengSu please review

BaochengSu (approved):

PR #578, opened 2025-01-30
iot2050-efivarfs-helper: handle efivarfs dynamically during runtime

rakeshk7097:
Instead of enforcing a persistent read-write (rw) mount via /etc/fstab, this change ensures efivarfs is mounted and remounted dynamically during command execution. This approach improves flexibility, avoids potential conflicts, and ensures efivarfs is available with the correct permissions only when needed.
Key improvements:

* Removed the need for a persistent rw entry in /etc/fstab.
* Ensures efivarfs is correctly mounted and remounted as needed during execution.
* Prevents unnecessary system-wide modifications while maintaining secure access.

This update enhances robustness and ensures a controlled, on-demand handling of efivarfs.

PR #592, opened 2025-03-24
[V2] Lee/bugfix: iot2050-efivarfs-helper: Fix the permission issue with efivarfs

huaqianli:
1.  [Revert "iot2050-efivarfs-helper: workaround the readonly efivarfs"](https://github.com/siemens/meta-iot2050/pull/592/commits)

This reverts commit https://github.com/siemens/meta-iot2050/commit/77ded9343348008e4da67cdd9a694615edd8d750.

The efivarfs filesystem is automatically mounted as read-only by the
kernel during boot. And Systemd treats certain filesystems, such as sys,
proc, dev, as API file systems, and it does not allow fstab or .mount to
mount these API filesystems now.

Signed-off-by: Li Hua Qian <huaqian.li@siemens.com>

2. [linux-iot2050: Ensure efivarfs is mounted as read-write](https://github.com/siemens/meta-iot2050/pull/592/commits)

The efivarfs filesystem was being mounted as read-only because systemd
attempted to mount it before the TEE-based EFI runtime variable services
(TEE_STMM_EFI) were registered. This caused the efivarfs to lack write
capabilities, which is incorrect behavior.

To resolve this issue, the TEE_STMM_EFI driver is changed from a loadable
module to a built-in kernel component. This ensures that the TEE-based EFI
runtime variable services are initialized earlier during the boot process,
allowing efivarfs to detect write support and be mounted as read-write.

Signed-off-by: Li Hua Qian <huaqian.li@siemens.com>

3. [meta-hailo: Update the README](https://github.com/siemens/meta-iot2050/pull/592/commits)

Signed-off-by: Li Hua Qian <huaqian.li@siemens.com>

4. [Fix boot issue when customizing UEFI keys](https://github.com/siemens/meta-iot2050/pull/592/commits)

After upgrading ISAR, the commit https://github.com/ilbers/isar/commit/f7bce40 of ISAR introduced
a mandatory fixed value for SOURCE_DATE_EPOCH (1709565251,
corresponding to 2024-03-04 12:14:11). This fixed value is problematic
because it results in an incorrect signing time, causing the device to
fail to boot when customizing UEFI keys.

This commit sets SOURCE_DATE_EPOCH to the timestamp of the latest V tag.
This approach ensures that the signing time is meaningful and aligns
with the source code state, making it more logical and reliable.

Signed-off-by: Li Hua Qian <huaqian.li@siemens.com>


recipes-app/iot2050-efivarfs-helper/files/iot2050-efivarfs-helper.tmpl
```diff
@@ -224,6 +242,13 @@ examples:
         exit_on_failure('You have to be root to run this program. \
 If you really want to run as non-root, please use the --force option.')
 
+    remount_efivarfs('rw')
+
     do_prepare()
```
BaochengSu, line 247: - remove the `check_mount_efivarfs` from `do_prepare`
- append the new `remount_efivarfs` to the end of  `do_prepare`
- add a new `do_cleanup` and call `remount_efivarfs('ro')` there
- call the `do_cleanup` before the final `sys.exit`

Or, refactor this code to use context manager so that the `with` can be used, which is more pythonic and easier to understand and maintain.

recipes-app/iot2050-efivarfs-helper/files/iot2050-efivarfs-helper.tmpl
```diff
@@ -144,6 +145,23 @@ def change_efivarfs_immutable_flag(immutable: bool):
...
+    """
+    if mode not in ['rw', 'ro']:
+        raise ValueError("Invalid mode. Use 'rw' for read-write or 'ro' for read-only.")
+
+    try:
+        subprocess.run(
+            ['mount', '-o', f'remount,{mode}', '/sys/firmware/efi/efivars'],
+            check=True
+        )
+    except subprocess.CalledProcessError as e:
```
BaochengSu, line 161: I don't think python provides a generic mount function, so choose either this subprocess way or a ctype libc.mount see [this](https://stackoverflow.com/questions/1667257/how-do-i-mount-a-filesystem-using-python). Personally I prefer the latter.

recipes-app/iot2050-efivarfs-helper/files/iot2050-efivarfs-helper.tmpl
```diff
@@ -144,6 +145,23 @@ def change_efivarfs_immutable_flag(immutable: bool):
             chattr_i(filepath, immutable)
             print("Done.")
 
+def remount_efivarfs(mode):
```
BaochengSu, line 148: I would model the parameter as `readonly=True`, then you don't have to perform any argument checking.

BaochengSu: But if you foresee any possible other modes, it will be better to have a generic mode.

BaochengSu:
Also, echo to #578 , please make sure the systemd's intention is to mount the efivarfs as read-only, since the immutable flag, I don't think it's reasonable for systemd to do that.

recipes-app/iot2050-efivarfs-helper/files/iot2050-efivarfs-helper.tmpl
huaqianli: Found better way, fix it in kernel.

huaqianli:
> Also, echo to #578 , please make sure the systemd's intention is to mount the efivarfs as read-only, since the immutable flag, I don't think it's reasonable for systemd to do that.

Correct, fixed in kernel.

recipes-kernel/linux/linux-iot2050-6.1.inc
```diff
@@ -34,8 +34,8 @@ SRC_URI += " \
     file://patches-6.1/ \
     file://${KERNEL_DEFCONFIG} \
     file://iot2050_defconfig_extra.cfg"
-SRC_URI:append:pvu-dma = " \
-    file://iot2050_defconfig_pvu.cfg \
+SRC_URI:append:secureboot = " \
+    file://iot2050_defconfig_secure.cfg \
```
BaochengSu, line 38: Good, this can be a starting point of our kernel security enhancement.

BaochengSu:
I think maybe we could upstream this patch to cip-core, any plan?

BaochengSu (approved):

conf/layer.conf
```diff
@@ -23,3 +23,6 @@ LAYERDIR_meta-iot2050 = "${LAYERDIR}"
 LAYERDIR_meta-iot2050[vardepvalue] = "meta-iot2050"
 
 ISAR_RELEASE_CMD = "git -C ${LAYERDIR_meta-iot2050} describe --long --tags --dirty --always || echo unknown"
+
+require classes/git-timestamp.bbclass
+SOURCE_DATE_EPOCH = "${@get_git_tag_timestamp(d)}"
```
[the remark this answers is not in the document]
huaqianli: I plan to arise this to upstream latter.
Why you thought it is strange, the `git-timestamp.bbclass` file?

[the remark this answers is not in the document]
jan-kiszka: Isar and Yocto already have a mechanism to define `SOURCE_DATE_EPOCH` per component - your workaround is global. Did you study https://github.com/ilbers/isar/blob/master/meta/lib/oe/reproducible.py?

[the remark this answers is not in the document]
huaqianli: Yes, I know creating a `__source_date_epoch.txt` file can works for the component.

However, I think a mandatory fixed value for SOURCE_DATE_EPOCH (1709565251, corresponding to 2024-03-04 12:14:11) is not a good practice, using the latest tag will make more sense, so I redefine the global SOURCE_DATE_EPOCH before arising to upstreamming.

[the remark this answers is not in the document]
jan-kiszka: I suspect you won't succeed with the global change upstream, specifically as this affects isar and even yocto in principle. But if you want to question what isar imported, go to the isar mailing list. If the signing recipe of isar-cip-core has time-stamping issue, you can discuss that on cip-dev.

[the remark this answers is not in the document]
huaqianli: It might be. I'll dive into deeper and bring it to public with a proposal.

PR #594, opened 2025-04-01
fix(packages): introduce PR in the version string for mraa

chombourger:
The upstream mraa project does not seem to create versions/tags for the changes they make (version 2.2.0 was released in 2020 but the project has continued to receive updates since). We anyhow have our own set of patches on-top that we also need to version. Use CHANGELOG_V to include PR into the version string of our mraa package. PR should be incremented whenever we change the upstream revision (SRCREV) and/or update our patches. This will help with binary distributions of the package (e.g. for users needing to pull packages on their IOT2050 using apt).

### DIFF
--- a/recipes-app/mraa/mraa_2.2.0+git20220805.8b1c5493.bb
+++ b/recipes-app/mraa/mraa_2.2.0+git20220805.8b1c5493.bb
@@ -7,6 +7,9 @@
 # This file is subject to the terms and conditions of the MIT License.  See
 # COPYING.MIT file in the top-level directory.
 #
+
+PR = "1"
+
 inherit dpkg
 
 DESCRIPTION = "Low Level Skeleton Library for Communication on GNU/Linux platforms"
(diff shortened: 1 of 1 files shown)


BaochengSu:
Hi @chombourger,

Thanks for the patch. Like [1], I suggest to also change the package name to `mraa_2.2.0+git20220806.8b1c5493`. The PR should be reset to 1 for each upstream version update, without the name change, it's impossible to identify different versions from the same `mraa_2.2.0+git-1`.

Also a thorough checking should be performed on all the similar recipes. I could arrange one myself.

[1]. https://packages.debian.org/bookworm/ymuse

chombourger:
> Thanks for the patch. Like [1], I suggest to also change the package name to `mraa_2.2.0+git20220806.8b1c5493`. The PR should be reset to 1 for each upstream version update, without the name change, it's impossible to identify different versions from the same `mraa_2.2.0+git-1`.

Updated. Changed the date to 20220805 since Aug 5th is the date that I see on GitHub when checking https://github.com/eclipse/mraa/commit/8b1c54934e80edc2d36abac9d9c96fe1e01cb669

BaochengSu (approved):

PR #598, opened 2025-06-11
iot2050-efivarfs-helper: include PR to track changes

ananthakr23:
There have been modification in the iot2050-efivarfs-helper.tmpl but without any bump in the version (PV). This causes the downsteam projects to miss the changes when building with cached apt. Incldue PR in the recipe to show the modifications.

### DIFF
--- a/recipes-app/iot2050-efivarfs-helper/iot2050-efivarfs-helper_1.0.0.bb
+++ b/recipes-app/iot2050-efivarfs-helper/iot2050-efivarfs-helper_1.0.0.bb
@@ -13,6 +13,7 @@ DESCRIPTION = "Efivarfs Helper"
 MAINTAINER = "baocheng.su@siemens.com"
 
 SRC_URI = "file://iot2050-efivarfs-helper.tmpl"
+PR = "1"
 
 TEMPLATE_FILES = "iot2050-efivarfs-helper.tmpl"
 


recipes-app/iot2050-efivarfs-helper/iot2050-efivarfs-helper_0.1.bb
```diff
@@ -13,6 +13,7 @@ DESCRIPTION = "Efivarfs Helper"
 MAINTAINER = "baocheng.su@siemens.com"
 
 SRC_URI = "file://iot2050-efivarfs-helper.tmpl"
+PR = "1"
```
BaochengSu, line 16: @ananthakr23 Per my understanding, `PR` is more useful when upstream version stays the same but the downstream recipe (debianized one, or contents living in /debian dir) need revision. However, in this case, we are both the upstream and downstream.

In general, I do think it's a good idea to introduce the `PR`, however, I am thinking that the `PV` need to be managed as well:
- Use semantic version (maybe we can use 1.0.0 now) to replace the current 0.1 scheme as PV.
- Introduce the `PR` as suggested by this pulling request.
- When there is change to the `iot2050-efivarfs-helper.tmpl`: Bump the `PV` in the semantic version way, and reset `PR` to its starting value.
- When change is only for the .bb file, (which only affect the debian dir): Bump the `PV` so that package manager can be notified.

Based on that, could you please change the package version to 1.0.0, and reset the PR to the starting value (only if 1 is not the starting one) ?

BaochengSu: @huaqianli 

I think we shall add some maintenance rules and apply them globally in this repo:

- For recipes that has 3rd party upstream: maintain a PR for the recipe, bump the PR for each change of the recipe (the debian part), reset the PR when upstream version bumps.
- For recipes without upstream (e.g. this one): maintain it according to my comment.

huaqianli: I completely agree. Let's add it to the to-do list.

ananthakr23: Hi, @BaochengSu, Thank you for explanation and suggestions. I have renamed the file to use 1.0.0 as PV and set PR=1. Modified the commit message accordingly.

PR #600, opened 2025-06-20
Upgrade recipes-app for volunerbilities

huaqianli:
iot2050-confi-webui: use --prefer-offline instead of --offline for npm install
npm.bbclass: use --prefer-offline instead of --offline for npm install
Revert "ci: Temporarily bypoass the pipeline issue"
recipes-app: Update recipe versions to follow new versioning guidelines
Add recipe maintenance and versioning guidelines
Update Debian snapshot to 2025-06-17
tcf-agent: Bump to 1.9.0
node-red-node-sqlite: Bump to 1.1.0
node-red-node-serialport: Bump to 2.0.3
node-red-dashboard: Bump to 3.6.5
node-red-contrib-opcua: Bump to 0.2.339
node-red-contrib-modbus: Bump to 5.43.0
node-red: Bump to 4.0.9
iot2050-conf-webui: bump dependencies to latest versions


recipes-app/iot2050-conf-webui/iot2050-conf-webui_1.0.0.bb
```diff
@@ -223,7 +223,7 @@ do_install() {
 }
 
 do_prepare_build:append() {
-    INSTALL_FLAGS="--offline --omit=dev --no-package-lock --verbose \
+    INSTALL_FLAGS="--prefer-offline --omit=dev --no-package-lock --verbose \
```
BaochengSu: Make sure it can pass air gaped offline build.

CONTRIBUTING.md
```diff
@@ -35,6 +35,42 @@ as linux kernel or u-boot. Please use the command below to generate such patches
...
+## Recipe Maintenance and Versioning Guidelines
+
+When maintaining recipes in this repository, it is important to distinguish
+between upstream and downstream changes, and to manage the recipe version (PV)
+and recipe revision (PR) consistently.
+
+- **PV (Recipe Version):**  
+  PV represents the version of the software being packaged. It should follow
+  semantic versioning (e.g., `1.0.0`) rather than a simple scheme like `0.1`.
+  This makes version tracking clearer and aligns with best practices.
```
BaochengSu: Actually `0.1` [is also semantic versioning](https://semver.org/#spec-item-4).

I suggest to make this sentence simple to just say `follow semantic versioning 2.0 (https://semver.org/spec/v2.0.0.html)`

recipes-app/iot2050-conf-webui/iot2050-conf-webui_1.0.0.bb
huaqianli: According to https://docs.npmjs.com/cli/v11/commands/npm-search, itt will bypass the staleness checks for cached data:
```sh
[prefer-offline](https://docs.npmjs.com/cli/v11/commands/npm-search#prefer-offline)
Default: false
Type: Boolean
If true, staleness checks for cached data will be bypassed, but missing data will be requested from the server. To force full offline mode, use --offline.

[offline](https://docs.npmjs.com/cli/v11/commands/npm-search#offline)
Default: false
Type: Boolean
Force offline mode: no network requests will be done during install. To allow the CLI to fill in missing cache data, see --prefer-offline.
```

CONTRIBUTING.md
```diff
@@ -35,6 +35,42 @@ as linux kernel or u-boot. Please use the command below to generate such patches
...
+
+- **PR (Recipe Revision):**  
+  PR is the recipe revision. It is incremented when changes are made to the
+  recipe itself (such as modifications in the debianized files or the `/debian`
+  directory) without changing the upstream software version (PV). PR should be
+  reset to its initial value whenever PV is updated.
+
+- **Recommended Workflow:**  
+  - When making changes to the core software, bump the PV using semantic
```
BaochengSu: s/core software/package code/

CONTRIBUTING.md
```diff
@@ -35,6 +35,42 @@ as linux kernel or u-boot. Please use the command below to generate such patches
...
+  - For recipes with a third-party upstream:
+    - Maintain a PR for the recipe.
+    - Bump the PR for each change to the recipe (especially the debian part).
+    - Reset the PR when the upstream version (PV) is updated.
+  - For recipes without an upstream (maintained solely in this repository):
+    - Follow the above semantic versioning for PV and increment PR as needed for
+     recipe changes.
+
+By applying these rules consistently, we ensure clear version tracking, easier
+maintenance, and reliable package updates across the repository.
```
BaochengSu: I think yocto has explained this concept rather clearing, I suggest to link their description directly then here we just tell our in-addition:
- `1` as the PR starting, not `r0`.
- we use semver 2.0 for own packages.

recipes-app/iot2050-conf-webui/README.md
BaochengSu, line 1: a typo in the commit message s/1.0.1/1.0.0/

recipes-app/iot2050-conf-webui/iot2050-conf-webui_1.0.0.bb
BaochengSu: Make sure this `prefer-offline` option never breaks the offline building suppose you are building (meta-iot2050) from a machine has no network functionality (air gap machine).

Such kind of option may impose some assumption on your network linking status, make sure this one does not.

classes/npm.bbclass
```diff
@@ -218,7 +218,7 @@ do_install() {
 }
 
 do_prepare_build:append() {
-    INSTALL_FLAGS="--offline --only=production --no-package-lock --verbose \
+    INSTALL_FLAGS="--prefer-offline --only=production --no-package-lock --verbose \
```
jan-kiszka: But online fetches will still fail, won't they? Is this only improving the readability of error?

We must not fetch during build, hope that is clear.

recipes-app/iot2050-conf-webui/iot2050-conf-webui_1.0.0.bb
jan-kiszka: As written on the other commit: We must not fetch during build, even when not being in strict offline mode.

classes/npm.bbclass
BaochengSu: I guess online fetch will success, otherwise this patch does not make sense.

I agree with you that the build shall not touch the fetching.

@huaqianli You had broken cache for outputs from do_fetch? Maybe a local storage issue? Could it be stably reproducible?

huaqianli: Online fetches are successful

huaqianli: The error looks like the following with `--offline` mode, which the package wasn't pre-fetched:

```sh
npm verb stack Error: request to https://registry.npmjs.org/@node-rs/bcrypt-linux-x64-musl/-/bcrypt-linux-x64-musl-1.10.4.tgz failed: cache mode is 'only-if-cached' but no cached response is available.
```

BaochengSu: So maybe do_npm_fetch is incomplete?

huaqianli: Currently I don't what is the problem, I'll do more test.

BaochengSu: If need more time to fix, maybe the last two commits can be dropped at this moment and we can merge others.

recipes-app/iot2050-conf-webui/iot2050-conf-webui_1.0.0.bb
huaqianli: Hi @jan-kiszka and @BaochengSu,

The `npm install --offline` issue is a known problem with recent npm versions (likely npm 7.x+ and Node.js 16.x+, though the exact versions are uncertain) that enforce stricter rules for offline mode installations, particularly when using the `-g` (global) flag. npm attempts to fetch package metadata (packument) from the registry to resolve dependencies, which causes offline installations to fail with:

npm ERR! code ENOTCACHED
npm ERR! request to https://registry.npmjs.org/<packagename\> failed: cache mode is 'only-if-cached' but no cached response is available.

Both Yocto and npm communities have encountered and addressed this issue using similar approaches - pre-caching metadata during the fetch phase:

**Yocto's approach:**
- Discussion: https://lists.openembedded.org/g/openembedded-core/topic/npm_bblcass_dysfunctional_in/91183010?p=
- Implementation:
  - `poky/meta/classes/npm.bbclass`
  - `poky/meta/lib/oe/npm_registry.py`
  - `meta-openembedded/meta-oe/recipes-devtools/nodejs/nodejs-oe-cache-16.20/oe-npm-cache`

**npm's recommendation:**
- https://github.com/npm/cli/issues/3764

I've implemented both solutions, but they come with significant build time overhead - increasing from 1.5h to over 5h:
- Pipeline run: https://github.com/siemens/meta-iot2050/actions/runs/16064043311

Implementation details following npm maintainers' suggestions can be found here: https://github.com/siemens/meta-iot2050/pull/600/commits/28deff878ab02c5456ff3dd7aa6fdac40147375c

I also implemented optimizations to the solution, but local testing still shows significant build time impact of over 5h:
- Optimized implementation: https://github.com/huaqianli/meta-iot2050/commit/dc9428aa80f8317c9071b3e96d17fc64dac66cb5


## Proposed Path Forward

The current implementation works but may not be practical for CI/CD pipelines due to the build time penalty. 

Given the significant build time impact, I suggest we:

1. **Short-term workaround**: Use `--prefer-offline` to mitigate the issue while maintaining reasonable build times
2. **Long-term solution**: Evaluate these approaches (discussed with baocheng):
   - Containerize the npm/Node.js build environment for better isolation and caching
   - Implement a custom npm install following Debian packaging practices
   - Investigate selective caching (only cache problematic packages)

What are your thoughts on this approach? Should we prioritize build time or completely offline capability?

classes/npm.bbclass
huaqianli: Let's discuss it in the first discussion.

recipes-app/iot2050-conf-webui/iot2050-conf-webui_1.0.0.bb
jan-kiszka: If we open up the offline gate, we have to ensure that really only metadata is fetched, no packages. And while that might be the case for the current setup as that was developed under --offline first, how to avoid that future versions will silently start to bypass the fetching step?

huaqianli: Maybe Layer 8 so far and near future.

jan-kiszka: And in addition, --prefer-offline will still break over real offline with meta-iot2050.

huaqianli: image

According above pic, only metadata is fetched by `--pref-offline`.

BaochengSu (approved):

PR #602, opened 2025-07-03
feat(kas): update kas-container to 4.8.1

asrinuva:


asrinuva:
@huaqianli please have a  look, thanks.

jan-kiszka:
Please compare you commit to previous kas updates in this repo.

asrinuva:
> Please compare you commit to previous kas updates in this repo.

You meant to capture the high level changelog in this commit?

jan-kiszka:
-> .gitlab-ci.yml. But I just noticed that the last update attempt was incomplete as well, so this was not totally obvious.

huaqianli:
> @huaqianli please have a look, thanks.

Thank you for raising this important point. I'd like to address 3 key concerns:

1. Frequency of KAS Updates
- Could you provide more context about why this update is necessary soon after the 4.7 version?
- What specific improvements or fixes in the new KAS version justify this quick update for our project?

2. Update Workflow Concerns
The current update process seems inefficient due to:
- Repeated file copying across multiple repositories, such as, `cip-core`, `meta-iot2050`,...
- Potential for inconsistency and manual overhead

3. Authorship and Attribution
- An important ethical consideration is that simply copying from upstream without crediting the original author is not a good open-source practice. 
- In this case, somehow we can treat it as your own modification.

I don't have a definitive solution right now, especial for the 2nd concern. This is an open discussion to collectively find a more efficient approach to managing these updates.

I'm interested in hearing your thoughts and any ideas you might have to improve our current workflow. Maybe @jan-kiszka and @BaochengSu also have some good ideas?

jan-kiszka:
We established the pattern of carrying a copy of the upstream kas-container script in your local repo in order to encode version dependencies on kas and the build container. Attributions are not lost if script is not modified.

In kas, we are slowly working towards a looser coupling between script and container. But even if that once exists, you will still have to tell both your local users and the CI pipelines which container version to use - in addition to a minimal script version. So I do not see a more handy pattern than carrying script copies in your layers yet, but I'm always open for ideas.

Regarding what the update brings: I think nothing essential yet in this case, but once isar starts to use bubblewrap instead of some of its sudo calls, you will need 4.8+. For more (e.g. kas diff, rootless docker), see https://github.com/siemens/kas/blob/master/CHANGELOG.md.

huaqianli:
> -> .gitlab-ci.yml. But I just noticed that the last update attempt was incomplete as well, so this was not totally obvious.

Hi @asrinuva  would you please proceed this `.gitlab-ci.yml` updating?

huaqianli (approved):

PR #603, opened 2025-07-10
Divide customizations recipe into smaller ones

colpane:
To not get unnecessary features in customizations-example, it is divided into smaller recipes that include feature specific changes. Downstream layers can easily get which feature they want and still override anything they want.

To use latest version of change-root-homedir, isar and isar-cip-core are updated, too.


colpane:
@jan-kiszka @huaqianli @BaochengSu let me know your feedback.

recipes-core/customizations-ssh/customizations-ssh_0.2-iot2050-debian.bb
```diff
@@ -0,0 +1,19 @@
...
+# Authors:
+#  Le Jin <le.jin@siemens.com>
+#
+# This file is subject to the terms and conditions of the MIT License.  See
+# COPYING.MIT file in the top-level directory.
+#
+
+inherit dpkg-raw
+
+DESCRIPTION = "IOT2050 reference image customizations ssh"
```
jan-kiszka: This only enables direct login as root via ssh - maybe make description and even recipe name more specific?

kas/iot2050.yml
```diff
@@ -29,13 +29,13 @@ repos:
 
   isar:
     url: https://github.com/ilbers/isar
-    commit: 938dab363006529f0b5296dfbc09ef306fc8c3d7
+    commit: 8e3122249a52db470efb32b92ed9c44e4b758712
```
BaochengSu, line 32: Better to mention in the commit message the motivations or the benefits of the updatings.

recipes-core/customizations-led/customizations-led_0.2-iot2050-debian.bb
```diff
@@ -0,0 +1,24 @@
+#
```
BaochengSu: The name "customizations-" is somehow mis-leading... 

Some parts are not customizations, but rather essential patches to fulfill the product definition.

For this one, it implements the defined LED behavior in iot2050 operating instructions.

It's reasonable for downstream to have its own status definition, just I suggest to use `iot2050-status-led` as the package name.

recipes-core/customizations-led/customizations-led_0.2-iot2050-debian.bb
```diff
@@ -0,0 +1,24 @@
...
+# Authors:
+#  Le Jin <le.jin@siemens.com>
+#
+# This file is subject to the terms and conditions of the MIT License.  See
+# COPYING.MIT file in the top-level directory.
+#
+
+inherit dpkg-raw
+
```
BaochengSu: Use 1.0 as package version, and also add PR, see https://github.com/siemens/meta-iot2050/commit/ed1ae5233ea3b76875853dd512bd5ae0fdd79c3f

BaochengSu: Same for others.

recipes-core/customizations-network/files/20-assign-ethernet-names.rules
BaochengSu, line 1: This file is also essential for IOT2050, otherwise the ethernet port number order will be reversed to against the printed label.

BaochengSu: Also I don't think downstream should touch this.

recipes-core/customizations-network/customizations-network_0.2-iot2050-debian.bb
```diff
@@ -1,5 +1,5 @@
 #
-# Copyright (c) Siemens AG, 2019
+# Copyright (c) Siemens AG, 2019-2025
 #
```
BaochengSu: Maybe make this as "iot2050-nm-settings"

recipes-core/customizations-serial-port/files/20-create-symbolic-link-for-serial-port.rules
BaochengSu, line 1: Also this file is essential as well, otherwise the ttyUSB for the serial port (x30) would change in case of multiple USB-to-serial devices are connected.

recipes-core/images/iot2050-image-example.bb
```diff
@@ -32,7 +32,11 @@ IMAGE_INSTALL += " \
     sshd-regen-keys \
     regen-rootfs-uuid \
     install-on-emmc \
-    customizations-example \
+    customizations-led \
+    customizations-network \
+    customizations-serial-port \
+    customizations-ssh \
+    change-root-homedir \
```
BaochengSu, line 39: Where does this "change-root-homedir" come from? isar?

BaochengSu:
> We also have customizations-base, originally with the idea that this should be something that everyone want to have. Looking at the NODE_PATH it sets, it looks like this diverged as well. This refactoring should cover the base elements as well.
> 
> For networking, I would split-off the interface aliases from interface configurations.

Indeed, also the X30 alias.

kas/iot2050.yml
huaqianli: Indeed, we just updated ISAR not too far ago (actually 3 commits in advance).

recipes-core/customizations-ssh/customizations-ssh_0.2-iot2050-debian.bb
colpane: Done

kas/iot2050.yml
colpane: I wanted to align with isar-cip-core's isar commit. I can mention this in commit msg or I can remove this. Just let me know :)

recipes-core/customizations-led/customizations-led_0.2-iot2050-debian.bb
colpane: Done

recipes-core/customizations-led/customizations-led_0.2-iot2050-debian.bb
colpane: Done

recipes-core/customizations-network/files/20-assign-ethernet-names.rules
colpane: Moved under base packages.

recipes-core/customizations-network/customizations-network_0.2-iot2050-debian.bb
colpane: Done

recipes-core/customizations-serial-port/files/20-create-symbolic-link-for-serial-port.rules
colpane: Moved under base packages

recipes-core/images/iot2050-image-example.bb
colpane: Comes from isar-cip-core

recipes-core/ssh-root-login/files/postinst
```diff
@@ -0,0 +1,11 @@
...
+# Copyright (c) Siemens AG, 2019-2025
+#
+# Authors:
+#  Le Jin <le.jin@siemens.com>
+#
+# This file is subject to the terms and conditions of the MIT License.  See
+# COPYING.MIT file in the top-level directory.
+#
+
+echo "PermitRootLogin yes" >> /etc/ssh/sshd_config
```
BaochengSu, line 11: It's not a good practice to modify this file directly, however, I think this is the current implementation already?

colpane: Unfortunately, yes. Any suggestion for alternative way? I can switch to it.

recipes-core/nodejs-module-path/files/postinst
```diff
@@ -8,8 +8,5 @@
 # COPYING.MIT file in the top-level directory.
 #
 
-echo $HOSTNAME > /etc/hostname
-echo "127.0.0.1	${HOSTNAME}" >> /etc/hosts
-
 # set nodejs environment for module searching
 echo "NODE_PATH=\"/usr/lib/node_modules/:/usr/local/lib/node_modules/\"" >> /etc/environment
```
BaochengSu, line 12: @huaqianli This is a pretty old code, do we still need it on Debian 12?

recipes-core/images/iot2050-image-example.bb
BaochengSu: Better to use a third commit to add this package?

kas/iot2050.yml
huaqianli: Now the window is still opening, then we can update it, but please mention it in commit msg. Thanks.

recipes-core/images/iot2050-image-example.bb
huaqianli: Agree, it is good to split out the `change-root-homedir` into its own commit separated from the refactoring.

huaqianli: Ah, it is because of the modification of `recipes-core/iot2050-nm-settings/files/postinst`, then better to explain a little bit in commit message.

recipes-core/nodejs-module-path/files/postinst
huaqianli: It is no longer recommended by Node.js or Debian 12. However, for the version we are using (18.13.0), the documentation (see [Node.js v18.13.0 Modules](https://nodejs.org/docs/v18.13.0/api/modules.html)) includes the following description:

```txt
In order to make modules available to the Node.js REPL, it might be useful to also add the /usr/lib/node_modules folder to the $NODE_PATH environment variable. Since the module lookups using node_modules folders are all relative, and based on the real path of the files making the calls to require(), the packages themselves can be anywhere.
```

huaqianli: Then, I think we can reserve it so far.

huaqianli:
Otherwise, LGTM

recipes-core/images/iot2050-image-example.bb
colpane: So should I separate it to another commit or explain it in current commit?

huaqianli:
BTW, @colpane Has it been successfully built and tested?

huaqianli (approved):

recipes-core/ssh-root-login/files/postinst
BaochengSu: Let's keep it at the moment.

recipes-core/customizations-led/customizations-led_0.2-iot2050-debian.bb
BaochengSu: PR is still missing? See this https://github.com/siemens/meta-iot2050/commit/de55d6f72beb611cff1ca62e62e85bdcb0c51a52 as reference.

colpane: Sorry, now it is done.

colpane:
> BTW, @colpane Has it been successfully built and tested?

Locally built successfully.

recipes-core/iot2050-base-packages/iot2050-base-packages_1.0.bb
huaqianli, line 1: @colpane could you also add 'PR' for this recipe.

colpane: Done

PR #609, opened 2025-07-23
fix(firmware-update): use semantic version comparison

rakeshk7097:
Plain string comparison fails for versions like "4.10" vs "4.2", causing incorrect results. This patch uses `packaging.version` to ensure proper semantic version comparison.

### DIFF
--- a/recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.tmpl
+++ b/recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.tmpl
@@ -114,6 +114,7 @@ import time
 import tarfile
 import textwrap
 import tempfile
+from packaging import version
 from abc import ABC, abstractmethod
 from ctypes import *
 from enum import Enum
@@ -635,7 +636,9 @@ class FirmwareTarball(object):
 
     def __check_os(self, target_os, os_info) -> bool:
         for tos in target_os:
-            if tos.key in os_info and os_info[tos.key] >= (tos.min_version):
+            current_version = version.parse(os_info[tos.key])
+            required_version = version.parse(tos.min_version)
+            if current_version >= required_version:
                 return True
             print("\nFirmware requires a minimal version of ",
                   tos.min_version, ", the current OS has ",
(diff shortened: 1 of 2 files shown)


rakeshk7097:
Having plain string comparison is giving wrong result and unable to update the firmware

`iot2050-firmware-update u-boot.tar.xz 
Current board: SIMATIC IOT2050 Advanced PG2
 
Firmware requires a minimal version of 4.2, the current OS has 4.10.
OS image version must be newer than the minimal version, no firmware image could be updated on this device`

So "4.10" is mistakenly considered older than "4.2"

This patch replaces plain comparison with proper semantic version parsing using the packaging.version module, which correctly interprets versions as major.minor.patch. With this fix, firmware updates now proceed as expected.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.tmpl
```diff
@@ -635,7 +636,9 @@ class FirmwareTarball(object):
 
     def __check_os(self, target_os, os_info) -> bool:
         for tos in target_os:
-            if tos.key in os_info and os_info[tos.key] >= (tos.min_version):
+            current_version = version.parse(os_info[tos.key])
```
BaochengSu, line 639: Better to update the comments (Line 90+ in this file) as well? To note version checking is based on semver 2.0

BaochengSu: Never mind, after checking the document of the packaging, it seems it also supports other schema.

recipes-app/iot2050-firmware-update/files/iot2050-firmware-update.tmpl
```diff
@@ -635,7 +636,9 @@ class FirmwareTarball(object):
...
+            required_version = version.parse(tos.min_version)
+            if current_version >= required_version:
```
BaochengSu, line 641: This script also support the example image versioning (two digits for one segment: V0x.0x.0x) as well, does this change cover the example versioning?

rakeshk7097: Thanks for pointing that out! I verified this with a quick test using packaging.version.parse("V01.05.01"), and it correctly parses it as 1.5.1.
So, the current patch supports our versioning scheme (V0x.0x.0x) as-is — no additional parsing or stripping needed, I think.
thanks!

BaochengSu (approved):

rakeshk7097:
@BaochengSu , kindly merge the MR if there are no additional comments, thanks!

PR #617, opened 2025-08-20
fix(swupdate-config): install swupdate-config machine specific

rakeshk7097:
Install correct swupdate-config package as per machine specific, this solve the below build issue for qemu image.

| DEBUG: Shell function rootfs_import_package_cache finished | DEBUG: Executing shell function rootfs_install_pkgs_download | Reading package lists...
| Building dependency tree...
| Reading state information...
| Package swupdate-config is a virtual package provided by:
|   swupdate-config-iot2050-qemu 0.2
|   swupdate-config-iot2050 0.2


jan-kiszka:
Why didn't that explode in CI so far? Or is the isar-cip-core version triggering this newer than what we currently select here...?

PR #635, opened 2025-12-17
Fix MAP3 (ASIC controller) firmware update failure

ChunJiaoZhao:
The MAP3 (ASIC controller) firmware cannot be updated because the updater is incompatible with libgpiod v2.x and the SPI device node is missing due to the SPI bus being disabled in the device tree.


meta/recipes-kernel/linux/files/patches/0014-WIP-dts-iot2050-Add-spidev-for-IOT2050-SM.patch
```diff
@@ -4,17 +4,25 @@ Date: Mon, 18 Dec 2023 10:47:06 +0800
 Subject: [PATCH] WIP: dts: iot2050: Add spidev for IOT2050-SM
 
 This spidev is used for communicating between main SoC and the SM ASIC.
+The parent SPI bus which is disabled by default also sbould be enabled.
```
huaqianli: no need to add this sentence.

ChunJiaoZhao: Fixed

PR #637, opened 2025-12-19
Fix missing broadcast for “EIO needs update”

ChunJiaoZhao:
When the EIO controller firmware differs from the system’s expected version, a prompt such as “Suggest to update the firmware via iot2050-eio” should be broadcast.
Currently it isn’t, because session information cannot be obtained via psutil. This change uses the who command to retrieve active sessions instead of psutil, ensuring the message is broadcast as intended.

### DIFF
--- a/meta-sm/recipes-app/iot2050-eio-manager/files/iot2050_eio_fwu_monitor.py
+++ b/meta-sm/recipes-app/iot2050-eio-manager/files/iot2050_eio_fwu_monitor.py
@@ -11,7 +11,7 @@ import sys
 import time
 import os
 import textwrap
-import psutil
+import subprocess
 import grpc
 from gRPC.EIOManager.iot2050_eio_pb2 import CheckFWURequest
 from gRPC.EIOManager.iot2050_eio_pb2_grpc import EIOManagerStub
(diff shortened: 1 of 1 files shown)


meta-sm/recipes-app/iot2050-eio-manager/files/iot2050_eio_fwu_monitor.py
huaqianli, line 1: typo: s/iot00/iot2050/

ChunJiaoZhao: Done

meta-sm/recipes-app/iot2050-eio-manager/files/iot2050_eio_fwu_monitor.py
```diff
@@ -47,6 +47,23 @@ def broadcast_update_info(message):
     '''
     os.system(f'wall "{prompt}"')
 
+def list_logged_in_users():
+    """Return a list of (user, tty) for logged-in sessions.
+    """  
+    try:
+        cp = subprocess.run(["who"], capture_output=True, text=True, check=False)
```
huaqianli: `result` would be a slightly more descriptive variable name here. However, `cp` is alos a common and acceptable abbreviation, so feel free to keep it as is.

ChunJiaoZhao: done

meta-sm/recipes-app/iot2050-eio-manager/files/iot2050_eio_fwu_monitor.py
```diff
@@ -11,7 +11,7 @@
 import time
 import os
 import textwrap
-import psutil
+import subprocess
```
huaqianli, line 14: While direct shell commands are generally discouraged for better portability and maintainability, this approach is temporarily necessary to address a specific compatibility issue. Once the incompatibility is addressed, this section should be refactored to use a more integrated or native method.

PR #640, opened 2026-01-07
Fix the base image building error due to missing dependency

BaochengSu:
Original error message:

> WARNING: iot2050-image-base-1.0-r0 do_rootfs_postprocess: Looks like you have ssh host keys in the image but did  not install "sshd-regen-keys". This image should not be  deployed more than once.
> ERROR: iot2050-image-base-1.0-r0 do_rootfs_postprocess: Install the package or forcefully remove this check!
> ERROR: iot2050-image-base-1.0-r0 do_rootfs_postprocess: ExecutionError('/work/build/tmp/work/iot2050-debian-arm64/iot2050-image-base-iot2050/1.0-r0/temp/run.image_postprocess_sshd_key_regen.4785', 1, None, None)
> ERROR: Logfile of failure stored in: /work/build/tmp/work/iot2050-debian-arm64/iot2050-image-base-iot2050/1.0-r0/temp/log.do_rootfs_postprocess.4785
> ERROR: Task (/work/build/../../repo/meta/recipes-core/images/iot2050-image-base.bb:do_rootfs_postprocess) failed with exit code '1'
> NOTE: Tasks Summary: Attempted 201 tasks of which 194 didn't need to be rerun and 1 failed.

PR #643, opened 2026-01-26
meta/images: fix missing dependency for iot2050-image-base

rakeshk7097:
Include the dependent package `sshd-regen-keys` in `iot2050-image-base` to fix the following build error during rootfs post-processing:

WARNING: iot2050-image-base-1.0-r0 do_rootfs_postprocess: Looks like you have ssh host keys in the image but did  not install "sshd-regen-keys". This image should not be  deployed more than once. ERROR: iot2050-image-base-1.0-r0 do_rootfs_postprocess: Install the package or forcefully remove this check!
ERROR: iot2050-image-base-1.0-r0 do_rootfs_postprocess: ExecutionError('/work /build/tmp/work/iot2050-debian-arm64/iot2050-image-base-iot2050/1.0-r0/temp/ run.image_postprocess_sshd_key_regen.171019', 1, None, None)

### DIFF
--- a/meta/recipes-core/ssh-root-login/ssh-root-login_1.0.0.bb
+++ b/meta/recipes-core/ssh-root-login/ssh-root-login_1.0.0.bb
@@ -8,13 +8,14 @@
 # COPYING.MIT file in the top-level directory.
 #
 
-PR = "1"
+PR = "2"
 
 inherit dpkg-raw
 
 DESCRIPTION = "Permit root login for ssh"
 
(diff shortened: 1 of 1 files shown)


huaqianli:
Considering `ssh-root-login` won't be a mandatory package soon in the future, we merge #640 for this dependency issue.

jan-kiszka:
Which was a mistake for the mentioned reasons. The base image has no business in ssh.

PR #648, opened 2026-02-10
fix base image build error due to unable to locate sshd-regen-keys

ananthakr23:
Base image build failed with error:
> | DEBUG: Executing shell function rootfs_install_pkgs_download
> | Reading package lists...
> | Building dependency tree...
> | Reading state information...
> | E: Unable to locate package sshd-regen-keys
> | WARNING: exit code 100 from a shell command.
> | DEBUG: Executing shell function rootfs_do_umounts
> | DEBUG: Shell function rootfs_do_umounts finished
> | DEBUG: Python function do_rootfs_install finished

use IMAGE_INSTALL instead of IMAGE_PREINSTALL to fix the issue as sshd-regen-keys is provided by isar.


ananthakr23:
@huaqianli Updated the MR. Please review.

PR #665, opened 2026-04-14
Add support for Third-party Trust Center signing flow and EFI/FWU integration

huaqianli:
This MR extends the IOT2050 secure boot flow with Trust Center-based signing support and integrates it into U-Boot, OTP provisioning, and firmware update package generation.

**Main changes:**

- add the external OpenSSL signing provider
- package Trust Center credentials and remote signing client
- enable Trust Center signing for U-Boot and OTP provisioning
- switch FIT image signing to SHA-512
- add package signing support for firmware update packages
- expose FWU secure-boot options in Kconfig
- enable EFI boot for the example image
- clean up the OTP helper script
- switch OTP kas includes to repo-relative paths
- drop the Isar patch that is already merged upstream

**Impact:**

- secure boot signing can now be performed either locally or through the Trust Center flow
- firmware update packages can participate in the secure signing flow
- the example image gains EFI boot support
- configuration and maintenance are simplified by removing the upstreamed Isar patch

**Why**

The goal is to make secure boot enabled by default easier to maintain and easier to integrate into production signing workflows. In particular, this MR adds the pieces required to use Trust Center-managed keys for image signing while keeping the existing build flow usable for non-production cases.

**Testing**

- verified builds for u-boot-iot2050, minimal image, example image, and all-image configurations
- verified both legacy/non-EFI and EFI boot for the example image
- verified secure-boot related build combinations with:
- secure boot disabled
- secure boot enabled without keys
- secure boot enabled with dummy/local keys
- secure boot enabled with Trust Center keys
- verified firmware-update package generation with signing support
- verified that non-Trust-Center builds do not require Trust Center payloads
- verified that the EFI-related changes do not break minimal/base image builds or the non-EFI boot path
- SWU example image build
- negative tests for missing keys / missing Trust Center payloads
- signed update package install/apply validation on target
- clean rebuilds when switching between non-secure, dummy-key, and Trust Center-secure configurations


meta/recipes-bsp/u-boot/files/0009-board-siemens-iot2050-Skip-distro-EFI-bootmgr-by-def.patch
```diff
@@ -0,0 +1,78 @@
...
+no-op command when it still has the generic default value. This keeps the
+change board-local and preserves any explicit user override stored in the
+environment.
+
+This is a temporary board-local workaround for the current EFI image setup,
+until the affected images provision EFI Boot#### entries explicitly.
+
+The secureboot flow is not affected because its secure-boot command path
+loads `efi/boot/bootaa64.efi` directly and does not rely on the distro EFI
+boot manager.
```
jan-kiszka: Then why not changing the boot command directly, just like in the secure path? Less hacky...

huaqianli: Good point. I will let the base/minimal image use the same EFI boot path, instead of keeping EFI support example-image-only. Then it is good to go with changing the boot command

BaochengSu (approved):

meta/recipes-bsp/u-boot/files/0009-board-siemens-iot2050-Skip-distro-EFI-bootmgr-by-def.patch
huaqianli: The only regression is that the firmware cannot boot the older image, which is acceptable.

PR #678, opened 2026-06-25
u-boot-iot2050: Regenerate patch series to fix formatting issues

huaqianli:
Previous patches contained formatting errors (likely AI-generated), which did not affect functionality but broke git patch validity. Regenerate the patch series using git to ensure correct format.

### DIFF
--- a/meta/recipes-bsp/u-boot/files/0009-lib-efi_loader-Silence-missing-EFI-var-file-on-first.patch
+++ b/meta/recipes-bsp/u-boot/files/0009-lib-efi_loader-Silence-missing-EFI-var-file-on-first.patch
@@ -1,11 +1,11 @@
-From 3de6b2f8ef5b8dbe206dcf4594908c39571db83f Mon Sep 17 00:00:00 2001
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
 From: Li Hua Qian <huaqian.li@siemens.com>
 Date: Thu, 16 Apr 2026 10:30:00 +0000
-Subject: [PATCH] lib: efi_loader: silence missing EFI var file on first boot
+Subject: [PATCH] lib: efi_loader: Silence missing EFI var file on first boot
 
-When EFI variables are stored in `ubootefi.var`, the first boot after
+When EFI variables are stored in ubootefi.var, the first boot after
 flashing may legitimately not have that file yet. The existing
-`efi_var_from_file()` comment already documents that this case should be
+efi_var_from_file() comment already documents that this case should be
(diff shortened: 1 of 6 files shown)


meta/recipes-bsp/u-boot/files/0009-lib-efi_loader-Silence-missing-EFI-var-file-on-first.patch
```diff
@@ -1,11 +1,11 @@
-From 3de6b2f8ef5b8dbe206dcf4594908c39571db83f Mon Sep 17 00:00:00 2001
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
 From: Li Hua Qian <huaqian.li@siemens.com>
 Date: Thu, 16 Apr 2026 10:30:00 +0000
 Subject: [PATCH] lib: efi_loader: silence missing EFI var file on first boot
```
BaochengSu: The "Silence" in the filename should be "silence", at least it is the lowercase for me if generating via git format-patch ....

huaqianli: Yes, `bcompare` brought this gap when I double checked it.  Since U-Boot does not follow a consistent style, I decided not to modify it.

Image

jan-kiszka:
Are you properly contributing all these downstream U-Boot patches to mainline first?

meta/recipes-bsp/u-boot/files/0009-lib-efi_loader-Silence-missing-EFI-var-file-on-first.patch
huaqianli: Let me update the U-Boot commit message, as I prefer to keep the filename style consistent.

huaqianli:
> Are you properly contributing all these downstream U-Boot patches to mainline first?

Unfortunately, not all of them. The latest three patches are still in the TODO state.

jan-kiszka:
Please also be careful with AI-generated stuff: https://lore.kernel.org/u-boot/20260515220758.GM1858239@bill-the-cat/. The pendulum might be moving, but it is not there yet.

jan-kiszka:
Also an important reminder: Upstreaming is no longer a "nice to have". With the CRA in place and the fact the U-Boot is so far not proving any stable branches, the only proper way for picking up security fixes is moving forward with the base versions. If you (or our customers, when doing their own signing) need to rebase these downstream patches on every update...

BaochengSu (approved):

PR #679, opened 2026-06-29
linux-iot2050: Enable heap memory zeroing on free by default

huaqianli:
## Summary
Enable zero-initialization of freed kernel memory by default on IOT2050.

## Change
- Set `CONFIG_INIT_ON_FREE_DEFAULT_ON=y` in:
  - `meta/recipes-kernel/linux/files/iot2050_defconfig_extra.cfg`

## Why
This turns on `init_on_free=1` by default, reducing residual sensitive data in freed memory and improving kernel hardening. and it is driven by EU CRA-aligned secure-by-default hardening requirements.

Trade-off: a small performance overhead on free paths is expected and accepted for this hardening benefit.

## Validation
- Confirmed merged kernel config contains:
  - `CONFIG_INIT_ON_FREE_DEFAULT_ON=y`
- Runtime check on target:
  - zcat /proc/config.gz | grep CONFIG_INIT_ON_FREE_DEFAULT_ON
  - cat /proc/cmdline (ensure no init_on_free=0)
- Functional check (debug kernel with LKDTM):
  - `READ_AFTER_FREE` / `READ_BUDDY_AFTER_FREE` show memory poisoning behavior

## Notes
- Config-only change, no kernel source modification.
- Expected minor performance overhead on free paths.


jan-kiszka:
Reasoning missing.

huaqianli:
> Reasoning missing.

Thanks for the feedback — agreed that the reasoning should be explicit.  

I will update the PR description to include this rationale and validation evidence.

BaochengSu:
Better to put a why into the commit message.

huaqianli:
> Better to put a why into the commit message.

Agree and done.

BaochengSu (approved):

PR #686, opened 2026-08-10
chore(kernel): build CONFIG_BLK_DEV_RAM as a module

colpane:
Switch CONFIG_BLK_DEV_RAM from y to m so the RAM disk block device driver is built as a loadable module instead of being linked into the kernel image.

Debian ships this option as a module in its kernel configs, and aligning with that reduces config drift and keeps our build closer to the well-tested upstream/Debian baseline.

Systems that need the RAM disk driver available before modules can be loaded (e.g. as part of early boot) must ensure the `brd` module is included in the initramfs.


colpane:
@huaqianli please check :)

huaqianli (approved):

PR #691, opened 2026-08-28
Rework Arduino IO hardware

BaochengSu:
The hardware design of the Arduino IO will be changed in next revision of the hardware of the PG2 Advanced & M.2 variants.

The new hardware will use the I2C IO expander directly to single the GPIO. Only GPIO is changed.

This requires the new SEBoot with the new "rev" field of the boardinfo to be workable, which is in the backlog and will be submitted soon.

### DIFF
--- a/meta/recipes-bsp/u-boot/files/0015-board-siemens-iot2050-Add-Arduino-rev-support.patch
+++ b/meta/recipes-bsp/u-boot/files/0015-board-siemens-iot2050-Add-Arduino-rev-support.patch
@@ -1,0 +1,950 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Baocheng Su <baocheng.su@siemens.com>
+Date: Mon, 10 Aug 2026 19:57:01 +0800
+Subject: [PATCH] board: siemens: iot2050: Add Arduino rev support
+
+Read the board revision from system information and select the matching
+device tree for IOT2050 Advanced PG2 and M2 variants.
+
+Add U-Boot and SPL support for the revised Arduino interface, including
+board revision handling and device tree selection.
+
+Sync the device tree files with the Linux kernel. This includes changes
+already upstream and IOT2050-specific changes that are not upstream yet.
+---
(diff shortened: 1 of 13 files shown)


meta/recipes-bsp/u-boot/files/0015-board-siemens-iot2050-Add-Arduino-rev-support.patch
```diff
@@ -0,0 +1,950 @@
...
++	};
++};
++
++&mcu_uart0 {
++	pinctrl-names = "default";
++	pinctrl-0 = <&mcu_uart0_pins_default>;
++	status = "okay";
++};
++
++&mcu_spi0 {
```
huaqianli, line 385: Shall we also add `status = "okay"`?

meta/recipes-bsp/u-boot/files/0015-board-siemens-iot2050-Add-Arduino-rev-support.patch
```diff
@@ -0,0 +1,950 @@
...
+Date: Mon, 10 Aug 2026 19:57:01 +0800
+Subject: [PATCH] board: siemens: iot2050: Add Arduino rev support
+
+Read the board revision from system information and select the matching
+device tree for IOT2050 Advanced PG2 and M2 variants.
+
+Add U-Boot and SPL support for the revised Arduino interface, including
+board revision handling and device tree selection.
+
+Sync the device tree files with the Linux kernel. This includes changes
```
huaqianli, line 12: Could we split it into two patches for rev support and syncing device tree form kernel?

BaochengSu: It is planned to do it during upstreaming.

meta/recipes-bsp/u-boot/files/0015-board-siemens-iot2050-Add-Arduino-rev-support.patch
BaochengSu: This is copied from old device tree, let me check it during upstream.

huaqianli (approved):

PR #697, opened 2026-09-09
Stabilize firmware update operation progress handling

huaqianli:
## Summary

This PR stabilizes firmware update progress handling across the system firmware
service, CLI, and firmware manager.

## Changes

1. `iot2050-firmware-update: Type operation progress contract`
   - Add typed operation states, stages, progress events, terminal results, and
     error responses.
   - Return a successful accepted response for asynchronous operations.
   - Recover unfinished operations as `interrupted`.
   - Fix operation finalization by avoiding duplicate `operation_id` arguments.

2. `iot2050-firmware-update: Consume typed progress in CLI`
   - Update the CLI to consume typed operation progress.
   - Avoid duplicate progress output using event sequence numbers.
   - Display progress messages and handle terminal states correctly.

3. `iot2050-fwmgr: Map typed system progress`
   - Update fwmgr to consume typed system firmware stages and results.
   - Propagate progress messages through the firmware manager.
   - Fix static stage mapping and callback handling.

## Validation

- Runtime-tested CLI firmware update flow.
- Runtime-tested Firmware Center/Web firmware update flow.
- Verified progress reporting across backup, U-Boot, and environment stages.
- Verified operation completion and failure handling.
- Operation contract tests pass.
- Python syntax check and `git diff --check` pass.
- Tested with the three packages deployed together:
  `iot2050-firmware-update`, `iot2050-fwmgr`, and
  `iot2050-firmware-common`.


huaqianli:
I do not recommend perfecting the API/design at the moment. For example:

- `GetXxx()` is a clear and (standard RPC naming convention)[https://google.aip.dev/131]
   for retrieval operations . There is no strong reason to rename it to an OO-style method
  name.

- Keeping both synchronous `Update()`/`Rollback()` and asynchronous
  `StartUpdate()`/`StartRollback()` is a reasonable trade-off. The synchronous
  APIs support blocking CLI-style callers, while the asynchronous APIs support
  long-running operations with persistent progress and result tracking.

- The typed operation state and stage contract should be kept. It allows
  clients to reliably poll server-side persistent state after a client
  disconnects, a browser refreshes, or the service restarts. A streaming
 API could be considered later as an optional real-time enhancement,


At this stage, the expected benefit of a perfect API design does not justify its
implementation scope and regression risk. The current design is a reasonable
trade-off and should be kept unless a concrete limitation or value-facing
requirement emerges.

huaqianli:
Closed due to #699 .

